PR #25, opened 2019-09-14
live refresh of local edits of crossplane docs

prasek:
Updated the readme.md with the trick that @suskin uses for live refresh of local crossplane docs edits.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
 # Run jekyll in development mode
+
+LOCAL_DOCS_DIR := docs/local
+
 run: _data/versions.json
 	docker run --rm -it \
 		-p 4000:4000 -p 4001:4001 \
@@ -6,6 +9,16 @@ run: _data/versions.json
 		-v "$(PWD):/srv/jekyll" \
 		jekyll/jekyll -- \
 		jekyll serve --livereload --livereload-port 4001
+
+run_docs_local: local_docs_dir _data/versions.json
+	docker run --rm -it \
+		-p 4000:4000 -p 4001:4001 \
+		-v="$(PWD)/vendor/bundle:/usr/local/bundle" \
+		-v "$(PWD):/srv/jekyll" \
+		-v "$(GOPATH)/src/github.com/crossplaneio/crossplane/docs:/srv/jekyll/$(LOCAL_DOCS_DIR)" \
+		jekyll/jekyll -- \
+		jekyll serve --livereload --livereload-port 4001
+	rm -d $(LOCAL_DOCS_DIR)
 
 # Build (output is in _site)
 build: _data/versions.json
@@ -38,3 +51,8 @@ _data/versions.json: node_modules docs
 node_modules: package.json package-lock.json
 	npm install
 	@touch node_modules
+
+# Create the local docs dir needed before _data/versions.json in some cases
+local_docs_dir: 
+	mkdir -p $(LOCAL_DOCS_DIR)
+
--- a/README.md
+++ b/README.md
@@ -7,5 +7,20 @@ This is the the source for http://crossplane.io. It is rendered using [Jekyll](h
 This runs locally watching for changes and live reloading.
 
 ```
+brew install npm
+
 make run
 ```
+
+Open http://localhost:4000 in your browser.
+
+## To run locally with local crossplane docs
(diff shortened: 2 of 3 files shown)


suskin:
Yessss 🚀

README.md
```diff
@@ -9,3 +9,27 @@ This runs locally watching for changes and live reloading.
...
 ```
+
+## To work locally on crossplane docs changes
+Create a new empty directory in `docs` that will be picked up by jekyll.
+```
+$ cd docs
+$ mkdir v0.3
+```
+
+Update `make run` to mount your local src path into the jekyll container, by adding the following:
```
negz: Should we have `make run` work this way by default, rather than suggesting folks edit it?

prasek: `make run` is probably used by @connorchan and others to only edit the mainline website without `$(GOPATH)/src/github.com/crossplaneio/crossplane/docs` present.

So added `make run_docs_local` which does everything needed and trimmed the README.md.

Makefile
```diff
@@ -1,4 +1,7 @@
 # Run jekyll in development mode
+
+LOCAL_DOCS_DIR := docs/v0.3.local
```
jbw976: does this "local docs" folder need to contain v0.3 explicitly? it would be nice not to hard code a version in there that we'll have to keep updating.

prasek: @jbw976 easy fix, done

preprocess.js
```diff
@@ -16,13 +16,24 @@ function getDirectories(srcpath) {
 
 const ROOT_DIR = `${__dirname}`;
 
+//alternate sort values for non-semver values
```
jbw976: this `versionMap` and `version` function look like some magic that could probably use a comment explaining WHY it's done.  For example:

```js
// This version map and version function allow versions that do not follow semver syntax to also
// be included in the version selection sorting for the site.  "local" is the developer version
// used when testing docs changes in a local development environment.  We set this "local"
// version as 7.7.7 (a high value) so that it will show up as the "latest" version in the site's
// version selection dropdown.
```

would you mind including something like that?

README.md
```diff
@@ -9,3 +9,14 @@ This runs locally watching for changes and live reloading.
...
 make run
 ```
+
+Open http://localhost:4000 in your browser.
+
+## To run locally with local crossplane docs
+Ensure `$(GOPATH)/src/github.com/crossplaneio/crossplane/docs` is present.
+
+```
+make run_docs_local
```
muvaf: This requires node to be installed. It'd be good to have a small requirements section I believe.

preprocess.js
prasek: done!

README.md
prasek: I had to brew install npm, will add that to the instructions

muvaf: brew instruction assumes that the OS is Mac OS X though

Makefile
```diff
@@ -38,3 +51,8 @@ _data/versions.json: node_modules docs
 node_modules: package.json package-lock.json
 	npm install
 	@touch node_modules
+
+# Create the local docs dir needed before _data/versions.json in some cases
+local_docs_dir:
```
suskin: If this is making the directory, I would recommend that we use `$(LOCAL_DOCS_DIR)` as the name of this recipe. Then we get dependency checking and incremental building for free!

Makefile
```diff
@@ -1,4 +1,7 @@
 # Run jekyll in development mode
+
+LOCAL_DOCS_DIR := docs/local
```
suskin: Nitpick: did we already consider making this configurable? I would recommend making it configurable by doing `LOCAL_DOCS_DIR ?= docs/local` so we can override it at runtime if desired:

```
make LOCAL_DOCS_DIR=docs/myversion run_docs_local
```

Makefile
```diff
@@ -7,6 +10,16 @@ run: _data/versions.json
 		jekyll/jekyll -- \
 		jekyll serve --livereload --livereload-port 4001
 
+run_docs_local: local_docs_dir _data/versions.json
```
suskin: Nitpick: I generally use hyphens for things that humans will type, because it takes fewer keystrokes than `_`. I wouldn't use `node_modules` as an example in this case, because it is only named that way because that's what the directory it builds is called.

PR #38, opened 2020-03-10
Update footer and Twitter desc meta tag

connorchan:


_includes/footer.html
```diff
@@ -2,7 +2,6 @@
   <div>
     <div class="main">
       <div class="logo"><img src="{{ "/images/logo.svg" | relative_url }}" /></div>
-      <span class="md-hidden">The open source multicloud control plane</span>
```
jbw976: should this be changed to "Manage your applications and infrastructure the Kubernetes way" instead of being removed?

connorchan: @jbw976 per Matt, we are simply removing this text from the footer

connorchan:
@grantgumina yes, we removed that in an earlier PR this morning

grantgumina:
Same with the copy changes and typo fixes for "the source of truth" section under "The Kubernetes Way"?

grantgumina:
Kk - looking at the other PRs from this morning. Looks great.

Issue #54, opened 2020-07-22
Add required footer with Linux Foundation copyright text


jbw976:
One of the final steps for onboarding Crossplane into the CNCF (https://github.com/cncf/toc/issues/466) is to add the required footer to the crossplane.io website.

We did this recently on https://rook.io/ in https://github.com/rook/rook.github.io/pull/90/files. Use that PR as a reference for the content that should be added to crossplane.io in a footer, obviously using the crossplane.io styling though 😎

jbw976:
Re-opening until https://github.com/crossplane/crossplane.github.io/pull/55#issuecomment-662590222 is resolved

rathpc:
@jbw976 This should now be resolved

PR #56, opened 2020-07-22
Adding Crossplane Authors to footer

rathpc:
Signed-off-by: Steven Rathbauer <steven@rathpc.com>


rathpc:
Addresses https://github.com/crossplane/crossplane.github.io/pull/55#issuecomment-662590222

Closes #54

PR #58, opened 2020-07-27
Add CNCF logo and language to the homepage

### LINKED ISSUE #57: State that Crossplane is a CNCF project on the crossplane.io homepage
The crossplane.io website should publicly state that it is a CNCF project, as per the [website guidelines](https://github.com/cncf/foundation/blob/master/website-guidelines.md).  Specific requirement:

> Sandbox-level projects should include the sentence “We are a Cloud Native Computing Foundation sandbox project.” and the CNCF logo

You can see a similar example from [rook.io](https://rook.io/) near the bottom, like so:

Screen Shot 2020-07-27 at 11 46 24 PM

Of course, that's a bit different since Rook is past the sandbox now, so please use the required text from above 🙇‍♂️

connorchan:
Resolves #57


connorchan:
@jbw976  
image

PR #59, opened 2020-07-29
Fix typo (Gitub > GitHub)

eliasp:


eliasp:
> Thanks @eliasp! Would you mind signing your commit so this will pass DCO check? https://github.com/apps/dco

Sorry, missed that - done!

PR #62, opened 2020-09-17
Add Hubspot tracking code across site and docs

connorchan:
Resolves https://github.com/upbound/upbound-frontend/issues/1751

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -29,5 +29,6 @@
     <div id="content">{{ content }}</div>
     {% include footer.html %}
     {% include google-analytics.html %}
+    {% include hubspot.html %}
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


hasheddan (approved):

_layouts/default.html
```diff
@@ -29,5 +29,8 @@
     <div id="content">{{ content }}</div>
     {% include footer.html %}
     {% include google-analytics.html %}
+    <!-- Start of HubSpot Embed Code -->
+    <script type="text/javascript" id="hs-script-loader" async defer src="//js.hs-scripts.com/5557732.js"></script>
+    <!-- End of HubSpot Embed Code -->
```
rathpc, line 34: I think we should implement this like we did with GA and create an includes file, that way if it needs to update it is just one place to change it.

rathpc (changes requested):

rathpc (approved):

PR #68, opened 2020-10-28
Update messaging and content

connorchan:


hasheddan (commented):
@connorchan would this be an appropriate PR to include https://github.com/crossplane/crossplane.github.io/issues/67 in or would you prefer to address separately?

connorchan:
@hasheddan I'm happy to add that in if we are able to get an image asset, otherwise I can tackle it in a followup PR later

hasheddan:
@displague could you help here?

rathpc (approved):

PR #86, opened 2021-02-01
Update Stack to Provider

GijsvanDulmen:
Should this be provider?


hasheddan (commented):
@GijsvanDulmen yep :) these docs are generated from the crossplane/crossplane repo based on the branches present there. These updates were made in `master` via https://github.com/crossplane/crossplane/pull/2086, but have not been cherry-picked back to the release-1.0 branch, which is why they have not been updated here

PR #117, opened 2021-08-09
Do not index old documentation versions

### LINKED ISSUE #107: avoid google search indexing old documentation
are we able to exclude old version content from being indexed at Google?

Searching for "crossplane aws rds" should not result in crossplane v0.3 being the first link.

hasheddan:
Updates old documentation pages to include a `noindex` directive to keep them out of search engine results. I verified locally by rendering docs and confirming that old versions do in fact include the `noindex` directive:

Fixes #107 

noindex


hasheddan (commented):
For completeness, I also verified that current docs do not include the tag 👍🏻

luebken (approved):
LGTM

PR #118, opened 2021-08-18
Links to cloud-providers/* for more config info

luebken:
There is some useful information in https://github.com/crossplane/crossplane.github.io/tree/master/docs/master/cloud-providers for the cloud providers that we should expose side by side to the link to Github and the API Docs.


hasheddan (changes requested):
@luebken all changes to docs are made in `crossplane/crossplane`, then automatically synced here :)

luebken:
@hasheddan Ack. Let me try to figure that out.

PR #119, opened 2021-08-22
Remove all unsupported Crossplane versions

hasheddan:
Removes all versions of Crossplane docs that are no longer supported and regenerates the versions list.

xref https://github.com/upbound/build/pull/166 https://github.com/crossplane/crossplane.github.io/issues/107

Sceenshot of site after updates:
only-supported


jbw976 (approved):
Thanks for cleaning up all this old (and no longer supported) content dan!

Is there anything about the build process that would re-sync the old doc versions into this tree again, effectively restoring them back from the dead?

hasheddan:
@jbw976 https://github.com/upbound/build/pull/166 should address this -- however, for these old versions I am manually removing here, it is possible if a commit was merged to those branches then they would get resynced. If we wanted to guard against that we could update their `build` submodules to include the aforementioned functionality 👍🏻

PR #131, opened 2022-01-05
Adding distribution information to the Getting Started guide

grantgumina:
In the current Crossplane documentation, Upbound's hosted control plane feature is presented like it's the same thing as a hosted upstream Crossplane deployment. While this is somewhat true, in reality, it serves very different use cases. 

Users reading the getting started guide see hosed control planes, create an Upbound account, and attempt to go through the Crossplane quickstart using a hosted control plane in Upbound. Because of the underlying differences, they quickly run into trouble, and Upbounders find themselves having to explain why a hosted control plane isn't exactly upstream Crossplane managed by Upbound. For example, a common issue is hosted Crossplane uses the `upbound-system` namespaces vs. the `crossplane-system` one mentioned in the Crossplane docs.

The changes in this PR attempt to do the following:

- Clarify that community members have a choice in distribution when using Crossplane
- Route users who would like to use the Upbound Crossplane distribution to the Upbound docs site in order to avoid confusion
- Make the nested tab structure more clear by commenting begin/end sections in the code


hasheddan (changes requested):
@grantgumina these changes need to be made in https://github.com/crossplane/crossplane

grantgumina:
My bad. I'll close this and reopen in the correct location.

PR #147, opened 2022-08-16
improve website performance, accessibility, SEO

reveurguy:
# before
Screenshot 2022-08-17 at 12 28 43 AM

# after
Screenshot 2022-08-17 at 12 28 57 AM
 
i have made changes to improve the performance, accessibility, SEO of the website.
for performance we can convert the images to next-gen format like webp. I have converted the hero images only and that has made some difference in the performance factor. i still need to convert other images.
for accessibility and SEO, i have added alt tags wherever needed, improved contrast ratios of text/background wherever required, added meta description.


thephred (changes requested):
This looks great. Thank you for the contribution! I wonder if you could include removing the old hero images as they are not used elsewhere and can be removed safely.

css/index.scss
```diff
@@ -351,8 +351,8 @@
   }
 
   button {
-    background-color: #3bbdc4;
-    border-color: #3bbdc4;
+    background-color: #193d54;
```
jbw976, line 354: can you comment a bit more on this button color change? do you have a screenshot of what it would look like? 

i'm hesitant to change the default color scheme, so it would be good to understand the visual impact of this change a bit more :)

reveurguy: # before
Screenshot 2022-08-17 at 11 19 38 AM 

# after 

Screenshot 2022-08-17 at 11 19 30 AM

reveurguy: made these changes to improve the contrast ratio for better visibility and improving SEO

reveurguy:
i have removed the old hero images.

css/index.scss
jbw976: thanks for sharing this screenshot to help us visualize the button change more easily!

jbw976:
Would it be possible for you to link to the the testing tool you are using? I'm hoping to see more details about the results.  I was thinking about that in the context of wondering how much impact the button color had on accessibility, but then thought it would be useful to understand all the results more deeply :)

reveurguy:
yes, i am using Google Lighthouse for the evaluation.

css/index.scss
reveurguy: should i keep this or revert it? this didn't make a huge improvement itself but small improvements contribute to bigger improvements

reveurguy:
I am attaching the before and after pdf reports.

## before
[crossplane-lighthouse.pdf](https://github.com/crossplane/crossplane.github.io/files/9387004/crossplane-lighthouse.pdf)

## after
[crossplane-changes-1.pdf](https://github.com/crossplane/crossplane.github.io/files/9387005/crossplane-changes-1.pdf)

jbw976 (approved):
Thank you @reveurguy for the continued answering of questions and providing deeper details so we can understand these changes and their scope/impact more effectively!  Your patience and interaction with us have been very appreciated.

Let's go ahead and accept the button color change - it's within the rest of the color scheme (it doesn't look out of place) and it contributes to the 100% score for accessibility now, so that's good enough for me, even though it diverges slightly from the original design.

@thephred I think you're all good here too? if so, feel free to switch your "changes requested" to "approved" :)

thephred (approved):
Thank you @reveurguy !

PR #157, opened 2022-10-07
Support /docs location and update menus

plumbis:
Signed-off-by: Pete Lumbis <plumbis@users.noreply.github.com>


hasheddan (approved):
Thanks @plumbis!

PR #167, opened 2022-10-13
Fixes the version dropdown menu

plumbis:
The creation of the v1.10 release exposed an issue where the version drop down menu treats the values as floats and sorts them. As a result `1.10` is treated as `1.1` and comes _after_ `1.9`.

This PR updates all of the version numbers from floats to strings and then sorts based on major, minor versions to build the menu.

An extra fix is in to correct the v1.10 index alias that was incorrectly pointing to master.

### DIFF
--- a/content/docs/v1.10/_index.md
+++ b/content/docs/v1.10/_index.md
@@ -3,9 +3,9 @@ title: "Overview"
 weight: -1
 toc_include: false
 aliases:
-    - /docs/master/index.html
+    - /docs/v1.10/index.html
 cascade:
-    version: master
+    version: "1.10"
 ---
 
 ![Crossplane](/docs/master/media/banner.png)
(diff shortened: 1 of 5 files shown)


content/docs/v1.10/_index.md
```diff
@@ -3,9 +3,9 @@ title: "Overview"
 weight: -1
 toc_include: false
 aliases:
-    - /docs/master/index.html
+    - /docs/v1.10/index.html
 cascade:
-    version: master
+    version: "1.10"
```
hasheddan, line 8: @plumbis i believe all of these fixes will need to happen in the Crossplane repo or else they will get overwritten when syncing -- but perhaps I'm misinterpreting

plumbis:
🤦 

Crossplane PRs:
1.10: https://github.com/crossplane/crossplane/pull/3380
1.9: https://github.com/crossplane/crossplane/pull/3379
1.8: https://github.com/crossplane/crossplane/pull/3381
1.7: https://github.com/crossplane/crossplane/pull/3382

plumbis:
Closing and re-creating PR to fix commits.

PR #168, opened 2022-10-14
Fix version drop down by sorting version strings instead of floats

plumbis:


hasheddan (approved):
Thanks @plumbis!

Issue #165, opened 2022-10-12
Make v1.10 come after v1.9


plumbis:
Menus currently treat v1.10 as a float and it becomes v1.1

plumbis:
Fixed in 3136c899023372c75c336e2f19026e31ed7bf5e4

PR #174, opened 2022-10-26
Bump latest release version to v1.10

hasheddan:
Updates latest release version to v1.10.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -33,5 +33,5 @@ params:
   prodFormLink: "https://docs.google.com/forms/d/e/1FAIpQLSev-5clADSdkwi_wiFqBCAECeIoAQDE91chBbeWbvyTjRCeYg/viewform"
   infoMailToLink: "mailto:info@crossplane.io"
   upboundGithubLink: "https://github.com/upbound"
-  latest: 1.9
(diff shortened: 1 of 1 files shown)


config.yaml
```diff
@@ -33,5 +33,5 @@ params:
   prodFormLink: "https://docs.google.com/forms/d/e/1FAIpQLSev-5clADSdkwi_wiFqBCAECeIoAQDE91chBbeWbvyTjRCeYg/viewform"
   infoMailToLink: "mailto:info@crossplane.io"
   upboundGithubLink: "https://github.com/upbound"
-  latest: 1.9
-  repoLink: "https://github.com/crossplane/crossplane"
\ No newline at end of file
+  latest: 1.10
```
plumbis, line 36: This needs to be a string. Found a place I missed with the fix in #165

plumbis (changes requested):

config.yaml
hasheddan: Thanks for catching @plumbis! Updated :)

plumbis (approved):

PR #209, opened 2022-11-22
Change wording for Github link to `Contribute on GitHub`

jonnylangefeld:
This could be confusing for first time visitors, as there is another path for 'Getting started', which is `Documentation > Getting Started`, which links [here](https://crossplane.io/docs/v1.10/getting-started.html). The GitHub repo itself as resources to contribute, but not on 'how to get started with Crossplane'.

I don't have data for this confusion happening often, but I know of one instance of a user that ended up on the wrong page.

So the new wording should separate the link to Github and the link to Documentation a bit more and lead potential contributors to Github and users to the Documentation page.

### DIFF
--- a/themes/crossplane/layouts/partials/top-nav.html
+++ b/themes/crossplane/layouts/partials/top-nav.html
@@ -14,7 +14,7 @@
     <ul class="links">
       <li>
         <a href="{{ .Site.Params.githubLink  }}">
-          <span class="md-hidden">Get Started on</span> GitHub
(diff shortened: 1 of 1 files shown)


plumbis:
Good feedback! We are going to be launching a new docs site soon and we remove that link and just have the github icon. 

After the site is live we can work on the landing page content to draw attention to a getting started/quickstart guide.

You can view the new site here: http://docs.crossplane.io

plumbis:
This is no longer relevant with the new docs.crossplane.io site. 

Thank you for the contribution!

PR #231, opened 2022-12-13
Add a survey banner

plumbis:
Adds a banner to the latest landing page asking to take the docs survey

### DIFF
--- a/themes/geekboot/layouts/_default/list.html
+++ b/themes/geekboot/layouts/_default/list.html
@@ -14,10 +14,8 @@
 
     <main class="bd-main order-1">
       <div class="bd-intro pt-2 ps-lg-2">
-        {{- if eq .Page.Params.version .Site.Params.latest -}}
         {{- if eq .Title "Overview"}}
           {{ partial "survey" . }}
-        {{ end }}
         {{ end }}
         <div class="d-md-flex flex-md-row-reverse align-items-center justify-content-between">
           <div class="mb-3 mb-md-0 d-flex">
--- a/themes/geekboot/layouts/_default/list.html
+++ b/themes/geekboot/layouts/_default/list.html
@@ -14,6 +14,11 @@
 
     <main class="bd-main order-1">
       <div class="bd-intro pt-2 ps-lg-2">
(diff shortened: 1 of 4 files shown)


themes/geekboot/layouts/_default/list.html
```diff
@@ -14,6 +14,11 @@
 
     <main class="bd-main order-1">
       <div class="bd-intro pt-2 ps-lg-2">
+        {{- if eq .Page.Params.version .Site.Params.latest -}}
```
hasheddan, line 17: Do you think we should make this active on all versions to get more visibility?

plumbis:
I'm open to it. I'm trying to manage how annoying I want to be. If you +1 all versions I'll make the change.

hasheddan:
@plumbis I'm +1 -- survey is to help users and it is only on the home page so I feel good about showing it to more folks 👍🏻

plumbis:
added.

themes/geekboot/layouts/partials/survey.html
```diff
@@ -0,0 +1,5 @@
+<div class="docs-survey card mb-3 text-center col-10 mx-auto">
+    <div class="card-body fs-4">
+        Click to <a href="https://s.surveyplanet.com/uiuwwdxb">take the Crossplane user survey</a>. <br />Survey closes December 31.
```
jbw976, line 3: Is there any way we could also promote the "Crossplane Community Annual Report Survey" located at https://tinyurl.com/3b5cyfap?

That would be really helpful to get more data there, and potentially avoid confusion with the multiple surveys happening right now.  One way to do that may be to make this more clear that it's a documentation focused survey.

plumbis: For sure! When do you plant to close the annual report survey?

jbw976: I take this back - I think i'd rather focus our graduation efforts around a public ADOPTERS.md file, instead of this survey.  Your docs feedback survey is timeboxed, let's just focus on that for now :)

jbw976 (approved):

PR #233, opened 2022-12-14
chore: Add npm.install make target

MisterMX:
Add `npm.install` as a make target to easily install the npm packages required by Hugo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,12 @@ $(HUGO):
 # ====================================================================================
 # Targets
 
+# Setup npm packages
+npm.init:
+	@$(INFO) Installing npm packages
+	@npm install
+	@$(OK) Installed npm packages
+
 # Run local development server.
 run: $(HUGO)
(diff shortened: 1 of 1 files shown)


plumbis:
If we are going to add a directive to install npm modules it should also ensure npm is installed locally.

NPM shouldn't be required unless the user is testing production builds that use [PurgeCSS](https://gohugo.io/hugo-pipes/postprocess/#css-purging-with-postcss)

plumbis (changes requested):
If we are going to add a directive to install npm modules it should also ensure npm is installed locally.

NPM shouldn't be required unless the user is testing production builds that use [PurgeCSS](https://gohugo.io/hugo-pipes/postprocess/#css-purging-with-postcss)

plumbis:
I'm going to close this since we no longer need make for the docs. If there's a desire to have a makefile optionally build the docs we can reopen/approve a new PR.

PR #237, opened 2022-12-15
Code copy (Take 2)

plumbis:
This PR makes the following changes to the code boxes:
* Adds a copy to clipboard button using [Clipboard.js](https://clipboardjs.com/)
* Optional support for defining which lines to copy out of the code box. 
* Adds styling and anchor links to the line numbers
* Adds support to link to line number anchors inside tabs
* Removes line numbers from tables (no user impact, changes styling and JS element targeting) 
* Makes the top NAV _not_ sticky. This is required to make the line number anchors work. Otherwise the line number linked to will go under the navbar and not be in view. Fixing this with a sticky nav is a very large lift (for me). Also note, the side bars are still sticky.
* Removes clipboard scsss files. All clipboard related styles are merged into `code-theme-base.scss`

Note, this doesn't solve #236 (which was discovered in this PR).

Documentation for this will be added to the docs contributing guide.

PR #238, opened 2022-12-16
Highlight codeblocks on hover

plumbis:
This adds support for "hover highlight" in code blocks. A shortcode can be wrapped around a command and on mouse over it will highlight a defined line in a code box. 

Here is an example of the functionality. 
Screen Recording 2022-12-15 at 9 01 24 PM 2022-12-15 21_24_13

This feature is opt-in. Authors must add a `{label="name"}` to a code block and wrap the command in a shortcode, for example, `{{<hover label="name" line="3" >}}MyCommand{{</hover>}}`

Since no docs content has been modified this won't be seen in the Netlify branch preview.

This is a branch of PR #237. This should be merged _after_ that.


utils/webpack/src/js/hoverHighlight.js
```diff
@@ -0,0 +1,60 @@
...
+
+// function remove_highlight_line(line_num, line_text, hl_span){
+//     // get the element's parent node
+//     var parent = hl_span.parentNode;
+//     // move all children out of the element
+//     while (hl_span.firstChild) parent.insertBefore(hl_span.firstChild, hl_span);
+//     // remove the empty element
+//     parent.removeChild(hl_span);
+//     line_text.classList.toggle("hl");
+// }
```
negz, line 58: We typically try to avoid checking in commented-out text.

plumbis: Removed! I ended up changing how the code block HTML was rendered and left in the v1 method.

utils/webpack/src/js/hoverHighlight.js
```diff
@@ -0,0 +1,39 @@
...
+
+    for (let i = 0; i < keywordList.length; i++) {
+        build_eventlistener(keywordList[i].id, keywordList[i].dataset.label, keywordList[i].dataset.line)
+    }
+}
+
+
+export function build_eventlistener(command_id, code_box_label, code_line_number){
+    // get the dynamic variables created by Hugo shortcode
+    // var command_id = document.currentScript.getAttribute('rand_id');
```
hasheddan, line 12: Do we need to include this commented out code?

plumbis: Removed!

hasheddan (approved):
Thanks @plumbis !

plumbis:
Adds support for "hover highlight" in code blocks.

PR #242, opened 2022-12-16
Create an issue template for new releases

plumbis:
When the docs source of truth moves from crossplane/crossplane ([#3461](https://github.com/crossplane/crossplane/issues/3461)) cutting a new Crossplane release will change.

To simplify the process for crossplane/crossplane maintainers this issue template will identify what needs to be done. By having a template we can create a link directly to this issue template.


plumbis:
Closing this to consolidate changes into a single PR

PR #257, opened 2023-01-06
Updated favicons

plumbis:
Looking through the netlify logs there are a large number of 404s looking for specific apple favicons.

Digging further it appears we're missing a few common favicon assets:
https://realfavicongenerator.net/favicon_checker?protocol=https&site=docs.crossplane.io

This regenerates the existing favicons using a higher resolution base image and generates new favicon assets to fix the 404s.

Verified fix of the deploy preview:
https://realfavicongenerator.net/favicon_checker?protocol=https&site=deploy-preview-257--crossplane.netlify.app/v1.10/


thephred (approved):
Thanks @plumbis! One thing to look out for is caching/cache-busting. I don't know how this deploys but depending on how these assets have been and are hosted they may not update to these new assets. There is an example of forcing this below that you removed (`/android-chrome-192x192.png?v=1.0`, specifically the `?v=1.0` part). It may be that you will be fine but mainly just raising the issue in case you haven't poked at it.

PR #259, opened 2023-01-11
Moves the search box JS from end of page to search div

plumbis:
The Algolia search box is created by the Algolia JS block based on a provided div `id`.

Today the script that does this is in the footer of the page, making it the last thing rendered. The problem is that it causes a content shift, moving the nav down to inject the box.

This puts the JS immediately after the div. The JS will be loaded before the rest of the page and prevents the content shift.


thephred (approved):
Thanks @plumbis!

PR #262, opened 2023-01-12
Combine list and single page layouts into a single file

plumbis:
Hugo maintains a different layout file for "list" pages and "single" pages. Our implementation does not differentiate between the two in presentation.

Before this commit, the contents of `single.html` and `list.html` were maintained as copy/paste duplicates.

This moves that content into a partial and both pages call the same partial. 

There are no changes to layout/styling.


thephred (approved):
Thanks @plumbis!

PR #263, opened 2023-01-12
Alert when a feature is alpha/beta

plumbis:
This creates an alert on the page for features tagged as alpha or beta.

Docs indicate this by setting front matter of `state: alpha` or `state: beta`

This generates a banner alert on the top of the page like this:

Screen Shot 2023-01-12 at 5 58 14 PM


negz (approved):

negz:
I'd prefer someone who understands the code also review/approve, but the result LGTM.

It would be useful to link from the banner to more context about the feature states - i.e. https://docs.crossplane.io/v1.10/reference/feature-lifecycle/

plumbis:
Banner adds a link to the lifecycle page. 

Screen Shot 2023-01-12 at 9 42 58 PM

PR #265, opened 2023-01-13
move content media to top level folder

plumbis:
Most of the media files are the same across versions. Instead of duplicating media files across versions, this moves all of the files to `/content/media`. This also updates the `img` shortcode to not assume the image is in the local version folder.


thephred (approved):
Thanks @plumbis!

PR #254, opened 2023-01-03
[Do not merge] Proposed Docs Reorg

plumbis:
This PR is a draft for the docs reorg discussed in [crossplane/crossplane #3398](https://github.com/crossplane/crossplane/issues/3398).

https://deploy-preview-254--crossplane.netlify.app/

This proposes a few major changes, including:
* The creation of a "[Knowledge Base](https://deploy-preview-254--crossplane.netlify.app/knowledge-base/)" for content that is not version specific and can be more in depth on a single topic or cross multiple topics.
* The creation of a "[Developer Guide](https://deploy-preview-254--crossplane.netlify.app/contribute/)" to document existing contribution process and rules for both Crossplane source and Crossplane docs.
* A docs "[landing page](https://deploy-preview-254--crossplane.netlify.app/)" to act as a point of entry for accessing the User Docs, KB or Developer Guide
* A [significant reorganization](https://deploy-preview-254--crossplane.netlify.app/v1.11/) of files within a docs version. Use `v1.11` as the version to see the proposed organization.
    * The diff of files between the current docs and the proposal is here: https://deploy-preview-254--crossplane.netlify.app/v1.11/file-map/

The proposed organization is focused on the existing content and does not map directly to the outline in  [crossplane/crossplane #3398](https://github.com/crossplane/crossplane/issues/3398). 

At this point no work has been done on some of the proposed content merge. Do not focus on the current content of the pages.

PR #269, opened 2023-01-13
Fix search and version drop down

plumbis:
Fixes two issues:
* Oversight on my part that #262 implies a `docs` param. This happened because it was pulled from the larger #254 PR and I missed the update to config.yml
* The script that loads the search box wasn't moved into the header in #259. My local previews worked because of caching. 


thephred (approved):
Thanks @plumbis!

PR #296, opened 2023-01-24
Update azure-provider.md

bartul:
Azure CLI report following error if `az ad sp create-for-rbac` command is executed without specifying `--scope`:
`Usage error: To create role assignments, specify both --role and --scopes.`


bartul:
Sure. Done. I think :)

plumbis (approved):

Issue #299, opened 2023-01-27
Only set Hugo write_stats in Netlify


plumbis:
Hugo [`write_stats`](https://github.com/crossplane/docs/blame/master/config.yaml#L17) is enabled in the Hugo config.  This is used for [PurgeCSS](https://purgecss.com/), which is only used in production builds. 

This is a huge bottleneck for local development causing builds to be 6-8x slower. 

**Enabled**:
```shell
hugo                                                                                                                                                                                                                          ✔
Start building sites …
hugo v0.109.0-47b12b83e636224e5e601813ff3e6790c191e371+extended darwin/arm64 BuildDate=2022-12-23T10:38:11Z VendorInfo=gohugoio

                   | EN
-------------------+------
  Pages            | 188
  Paginator pages  |   0
  Non-page files   |  86
  Static files     | 141
  Processed images |  18
  Aliases          |   0
  Sitemaps         |   1
  Cleaned          |   0

Total in 7865 ms
```

**Disabled**:
```shell
hugo                                                                                                                                                                                                                          ✔
Start building sites …
hugo v0.109.0-47b12b83e636224e5e601813ff3e6790c191e371+extended darwin/arm64 BuildDate=2022-12-23T10:38:11Z VendorInfo=gohugoio

                   | EN
-------------------+------
  Pages            | 188
  Paginator pages  |   0
  Non-page files   |  86
  Static files     | 141
  Processed images |  18
  Aliases          |   0
  Sitemaps         |   1
  Cleaned          |   0

Total in 1049 ms
```

## Fixes
Update the netlify.toml file to add a new [command](https://docs.netlify.com/configure-builds/file-based-configuration/) to uncomment the setting for production builds.

PR #300, opened 2023-01-28
Improve Build Speed

plumbis:
Disable `writeStats` by default and add a `sed` replace to enable it for Netlify builds.


plumbis:
Resolves #299

jeanduplessis (approved):

hasheddan (approved):

PR #320, opened 2023-02-01
corrects link to managed-resources that has incorrect quotes

plumbis:
A markdown link is breaking across lines due to line wrap and is generating a link to `"managed-resources"` (with quotes).

This changes those links to use `ref` shortcode to catch broken links in the future. 


jeanduplessis (approved):
Tested via the preview app, and both links are working.

@plumbis for my interest, why do we have `./` in `{{<ref "./managed-resources" >}}` on the one page but not the other? And why do we have the link defined, `managed-resources]: "managed-resources"` on one page and not the other?

Issue #313, opened 2023-02-01
Inconsistent abbreviations for Composite Resource Claim


negz:
### URL

https://github.com/crossplane/docs/blob/542668/content/v1.11/getting-started/introduction.md?plain=1#L46
https://github.com/crossplane/docs/blob/542668/content/v1.11/getting-started/provider-aws.md?plain=1#L14

Possibly more - I haven't checked all of https://github.com/search?q=repo%3Acrossplane%2Fdocs+XC&type=code

### What's Wrong?

The documentation has a few places that abbreviate claims to "XC". Historically we've abbreviated them to "XRC" for "Composite Resource Claim", per:

* https://github.com/search?q=repo%3Acrossplane%2Fcrossplane+XRC&type=code
* https://github.com/crossplane/docs/blob/master/content/v1.11/concepts/terminology.md#composite-resource-claim

My priority is to be consistent. That said, I'd prefer to standardize on XRC.

negz:
I'm happy to open a PR for this, just checking that we're aligned before I do.

plumbis:
Fine with me, I have no opinions here. I don't remember where I found the `XC` reference but I missed `XRC` originally. Thanks for catching and fixing it.

PR #271, opened 2023-01-13
Docs Reorg for v1.11

plumbis:
_This is a rework of PR #254._

## Summary of changes

- The creation of a "[Knowledge Base](https://deploy-preview-271--crossplane.netlify.app/knowledge-base/)" for content that is not version specific and can be more in depth on a single topic or cross multiple topics.

- The creation of a "[Developer Guide](https://deploy-preview-271--crossplane.netlify.app/contribute/)" covering contributing to the Crossplane docs and a pointer to the [crossplane/crossplane CONTRIBUTING.md](https://github.com/crossplane/crossplane/blob/master/CONTRIBUTING.md)

- A [significant reorganization](https://deploy-preview-271--crossplane.netlify.app/master/) of files within a docs version. This reorg is currently only for the `master` branch docs and will be applied to version v1.11 and new versions going forward. Existing version pages will **not** be changed (read: no broken bookmarks)

## Changes from #254 
- All information about contributing to Crossplane source was removed at the request of @negz.  
_Ref: https://github.com/crossplane/docs/pull/254#discussion_r1067429448_

- Many additional file changes were pushed into smaller PRs that are already merged including:
    - #268 
    - #267 
    - #266 
    - #265 
    - #263 
    - #262
    - #259

## Work in progress
A few things are still required to have MVP content for v1.11's release:
- Install/Uninstall/Upgrade chapters need to be written
- Quickstart guides need to be reviewed extended to include XRs and Claims

PR #326, opened 2023-02-02
Retroactive review of #271

negz:
Fixes https://github.com/crossplane/docs/issues/313

I didn't get a chance to review https://github.com/crossplane/docs/pull/271 very deeply. I spent some time this afternoon reviewing some of the new content. This PR contains a series of small tweaks. It may make sense to review at a commit-by-commit level. Some commit messages have context around why I'm proposing the changes.

### DIFF
--- a/content/v1.11/getting-started/introduction.md
+++ b/content/v1.11/getting-started/introduction.md
@@ -246,10 +246,10 @@ in a _Composite Resource_ .
 * The {{< hover label="specGroup" line="7" >}}versions.name{{</hover >}} 
 that defines the version used in a _Composite Resource_.
 * A {{< hover label="specGroup" line="5" >}}names.kind{{</hover >}}
-to define the _Custom Resource_ 
+to define the _Composite Resource_ 
 {{< hover label="xr2" line="3" >}}kind{{</hover>}}.
 * A {{< hover label="specGroup" line="8" >}}versions.schema{{</hover>}} section
-to define the _Custom Resource_ {{<hover label="xr2" line="6" >}}spec{{</hover >}}.
+to define the _Composite Resource_ {{<hover label="xr2" line="6" >}}spec{{</hover >}}.
 
 ```yaml {label="specGroup"}
 # Composite Resource Definition (XRD)
@@ -322,19 +322,20 @@ spec:
             - size  
 ```
 
-A _Custom Resource Definition_ can define a wide variety of settings and options. 
+A _Composite Resource Definition_ can define a wide variety of settings and
+options. 
 
-Creating a _Custom Resource Definition_ enables the creation of _Custom
-Resources_ but can also create a _Claim_.
+Creating a _Composite Resource Definition_ enables the creation of _Composite
+Resources_. It can also enable the creation of _Claims_.
 
-_Custom Resource Definitions_ with a `spec.claimNames` allow developers to
+_Composite Resource Definitions_ with a `spec.claimNames` allow developers to
 create _Claims_.
 
 For example, the 
 {{< hover label="xrdClaim" line="6" >}}claimNames.kind{{</hover >}}
 allows the creation of _Claims_ of `kind: computeClaim`.
 ```yaml {label="xrdClaim"}
-# Custom Resource Definition (XRD)
+# Composite Resource Definition (XRD)
 spec:
   group: test.example.org
   names:
@@ -347,11 +348,11 @@ spec:
 ## Claims
 _Claims_ are the primary way developers interact with Crossplane. 
 
-_Claims_ access the custom APIs defined by the platform team in a _Custom
+_Claims_ access the custom APIs defined by the platform team in a _Composite
 Resource Definition_.
 
-_Claims_ look like _Custom Resources_, but they're namespace scoped,
-while _Custom Resources_ are cluster scoped. 
+_Claims_ look like _Composite Resources_, but they're namespace scoped,
+while _Composite Resources_ are cluster scoped. 
 
 {{< hint "note" >}}
 **Why does namespace scope matter?**  
@@ -359,7 +360,7 @@ Having namespace scoped _Claims_ allows multiple teams, using unique namespaces,
 to create the same types of resources, independent of each other. The compute
 resources of team-A are unique to the compute resources of team-B.
 
-Directly creating _Custom Resources_ requires cluster-wide permissions,
+Directly creating _Composite Resources_ requires cluster-wide permissions,
 shared with all teams.   
 _Claims_ create the same set of resources, but on a namespace level.
 {{< /hint >}}
@@ -373,7 +374,7 @@ Claims use the same
 defined in _Composite Resource Definition_ and also used by 
 _Composite Resources_.
 ```yaml {label="xrdClaim2"}
-# Custom Resource Definition (XRD)
+# Composite Resource Definition (XRD)
 spec:
   group: test.example.org
   names:
@@ -386,10 +387,10 @@ spec:
 In an example _Claim_ the 
 {{<hover label="claim" line="2">}}apiVersion{{< /hover >}}
 matches the {{<hover label="xrdClaim2" line="3">}}group{{< /hover >}} in the
-_Custom Resource Definition_. 
+_Composite Resource Definition_. 
 
 The _Claim_ {{<hover label="claim" line="3">}}kind{{< /hover >}} matches the
-_Custom Resource Definition_ 
+_Composite Resource Definition_ 
 {{<hover label="xrdClaim2" line="7">}}claimNames.kind{{< /hover >}}.
 
 ```yaml {label="claim"}
@@ -404,8 +405,8 @@ spec:
 ```
 
 A _Claim_ can define a {{<hover label="claim" line="6">}}namespace{{</hover >}}.  
-The _Custom Resource Definition_ defines the 
-{{<hover label="claim" line="7">}}spec{{< /hover >}} like a _Custom Resource_.
+The _Composite Resource Definition_ defines the 
+{{<hover label="claim" line="7">}}spec{{< /hover >}} like a _Composite Resource_.
 
 
 _Claims_ are namespace scoped.
--- a/content/v1.11/getting-started/introduction.md
+++ b/content/v1.11/getting-started/introduction.md
@@ -172,9 +172,9 @@ Using the previous example, the platform team may set a compute resource size
 and virtual network settings. But the platform team allows users to define the 
 storage resource size.
 
-Creating a _Composition_ Crossplane doesn't create any managed
-resources. The _Composition_ is only a template for a collection of _managed
-resources_ and their settings. A _Composite Resource_ creates the specific resources.
+Creating a _Composition_ doesn't create any managed resources. The _Composition_
+is only a template for a collection of _managed resources_ and their settings. A
+_Composite Resource_ creates the specific resources.
 
 {{< hint "note" >}}
 The _[Composite Resources]({{<ref "#composite-resources">}})_ section discusses
--- a/content/v1.11/getting-started/introduction.md
+++ b/content/v1.11/getting-started/introduction.md
@@ -197,8 +197,8 @@ example, a _Composition_ template can create a compute, storage and networking
 set of _managed resources_. Crossplane uses the same _Composition_ template
 every time a user requests this set of resources.
 
-If a _Composition_ allows a user to define resource settings, users apply them
-in a _Composite Resource_.
+If a _Composite Resource Definition_ (`XRD`) allows a user to define resource
+settings, users apply them in a _Composite Resource_.
 
 
 <!-- A _Composition_ defines which _Composite Resources_ can use the _Composition_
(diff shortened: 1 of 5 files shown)


content/v1.11/getting-started/introduction.md
```diff
@@ -189,60 +189,15 @@ Use `kubectl get compositions` to view all _compositions_.
...
-{{<hover label="comp" line="7">}}apiVersion{{</hover >}} and the 
-_Composite Resource_  {{<hover label="xr" line="2">}}kind{{< /hover >}}
-matches the _Composition_ {{<hover label="comp" line="8">}}kind{{< /hover >}}.
-
-In this example, the _Composite Resource_ also sets the 
-{{<hover label="xr" line="7">}}storage{{< /hover >}} setting. The
-_Composition_ uses this value when creating the associated _managed resources_
-owned by this _Composite Resource_. -->
+If a _Composite Resource Definition_ (`XRD`) allows a user to specify resource
+settings, users apply them in a _Composite Resource_.
```
jeanduplessis, line 200:
```suggestion
If a _Composite Resource Definition_ (`XRD`) allows a user to specify resource
settings, users apply them in a _Composite Resource_ (`XR`).
```

**suggestion (non-blocking)**: Since we're clarifying the difference between Composite Resource Definition and Composite Resource it would make sense to me to also add the Composite Resource abbreviation as means to reinforce it to the reader.

jeanduplessis (approved):
Approving with one non-blocking suggestion.

content/v1.11/getting-started/provider-aws.md
```diff
@@ -1177,10 +1176,6 @@ EOF
 
 The {{< hover label="xr" line="3">}}apiVersion{{< /hover >}} and {{< hover label="xr" line="4">}}kind{{</hover >}} are from the provider's CRDs.
```
plumbis, line 1177:
```suggestion
The {{< hover label="xr" line="2">}}apiVersion{{< /hover >}} and {{< hover label="xr" line="3">}}kind{{</hover >}} are from the provider's CRDs.
```

content/v1.11/getting-started/provider-aws.md
```diff
@@ -1156,17 +1156,16 @@ The {{< hover label="providerconfig" line="11">}}spec.credentials.secretRef.name
...
 {{< /hint >}}
 
 ```yaml {label="xr"}
-bucket=$(echo "crossplane-bucket-"$(head -n 4096 /dev/urandom | openssl sha1 | tail -c 10))
 cat <<EOF | kubectl apply -f -
 apiVersion: s3.aws.upbound.io/v1beta1
 kind: Bucket
 metadata:
-  name: $bucket
+  generateName: crossplane-bucket-
```
plumbis, line 1168: What's the suffix to the generated resource? Will it look like the `kubectl get bucket` output below?

plumbis: Also is there an easy/deterministic way to provide a `kubectl delete` command for the generated name?

plumbis (approved):
This is great, thank you for taking the time to do a thorough review. A few small notes but nothing blocking.

Can you update the pages in `content/master/` (you can just copy the whole file) as well?

Also may be worth a skim through AWS guides part 2 and 3 to ensure I'm sticking to the YAML styles you called out. I'll write an update to the writing guide to specifically mention this guideline.

content/v1.11/getting-started/introduction.md
```diff
@@ -367,95 +322,94 @@ spec:
...
   size: "large"
 ```
 
-A _Claim_ can define a {{<hover label="claim" line="6">}}namespace{{</hover >}}.  
-The _Custom Resource Definition_ defines the 
-{{<hover label="claim" line="7">}}spec{{< /hover >}} like a _Custom Resource_.
-
-
-_Claims_ are namespace scoped.
+_Claims_ are namespace scoped. A _Claim_ must specify a {{<hover label="claim"
```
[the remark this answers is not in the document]
negz: They can. I added this because I thought you had to include `namespace: default` in your manifest to create something in the default namespace. I just double checked and that's not the case. If you omit `metadata.namespace` kubectl will create your thing in the default namespace.

_Technically_ this is something `kubectl` is doing for you locally behind the scenes. At the Kubernetes API level you do need to specify a namespace (even if it's `default`). That said, it's probably safe to revert this change.

content/v1.11/getting-started/provider-aws.md
negz: It will look quite similar - I think technically it's always a random 5 character suffix though. I think adding a label would be the best way to deterministically delete. I'll update.

negz:
> Also may be worth a skim through AWS guides part 2 and 3 to ensure I'm sticking to the YAML styles you called out. I opened https://github.com/crossplane/docs/issues/330 to make sure this is covered in the style guide.

Good call. I did that and found quite a few extra things. I think it's worth another review pass before we merge.

> Can you update the pages in content/master/ (you can just copy the whole file) as well?

Yes, I'll do that once we're sure the v1.11 content is good to go. I don't want to push the issue right now, but FWIW this workflow does feel like a regression to me in comparison to using git branches.

content/v1.11/getting-started/provider-aws-part-2.md
```diff
@@ -472,9 +473,8 @@ spec:
 ```
 
 {{< hint "tip" >}}
-An _XRD_ is a Kubernetes _custom resource definition_.
```
negz, line 475: I removed this because I interpreted it the first 2-3 times as saying an XRD and a CRD are the same thing. I later realized it was saying an XRD is defined by a CRD. This is true, but I suspect a little mind-bending. I figured we could gloss over this here, especially given `kubectl explain` is the better/recommended way to explore a schema than `kubectl describe crd`.

content/v1.11/getting-started/provider-aws-part-2.md
```diff
@@ -515,20 +515,30 @@ required:
 ```
 
 ### Enable claims to the API
-Allow a _claim_ to use this _XRD_ by defining the _claim_ API endpoint under the _XRD_ 
-{{<hover label="XRDclaim" line="4">}}spec{{< /hover >}}.
+Tell this _XRD_ to offer a _claim_ by defining the _claim_ API endpoint under
+the _XRD_ {{<hover label="XRDclaim" line="4">}}spec{{< /hover >}}.
```
negz, line 519: I'm not tied to this phrasing, but "allow a claim to use this XRD" doesn't seem accurate to me. It misses the context that the XRD is defining a type of claim into existence.

More broadly I think it's worth being aware of the different types of relationships we have and using more specific words than "uses". For example:

* XRD _defines_ XR
* XRD _defines_ XRC
* XRC _claims_ XR
* Composition _specifies_ MRs (maybe _templates_? I'm a little resistant to that word)
* XR _uses_ Composition (in this case "uses" does seem the most accurate to me)
* XR _composes_ MRs

content/v1.11/getting-started/provider-aws-part-2.md
```diff
@@ -365,10 +360,16 @@ metadata:
...
-    plural: databases
+    kind: XDatabase
+    plural: xdatabases
 ```
 
+{{< hint "tip" >}}
+This new API endpoint is a _Composite Resource_ (XR) API endpoint. It is
+convention for Composite Resource kinds to start with `X`. For example
+`XDatabase`. This is only convention and not required.
+{{< /hint >}}
```
negz, line 371: It may make sense to move (and the claim equivalent) this into introduction.md.

plumbis: I know that there is a practice of using X as a preface to defining the resource but I found it confusing as I was getting started. It is obscuring the thing we're talking about.

I have a custom API endpoint and I'm creating my definition of a database from it. That is straight forward. In the XRD and Composition concept chapters we can expand on this and explain why X.

content/v1.11/getting-started/provider-aws-part-2.md
plumbis: The point in the original text was to say "you can't have a claim if it's not defined in the XRD".  Maybe my mental model is wrong but I view an XRD as almost being an element with permissions. You can't access it via a claim unless the XRD _allows_ it.  

I'm open to changing it but "offers" doesn't feel like the right word here, that feels like a pub/sub thing, as if the XRD is producing an artifact (maybe it is, but it's not directly relevant to this part of the doc). I'm coming up blank on what to use an alternative. 
Tell this XRD to
 - allow a claim
 - allow the creation of claims 
 - create a claim type (this doesn't seem right) 
 - admit claims....something....

plumbis (approved):

plumbis:
> FWIW this workflow does feel like a regression to me in comparison to using git branches.

For editing net-new content between `latest` and `master` I totally agree. For normal care and feeding the versions tend to drift as extensions and enhancements are applied to new versions making them slightly different from older versions. When it's time to fix a typo it's much easier to find and replace in a single repo/directory to fix it everywhere. My personal experience with backporting docs fixes in the old method was not fun.

A secondary question to raise is "should we have docs for master?" I'm not sure we provide much value to users by having master docs. They are drafts for the `v.NEXT` more than anything. Instead of having master docs I'd rather carve off a new `v.NEXT` folder and have that be the new draft space. Probably worth it to take this conversation outside of this PR.

content/v1.11/getting-started/provider-aws-part-2.md
negz: Let me try a dumb analogy - I'm at the beach making a sandcastle. The previous phrasing reads to me as equivalent to "Allow a sandcastle to use Nic to create it".

Where _offer_ comes from is the idea that:

* The platform team is defining a new API type.
* By default is not available to developers to claim. As an XR (only) it's private to the platform team.
* The platform team may choose to offer the API type to their end users. They do this by offering a claim.
* Offering a claim is done by specifying `spec.claimNames` on an XRD.

My preference would be to stick with the concept of "offering a claim". This is an established concept that permeated our previous documentation and shows up in Kubernetes events, errors, etc pertaining to the concept. e.g. https://github.com/search?q=org%3Acrossplane+offer&type=code

content/v1.11/getting-started/provider-aws-part-2.md
negz: Saying this convention is too confusing for the getting started guide but okay to introduce later doesn't sit well with me. If it's confusing it shouldn't exist. I'm also wary for the first experience most folks have with these concepts to differ from most other examples. All of Upbound's reference platforms use this pattern for example - https://marketplace.upbound.io/configurations/upbound/platform-ref-aws/v0.5.0.

negz: > I have a custom API endpoint and I'm creating my definition of a database from it. That is straight forward.

I'm not following this rationale at all, sorry.

Part of the reason we use the `X` convention is that, Crossplane concepts aside, the XR and the XRC represent the same thing. They're both API abstractions that represent "a database". "Database" and "CustomDatabase" doesn't make sense to me - one is not more custom than the other.

In an ideal world we might have both XR and XRC be of `kind: Database` with some other kind of differentiator, but Kubernetes makes that difficult. Early on we considered differentiating them by having the XR be `kind: Database` and the XRC be `kind: DatabaseClaim`. @bassam pointed out that that pattern is not ideal because the developers using this system probably just want to think about the claim as "my database", not "my database claim". The distinction between XRC and XR is more relevant to the platform engineers configuring Crossplane. So where we landed is that the claim gets `kind: Database` and the XR (which from the developer perspective is just "weird Crossplane machinery") gets the `kind: XDatabase` differentiator.

content/v1.11/getting-started/provider-aws-part-2.md
plumbis:
```suggestion
Tell this _XRD_ to offer developers a _claim_ by defining the _claim_ API endpoint under
the _XRD_ {{<hover label="XRDclaim" line="4">}}spec{{< /hover >}}.
```

The offer idea breakdown is super helpful. 

Let's just specific _who_ the claim is being offered to. It reads as if it's a Kubernets object offering something to another object. Can we just add "developers" to it's clear who the claim is being offered to? I'm sorry if this is too nit picky, I'm trying to write for someone like me who has little background knowledge on Crossplane and I'm trying to hammer out all of the assumptions or implied meanings.

content/v1.11/getting-started/provider-aws-part-2.md
plumbis: I'm not a fan of the convention in general but I'm the new guy. I've struggled with understanding who has the `database` and who has the `Xdatabase`. I'd prefer to using more unique names or indicating who is creating the resource, like `compositeDatabase`

I don't see a problem changing the terms since a guide like this is about introducing the minimum number of steps/topics and not about best practices. The goal here is for a reader to understand each `kind` and what is' responsible for. From there we can introduce more complex topics and best practices or standards.

Regarding the reference platforms, they are great if you know Crossplane. If you're new it's very hard to connect the dots. I don't see a problem keeping them in the reference platforms.

negz:
@plumbis Do you feel okay merging this PR and following up on our open threads separately? I'm assuming yes because I see you re-approved the PR but I want to make sure.

plumbis:
Yeah, no objections.

PR #351, opened 2023-02-09
Correct xfn sidercar to container rather than pod

hasheddan:
Updates docs to state that the xfn function runner is deployed as a sidecar container rather than a pod.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>


negz (approved):
Thanks!

PR #358, opened 2023-02-12
Initial Vale style guidelines

### LINKED ISSUE #339: Add Vale linting - non-blocking
The docs should enable Vale linting to increase quality and provide style enforcement. The Vale definitions will be based on the [Upbound/vale](https://github.com/upbound/vale) repo. 

The Vale linting should be enabled and flag any violations but will not block PRs. 

After the [v1.12 release](https://github.com/crossplane/crossplane#releases) on April 25 we can consider requiring Vale passing for any PRs.

plumbis:
This commit adds the initial support for Vale in the docs repo. 

Vale is enabled as a Github Action for PRs. Only the changed files are checked, since most of the docs fail Vale testing today.

The contributing guide has been expanded to explain some of the style guide rules and how to install and use Vale.

The initial Vale style guides added are:
* [Alex](https://github.com/errata-ai/alex) - for insensitive, inconsiderate writing.
* [GitLab](https://gitlab.com/gitlab-org/gitlab/-/tree/master/doc/.vale/gitlab)
  - for the [GitLab documentation style guide](https://docs.gitlab.com/ee/development/documentation/styleguide/).
* [Google](https://github.com/errata-ai/google) - for the [Google developer documentation style guide](https://developers.google.com/style).
* [Microsoft](https://github.com/errata-ai/Microsoft) - for the [Microsoft style guide](https://learn.microsoft.com/en-us/style-guide/welcome/).
* [proselint](https://github.com/errata-ai/proselint) - for higher quality writing.
* [Upbound](https://github.com/upbound/vale/tree/main/styles/Upbound) - for
  customized terms and preventing corporate jargon.
* [write-good](https://github.com/errata-ai/write-good) - for higher quality writing.

There are likely duplicate rules that need to be cleaned up, mainly across Microsoft and Google style guides.

As mentioned in #339 failing Vale tests won't be PR blockers yet but it should be a bar for us to try and meet for all new contributions. 

Resolves #339 


.github/workflows/vale.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  pull_request:
+    paths:
+      - 'content/**'
+
+  # Allows manual workflow run (must in default branch to work)
+  workflow_dispatch:
+
+jobs:
+  build:
```
jeanduplessis, line 12:
```suggestion
  lint-content:
```
Non-blocking: `build` as a name for the job feels out of place in this context.

.github/workflows/vale.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Install Vale
+        env:
+          VALE_VERSION: "2.22.0"
+        run: wget https://github.com/errata-ai/vale/releases/download/v${{env.VALE_VERSION}}/vale_${{env.VALE_VERSION}}_Linux_64-bit.tar.gz && tar -xvzf vale_${{env.VALE_VERSION}}_Linux_64-bit.tar.gz
+
+      - name: Run Vale
+        if: steps.changed-files.outputs.any_changed == 'true'
+        # Run vale and display the output. Fail if the output is greater than 1 line.
+        # Vale uses a 0 return code for warnings and suggestions. These should be considered failures.
```
jeanduplessis, line 32: Non-blocking: Are we always going to treat suggestions as failures? That feels a bit harsh for when we move to make this a mandatory step that has to pass for a PR.

content/contribute/writing-style-guide.md
```diff
@@ -4,33 +4,118 @@ weight: 10
...
+  name: my-resource
+```
 
-Beyond Vale, Crossplane recommends [Grammarly](https://www.grammarly.com/) and [Hemingway
-Editor](https://hemingwayapp.com/) to improve writing quality.
+### Inline
 
-The Crossplane maintainers recommend the [Google developer documentation style guide](https://developers.google.com/style) 
-to help answer questions on proper styling. 
+Kubernetes objects mentioned inline don't require styling unless drawing special
```
jeanduplessis, line 120:
```suggestion
<!-- vale off -->
<!-- turn off vale checking for this line as special doesn't refer to a person. [alex.Ablist] -->
Kubernetes objects mentioned inline don't require styling unless drawing special
<!-- vale on -->


```

jeanduplessis: I would also consider whether we should remove the rule for `special`, since it is highly unlikely we would reference people with disabilities in our documentation.

jeanduplessis (changes requested):
Thanks for getting this up and running @plumbis. 
Just one minor change request as it would be nice to merge this without Vale failing on this PR 😉

.github/workflows/vale.yml
plumbis: Shh, you'll expose the fact I don't know what I'm doing and just copy/pasta things to get them to work ;) 

Will fix.

content/contribute/writing-style-guide.md
plumbis: +1. Committing a fix to remove that check. This is exactly the kind of "figuring out the guide/checks" we need to work though.

.github/workflows/vale.yml
plumbis: The problem is that we are inheriting the warning/suggestion/error flags from the upstream style libraries. I think it would be significant effort to apply our editorial opinions to them.

What I'd rather do is let any vale warn/suggest/error be a decision point to either:
* update the rules/style (e.g., "special")
* improve the writing (e.g., using latin) 
* Set a vale exception for that line/paragraph to ignore the rule.

This also keeps the results clean. I worry about a future where vale runs are 1400 suggestions creating noise against the real errors and problems. 

If it becomes a thing™ in the future I'd be happy to reevaluate error levels.

content/contribute/writing-style-guide.md
plumbis: Fixed in style update. aa43e183d282b2388a794b9775655c049591f2cb

Confirmed via updated CI run.

.github/workflows/vale.yml
jeanduplessis: This sounds like a sound approach to me.

jeanduplessis (approved):

jbw976 (commented):
Thank you @plumbis for taking more steps to improve (and enforce) the quality of our docs!

Maintenance question for you: when we find rules that don't work for our project, how does the removal or ignoring of those rules work? I'm mostly curious from a long term maintenance perspective of our rule sets - e.g. we remove a rule that we don't like, then google updates their rule set that we are using.  Does that cause conflicts or make it hard to update our rulesets over time? 🤔

plumbis:
Good news, bad news @jbw976.

Bad news is that we're maintaining forks of all the styles now. This is a hassle of course. I did a bunch of experiments trying to use them as submodules or using them unmodified locally so we can just drop-in update them but there wasn't enough flexibility in customizing the rules in a user friendly way.

The good news is most of these aren't updated that often and over time we'll end up diverging anyway as we refine our own styles. I'm not too worried about the overhead this creates.

Regarding specifically managing rules there are three things:
1. Disabling a rule entirely.
To do this we edit the .vale.ini file and turn the specific rule off, like this:
https://github.com/crossplane/docs/pull/358/files#diff-1274633810a31af03dde0fa78893f8cb17e4eb416735ff19488805375ac35bb5R10

2. Modifying a rule. 
This is something like adding a permanent exception or removing a style that doesn't fit with our writing. For those the "style" files are modified directly. For example commit [`aa43e18`](https://github.com/crossplane/docs/pull/358/commits/aa43e183d282b2388a794b9775655c049591f2cb) removes "special" from the word list, since it is unlikely to be used in a negative context, given our content. For universal things like `Crossplane` that should never fail a test they are added to the spelling.yml file here: https://github.com/crossplane/docs/pull/358/files#diff-9be804b9956ca9ceba77ba60ed6f2ac2c8ff5c38caf05ca29bec5108d6e179feR5

3. Disabling a rule for just a line of text.
There are lots of valid reasons we'd want to disable something when writing, where the rule is good but the writing creates an exception that needs to be ignored. For that we can use `<!-- vale off -->` or `<!-- vale <style> = NO` remarks. This is [described here](https://github.com/crossplane/docs/pull/358/files#diff-7fb2a473ee291bd77198c0536794da331528a3216c42eddfeacd9b7626502565R53)

jbw976:
Awesome @plumbis, thanks for those pointers on how to tweak the rules to our needs, and your assessment of the ongoing maintenance burden/risk!

PR #369, opened 2023-02-27
GCP Quickstart Update

plumbis:
Updates the GCP quickstart to follow the same 3 part series that the AWS quickstart follows.

This PR also includes Vale updates for GCP related terms.


jeanduplessis (approved):
I'm approving from a "this works as expected" pov, not commenting on the technical accuracy of the content.

Feedback (non-blocking): I find the "Skip part... and just get started" callout ambiguous and attention-grabbing. Let me explain:
- At first, it didn't dawn on me that I was getting the TL;DR version of what you needed to do in parts 1 and 2. A slight wording change here to make it more explicit what the section is for could be helpful. Potentially, "Follow these quick steps to bring you up to speed on the first two parts before continuing."
- Because the section is in white (in dark mode, at least), it's the first thing that grabs your attention and can easily distract you from reading the context around it. I believe this caused me not to understand what this section was for immediately.

CleanShot 2023-03-13 at 15 59 24@2x

PR #374, opened 2023-03-01
Script to clear or reset span contents for editable elements

### LINKED ISSUE #329: Clear editable fields on click
Suggested by @jeanduplessis in https://github.com/crossplane/docs/pull/311#issuecomment-1413458680

The new `contenteditable` field would be improved it the contents cleared on click. Bonus points if we restore the contents if nothing is entered.

plumbis:
Resolves #329 

When clicking on an editable field, if the contents are the placeholder text, the contents are cleared.

If you click away and the contents are blank the placeholder text is restored.

To support this the placeholder text is added as a `data-` element in the span. 

 
329-demo

Demo: https://deploy-preview-374--crossplane.netlify.app/v1.11/getting-started/provider-aws/#generate-an-aws-key-pair-file

### DIFF
--- a/themes/geekboot/assets/scss/_code-theme-base.scss
+++ b/themes/geekboot/assets/scss/_code-theme-base.scss
@@ -16,6 +16,15 @@
     margin-top: 1rem; // prevent boxes from running together
     margin-bottom: 1rem;
 
+    [contentEditable=true]:empty:before{
+      content:attr(data-default);
+    }
+
+
+    [contentEditable=true]:empty:focus:before{
+      color:#ddd;
+    }
+
     :target{
       margin-top: 0;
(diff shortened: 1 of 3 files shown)


jeanduplessis:
@plumbis how would you feel about not removing the text on focus, and rather fading it out a bit so that the user still has context as to what value is expected, rather than just seeing an empty space. It will also help with the jumping that happens on focus currently.

Example: https://jsfiddle.net/8dfr73bc/2/

plumbis:
When you include a jsfiddle, how do I say no? I think that's much better way to do this. I also had no idea most of those CSS elements/selectors existed!

jeanduplessis (approved):
LGTM 🚀

PR #378, opened 2023-03-06
Add notices to top of all v1.10 pages that moved after PR #271

plumbis:
This adds an `important` notice to all pages that were moved from v1.10 to v1.11 with the reorg in PR #271 

For example:

Screenshot 2023-03-06 at 10 54 47 AM

Vale failures are expected since this is old content.

Resolves #377

PR #379, opened 2023-03-06
Fix dark mode stylings

plumbis:
In #378 @jeanduplessis called attention to a day-1 bug in dark mode where dark mode toggles and some stylings don't work properly.

This is attributed to two things:
1. Some color-mode variables are not defined in both light and dark themes causing some browsers to not render them.
2. The Javascript toggle didn't properly save/apply the color theme when it didn't match the system color mode.

## The Fix
* Match all variable names in both light and dark modes
* Remove the `@media` check for light/dark mode and move it into JavaScript. This allows for deterministic behavior between a system color mode preference and the toggle preference.
* Allow a toggle selection to override system defaults (i.e., light mode docs with a dark mode system)


jeanduplessis (approved):

jeanduplessis:
@plumbis it works great now!

PR #360, opened 2023-02-15
AWS Quickstart Updates

plumbis:
Additional xrd fixes for #356.

Also adds updated vale standards. The Crossplane/Headings file was removed as it's a duplicate of the Google/Headings style definitions. 

Signed-off-by: Pete Lumbis <pete@upbound.io>

PR #381, opened 2023-03-09
Corrected few typo errors.

basil1987:
Wrongly Mentioned "compositions" in place of "composite resources / xdatabase". Corrected it now. Please verify and merge.


plumbis:
Thank you for your PR! This is fixed in #360 which is currently in review. 

I'll close this one out to prevent conflicts but thanks for filing the issue.

PR #389, opened 2023-03-31
Add missing and remove duplicate link definitions

### LINKED ISSUE #387: [Web Bug] - Composite Resources
Broken link for Composition Reference

If you’re already familiar with Composite Resources and looking for a detailed configuration reference or some tips, tricks, and troubleshooting information, try the **[Composition Reference][xr-ref]**.

URL: /v1.11/concepts/composition/

jeanduplessis:
Fixes #387 

Vale errors are expected and not related to the changes in this PR.


content/master/concepts/composition.md
```diff
@@ -1250,7 +1249,8 @@ so:
 [issue-2024]: https://github.com/crossplane/crossplane/issues/2024
 [xrs-and-mrs]: /media/composition-xrs-and-mrs.svg
 [how-it-works]: /media/composition-how-it-works.svg
-[crd-docs]: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/
 [provider-kubernetes]: https://marketplace.upbound.io/providers/crossplane-contrib/provider-kubernetes
 [provider-helm]: https://marketplace.upbound.io/providers/crossplane-contrib/provider-helm/
 [claims-and-xrs]: /media/composition-claims-and-xrs.svg
+[xr-ref]: {{<ref "../getting-started/introduction" >}}
```
jbw976, line 1255: i feel like this isn't the right target for what this link intends.  It takes the reader to the generic introduction page, but it intends to take you to the details composition reference.

The tricky part is that I would argue we're already on the composition reference page 🤪 - all the detailed info and config options etc. are on this very page.

Perhaps the best link here would instead be to `/concepts/composition/#compositions`, which is the specific section that goes through every composition field, type, config option?  What do you think?

content/v1.11/concepts/composition.md
```diff
@@ -1251,7 +1251,7 @@ so:
 [issue-2024]: https://github.com/crossplane/crossplane/issues/2024
 [xrs-and-mrs]: /media/composition-xrs-and-mrs.svg
 [how-it-works]: /media/composition-how-it-works.svg
-[crd-docs]: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/
 [provider-kubernetes]: https://marketplace.upbound.io/providers/crossplane-contrib/provider-kubernetes
 [provider-helm]: https://marketplace.upbound.io/providers/crossplane-contrib/provider-helm/
 [claims-and-xrs]: /media/composition-claims-and-xrs.svg
+[xr-ref]: {{<ref "../getting-started/introduction" >}}
```
jbw976, line 1257: same issue here for the "composition reference" link, I don't think the best target for that is the generic intro page.

content/master/concepts/composition.md
jbw976: and if we do then link to a section that is on this same page, we might want to change the language from "reference page" to something more like "reference section"

jbw976 (commented):
Thanks for making this clean up pass @jeanduplessis!! 

Did anything else need to be cleaned up for v1.9 also?

content/master/concepts/composition.md
jeanduplessis: I've updated the link to the suggested sub section.

jbw976 (approved):
Alrighty sweet, I like the link we've landed on, I think it lands the reader at a more relevant target.  Thanks @jeanduplessis! 💪

PR #391, opened 2023-04-10
Math Transforms: Document `ClampMin` and `ClampMax`

julienduchesne:
Feature added in this PR: https://github.com/crossplane/crossplane/pull/3917

### DIFF
--- a/content/master/concepts/composition.md
+++ b/content/master/concepts/composition.md
@@ -905,8 +905,8 @@ transform.
 
 `math`. Transforms values using math. The input value must be an integer.
 * math transform type `Multiply`, multiplies the input by the given value.
-* math transform type `ClampMin`, sets a minimum value for the input.
-* math transform type `ClampMax`, sets a maximum value for the input.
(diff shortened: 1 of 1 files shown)


content/master/concepts/composition.md
```diff
@@ -904,14 +904,39 @@ transform.
 ```
 
 `math`. Transforms values using math. The input value must be an integer.
-Currently only `multiply` is supported.
+* math transform type `Multiply`, multiplies the input by the given value.
+* math transform type `ClampMin`, sets a minimum value for the input.
+* math transform type `ClampMax`, sets a maximum value for the input.
```
turkenh, line 909:
```suggestion
* math transform type `ClampMax`, sets a maximum value for the output.
```

Would `output` make more sense?

turkenh (approved):

content/master/concepts/composition.md
julienduchesne: Done!

PR #402, opened 2023-04-17
document feature flags available

### LINKED ISSUE #397: Centralised available feature flags docs
### What's Missing?

Currently there is no centralised place where users can discover the available feature flag and their status, e.g. `--enable-environment-config`, this information is spread across the guides or is only available as a help message for the `crossplane` binary. I feel it should be easier to discover, so that users can start using them more broadly and provide valuable feedback to foster features' development.

Below for the current supported flags and their status at the time I'm writing this:
```shell
$ crossplane core start --help

Usage: crossplane core start

Start Crossplane controllers.

Flags:
  -h, --help                                                  Show context-sensitive help.
  -d, --debug                                                 Print verbose logging statements.
  -v, --version                                               Print version and quit.

  -n, --namespace="crossplane-system"                         Namespace used to unpack and run packages ($POD_NAMESPACE).
      --service-account="crossplane"                          Name of the Crossplane Service Account ($POD_SERVICE_ACCOUNT).
  -c, --cache-dir="/cache"                                    Directory used for caching package images ($CACHE_DIR).
  -l, --leader-election                                       Use leader election for the controller manager ($LEADER_ELECTION).
  -r, --registry="index.docker.io"                            Default registry used to fetch packages when not specified in tag ($REGISTRY).
      --ca-bundle-path=STRING                                 Additional CA bundle to use when fetching packages from registry ($CA_BUNDLE_PATH).
      --webhook-tls-secret-name=STRING                        The name of the TLS Secret that will be used by the webhook servers of core Crossplane and providers ($WEBHOOK_TLS_SECRET_NAME).
      --webhook-tls-cert-dir=STRING                           The directory of TLS certificate that will be used by the webhook server of core Crossplane. There should be tls.crt and tls.key files ($WEBHOOK_TLS_CERT_DIR).
      --user-agent="crossplane/v1.12.0-rc.0.292.gfb52077d"    The User-Agent header that will be set on all package requests ($USER_AGENT).
  -s, --sync-interval=1h                                      How often all resources will be double-checked for drift from the desired state.
      --poll-interval=1m                                      How often individual resources will be checked for drift from the desired state.
      --max-reconcile-rate=10                                 The global maximum rate per second at which resources may checked for drift from the desired state.
      --esstls-secret-name=STRING                             The name of the TLS Secret that will be used by Crossplane and providers as clients of External Secret Store plugins ($ESS_TLS_SECRET_NAME).
      --esstls-certs-dir=STRING                               The path of the folder which will store TLS certificates to be used by Crossplane and providers for communicating with External Secret Store plugins ($ESS_TLS_CERTS_DIR).

Beta Features:
  --enable-composition-revisions    Enable support for CompositionRevisions.

Alpha Features:
  --enable-environment-configs                      Enable support for EnvironmentConfigs.
  --enable-external-secret-stores                   Enable support for External Secret Stores.
  --enable-composition-functions                    Enable support for Composition Functions.
  --enable-composition-webhook-schema-validation    Enable Composition validation using schemas.
```

phisco:
Fixes #397

### DIFF
--- a/content/master/software/install.md
+++ b/content/master/software/install.md
@@ -263,7 +263,7 @@ or by looking at the table below.
 {{< /table >}}
 {{< /expand >}}
 
-These can be set either in the `values.yaml` file or at install time using the
+Set these flags either in the `values.yaml` file or at install time using the
 `--set` flag, for example: `--set
 args=["--enable-composition-functions","--enable-composition-webhook-schema-validation"]`.
 
--- a/content/master/software/install.md
+++ b/content/master/software/install.md
@@ -264,7 +264,7 @@ or by looking at the table below.
 {{< /expand >}}
 
 These can be set either in the `values.yaml` file or at install time using the
-`--set` flag, e.g. `--set
+`--set` flag, for example: `--set
 args=["--enable-composition-functions","--enable-composition-webhook-schema-validation"]`.
 
 ### Install pre-release Crossplane versions
--- a/content/master/software/install.md
+++ b/content/master/software/install.md
@@ -243,6 +243,30 @@ crossplane-stable/crossplane \
 -f settings.yaml
 ```
 
+#### Feature flags
+
+Crossplane usually introduces new features behind feature flags. Alpha features
+are disabled by default. Beta features are enabled by default. To enable or
+disable a feature flag, set the `args` value in the Helm chart. Available
+feature flags can be directly found by running `crossplane core start --help`,
+or by looking at the table below.
+
+{{< expand "Feature flags" "v" >}}
+{{< table caption="Feature flags" >}}
+| Status | Flag | Description |
+| --- | --- | --- |
+| Beta | `--enable-composition-revisions` |Enable support for CompositionRevisions |
+| Alpha | `--enable-composition-functions` | Enable support for Composition Functions. |
+| Alpha | `--enable-composition-webhook-schema-validation` | Enable Composition validation using schemas.
+| Alpha | `--enable-environment-configs` | Enable support for EnvironmentConfigs. |
+| Alpha | `--enable-external-secret-stores` | Enable support for External Secret Stores. |
+{{< /table >}}
+{{< /expand >}}
+
+These can be set either in the `values.yaml` file or at install time using the
+`--set` flag, e.g. `--set
+args=["--enable-composition-functions","--enable-composition-webhook-schema-validation"]`.
+
 ### Install pre-release Crossplane versions
 Install a pre-release versions of Crossplane from the `master` Crossplane Helm channel.
 
--- a/content/master/software/install.md
+++ b/content/master/software/install.md
@@ -245,19 +245,19 @@ crossplane-stable/crossplane \
 
 #### Feature flags
 
-Crossplane usually introduces new features behind feature flags. Alpha features
-are disabled by default. Beta features are enabled by default. To enable or
-disable a feature flag, set the `args` value in the Helm chart. Available
-feature flags can be directly found by running `crossplane core start --help`,
-or by looking at the table below.
+Crossplane usually introduces new features behind feature flags. By default
+alpha features are off, while beta features are enabled. To enable a feature
+flag, set the `args` value in the Helm chart. Available feature flags can be
+directly found by running `crossplane core start --help`, or by looking at the
+table below.
 
-{{< expand "Feature flags" "v" >}}
+{{< expand "Feature flags" >}}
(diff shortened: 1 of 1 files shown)


jeanduplessis (approved):

content/master/software/install.md
```diff
@@ -243,6 +243,30 @@ crossplane-stable/crossplane \
 -f settings.yaml
 ```
 
+#### Feature flags
+
+Crossplane usually introduces new features behind feature flags. Alpha features
+are disabled by default. Beta features are enabled by default. To enable or
+disable a feature flag, set the `args` value in the Helm chart. Available
```
plumbis, line 250:
```suggestion
turn off a feature flag, set the `args` value in the Helm chart. Available
```

content/master/software/install.md
```diff
@@ -243,6 +243,30 @@ crossplane-stable/crossplane \
 -f settings.yaml
 ```
 
+#### Feature flags
+
+Crossplane usually introduces new features behind feature flags. Alpha features
```
plumbis, line 248:
```suggestion
Alpha and beta features are behind feature flags. Alpha features
```

content/master/software/install.md
```diff
@@ -243,6 +243,30 @@ crossplane-stable/crossplane \
...
+| --- | --- | --- |
+| Beta | `--enable-composition-revisions` |Enable support for CompositionRevisions |
+| Alpha | `--enable-composition-functions` | Enable support for Composition Functions. |
+| Alpha | `--enable-composition-webhook-schema-validation` | Enable Composition validation using schemas.
+| Alpha | `--enable-environment-configs` | Enable support for EnvironmentConfigs. |
+| Alpha | `--enable-external-secret-stores` | Enable support for External Secret Stores. |
+{{< /table >}}
+{{< /expand >}}
+
+These can be set either in the `values.yaml` file or at install time using the
```
plumbis, line 266:
```suggestion
Set these flags either in the `values.yaml` file or at install time using the
```

plumbis (changes requested):
Small edits to remove passive voice and `disabled` to pass the Vale checks. 

Thanks for the contribution!

plumbis:
> we might want to consider removing the Vale rule for disabled.

@jeanduplessis I went back and forth on this one, even opening a [PR](https://github.com/crossplane/docs/pull/404) that I closed a few minutes later.

Although I think we can justify our use of the word `disabled` we can just as easily use `turn off`. I'd rather err on the side of inclusive language.

content/master/software/install.md
```diff
@@ -243,6 +243,30 @@ crossplane-stable/crossplane \
 -f settings.yaml
 ```
 
+#### Feature flags
+
+Crossplane usually introduces new features behind feature flags. Alpha features
+are disabled by default. Beta features are enabled by default. To enable or
+disable a feature flag, set the `args` value in the Helm chart. Available
+feature flags can be directly found by running `crossplane core start --help`,
+or by looking at the table below.
```
phisco, line 252:
```suggestion
Crossplane usually introduces new features behind feature flags. By default
alpha features are off, while beta features are enabled. To enable a feature
flag, set the `args` value in the Helm chart. Available feature flags can be
directly found by running `crossplane core start --help`, or by looking at the
table below.
```

jeanduplessis:
I'm okay with disable = turn off. However, it reads funny when we still use enable. 

So I turn off something and then enable it? That feels wrong.

plumbis:
> So I turn off something and then enable it? That feels wrong.

@jeanduplessis I agree. Would you support adding a rule suggesting `turn on` instead of `enable`? Any commands that are `--enable` would be exempt.

For example,

"Turn on with `--enable`. Blah is turned off by default"

jeanduplessis:
@plumbis I'm wondering if we are going too far here. If I take a step back and look at "acceptable" language in engineering, I don't see folks objecting to enable/disable as a language choice in the same way that things like master and whitelist/blacklist are objected to. The rule in Vale also clearly speaks to the subject being a person, not an inanimate/abstract concept. 

It is highly unlikely that we would be talking about people in the context of a disability in our technical documentation. The one area I could imagine it being relevant is in our contribution guidelines to give guidance on making content accessible for those with a disability. 

That said, it's such a small part of the potential use case that I favor disabling the rule and relying on the maintainers of this repo to be vigilant when reviewing PRs.

content/master/software/install.md
phisco: pushed this, let me know if there is any other change needed @plumbis

plumbis (approved):

PR #405, opened 2023-04-17
Add more information about vale error and warning policy

plumbis:
Adds information to the Vale section of the docs contrib guide to call out:

1.) We're in the process of editing content to pass vale and errors in old content are okay, but we want vale to pass for new content.

2.) Warnings == Errors


phisco:
My two cents: I find Vale to be too pedantic, putting too much emphasis on minor details instead of the actual documentation's meaning.

jeanduplessis (approved):

PR #412, opened 2023-04-24
New page on crossplane pods

plumbis:
Creates a new concept page covering the crossplane core pod and rbac manager pod.

### DIFF
--- a/content/master/concepts/pods.md
+++ b/content/master/concepts/pods.md
@@ -79,7 +79,7 @@ Crossplane monitors resources through a Kubernetes
 _[watch](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes)_
 or through periodic polling. Some resources may be both watched and polled. 
 
-Crossplane requests that Kubernetes notifies Crossplane of any changes on
+Crossplane requests that the API server notifies Crossplane of any changes on
 objects. This notification tool is a _watch_. 
 
 Watched objects include Providers, managed resources and 
@@ -105,10 +105,10 @@ Managed resources use polling.
 {{< hint "note" >}}
 Managed resources watch for Kubernetes events like deletion or changes to
 their `spec`. Managed resources rely on polling to detect changes in the
-provider environment.
+external system.
 {{< /hint >}}
 
-Crossplane checks all resources to
+Crossplane double-checks all resources to
 confirm they're in the desired state. Crossplane does this every one hour by
 default. Use the `--sync-interval` Crossplane pod argument to change this
 interval. 
@@ -141,13 +141,10 @@ implements their own max reconcile rate setting.
 
 ##### Number of reconcilers
 
-The second value `--max-reconcile-rate` defines is the number of threads
-created to reconcile resources. Each thread manages the reconciliation of a
-specific resource. If there are more resources than configured threads,
-remaining resources must wait until a new thread is available. 
-
-Changing `--max-reconcile-rate` changes the number of threads Crossplane creates
-when more than one resource need reconciliation.
+The second value `--max-reconcile-rate` defines is the number of
+resources that can reconcile at once. If there are more resources than
+configured `--max-reconcile-rate` the remaining resources must wait until
+Crossplane reconciles a an existing resource.
 
 Read the [Change Pod Settings]({{<ref "#change-pod-settings">}}) section for
 instructions on applying these settings. 
@@ -229,7 +226,7 @@ The `crossplane-admin` ClusterRole has the following permissions:
 View the full RBAC policy with 
 
 ```shell
-kubectl describe clusterrole crossplane:aggregate-to-admin
+kubectl describe clusterrole crossplane:admin
 ```
 
 ##### crossplane-edit
@@ -243,7 +240,7 @@ The `crossplane-edit` ClusterRole has the following permissions:
 View the full RBAC policy with 
 
 ```shell
-kubectl describe clusterrole crossplane:aggregate-to-edit
+kubectl describe clusterrole crossplane:edit
 ```
 
 ##### crossplane-view
@@ -256,7 +253,7 @@ The `crossplane-view` ClusterRole has the following permissions:
 View the full RBAC policy with 
 
 ```shell
-kubectl describe clusterrole crossplane:aggregate-to-view
+kubectl describe clusterrole crossplane:view
 ```
 
 ##### crossplane-browse
@@ -269,7 +266,7 @@ The `crossplane-browse` ClusterRole has the following permissions:
 View the full RBAC policy with 
 
 ```shell
-kubectl describe clusterrole crossplane:aggregate-to-browse
+kubectl describe clusterrole crossplane:browse
 ```
 
 #### Crossplane Roles
--- a/content/master/concepts/pods.md
+++ b/content/master/concepts/pods.md
@@ -28,10 +28,10 @@ packages with Crossplane, customizing the namespace Crossplane installs in and
 defining webhook configurations. 
 
 The core CRDs installed by the init container include: 
-* Composite Resource Definitions, Compositions, Configurations and Providers
-* `locks` to manage package dependencies
-* `controllerconfigs` to apply settings to installed Providers
-* `storeconfigs` for connecting external secret stores like 
+* CompositeResourceDefinitions, Compositions, Configurations and Providers
+* Locks to manage package dependencies
+* ControllerConfigs to apply settings to installed Providers
+* StoreConfigs for connecting external secret stores like 
 [HashiCorp Vault](https://www.vaultproject.io/)
 
 {{< hint "note" >}}
@@ -63,17 +63,17 @@ The main Crossplane container, called the _core_ container, enforces
 the desired state of Crossplane resources, manages leader elections and process
 webhooks. 
 
+{{<hint "note" >}}
+The Crossplane pod only reconciles core Crossplane components, including Claims
+and composite resources. Providers are responsible for reconciling their managed
+resources. 
+{{< /hint >}}
+
 #### Reconcile loop
 
 The core container operates on a _reconcile loop_, constantly checking the 
 status of deployed resources and correcting any "drift." After checking a 
 resource Crossplane waits some time and checks again.
-
-{{<hint "note" >}}
-The Crossplane pod only reconciles core Crossplane components, including claims
-and composite resources. Providers are responsible for reconciling their managed
-resources. 
-{{< /hint >}}
 
 Crossplane monitors resources through a Kubernetes 
 _[watch](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes)_
@@ -82,8 +82,8 @@ or through periodic polling. Some resources may be both watched and polled.
 Crossplane requests that Kubernetes notifies Crossplane of any changes on
 objects. This notification tool is a _watch_. 
 
-Watched objects include providers, managed resources and composite resource 
-definitions.
+Watched objects include Providers, managed resources and 
+CompositeResourceDefinitions.
 
 For objects that Kubernetes can't provide a watch for, Crossplane
 periodically poll the resource to find it's state. The default polling rate is
@@ -125,25 +125,19 @@ CPU resources Crossplane uses but allows Crossplane to reconcile all resources
 faster. 
 
 {{< hint "important" >}}
-Most providers use their own `--max-reconcile-rate`. This determines the
+Most Providers use their own `--max-reconcile-rate`. This determines the
 same settings for Providers and their managed resources. Applying the
 `--max-reconcile-rate` to Crossplane only controls the rate for
 core Crossplane resources. 
 {{< /hint >}}
 ##### Reconcile retry rate
 
-When Crossplane checks a resource and it's not in the desired
-state they 100 times per second. If the resource is still not in the desired 
-state Crossplane checks the `--max-reconcile-rate` times per second. By default,
-the `--max-reconcile-rate` value of 10 means checking 10 times per second. 
+The `--max-reconcile-rate` setting configures the number of times per second
+Crossplane or a provider attempts to correct a resource. The default value is 
+10 times per second.
 
-{{< hint "note" >}}
-Kubernetes defines the initial rate of 100 times per second. This isn't
-configurable.
-{{< /hint >}}
-
-All core Crossplane components share the retry rate. Each Provider
-implements their own retry rate. 
+All core Crossplane components share the reconcile rate. Each Provider
+implements their own max reconcile rate setting. 
 
 ##### Number of reconcilers
 
@@ -175,6 +169,20 @@ The
 provides more comprehensive details on the Crossplane RBAC requirements.
 {{< /hint >}}
 
+### Disable the RBAC manager 
+
+Disable the RBAC manager after installation by deleting the
+`crossplane-rbac-manager` deployment from the `crossplane-system` namespace.
+
+Disable the RBAC manager before installation by editing the Helm `values.yaml`
+file, setting `rbacManager.deploy` to `false`.
+
+{{< hint "note" >}}
+
+Instructions for changing Crossplane pod settings during installation are in the
+[Crossplane Install]({{<ref "../software/install">}}) section. 
+{{< /hint >}}
+
 <!-- vale Microsoft.HeadingAcronyms = NO -->
 <!-- allow 'RBAC' since that's the name -->
 ### RBAC init container
@@ -186,44 +194,99 @@ The RBAC manager requires the `CompositeResourceDefinition` and
 The RBAC manager init container waits for these resources before starting the 
 main RBAC manager container. 
 
-
-<!-- vale Microsoft.HeadingAcronyms = NO -->
-<!-- allow 'RBAC' since that's the name -->
 ### RBAC manager container
-<!-- vale Microsoft.HeadingAcronyms = YES -->
 
 The RBAC manager container preforms the following tasks:
-* creating and binding RBAC roles to Provider Service Accounts, allowing 
+* creating and binding RBAC roles to Provider ServiceAccounts, allowing 
   them to control their managed resources
-* allowing the `crossplane` Service Account to create managed resources
-* creating and binding RBAC roles for Crossplane composite resources
+* allowing the `crossplane` ServiceAccount to create managed resources
+* creating ClusterRoles to access Crossplane resources in all namespaces
+* creating Roles to access Crossplane resources in specific namespaces
 
-When installing a Provider their service account needs permissions to create and 
-manage managed resources. 
+Use the [ClusterRoles]({{<ref "#crossplane-clusterroles">}}) to grant access to all Crossplane resources in the
+cluster.  
+Use the [Roles]({{<ref "#crossplane-roles" >}}) to only grant access to Crossplane Claims. 
 
-When creating composite resources the Crossplane service account needs
-permissions for the newly created API endpoints. 
+#### Crossplane ClusterRoles
 
-The RBAC manager container manages these functions. 
+The RBAC manager creates four Kubernetes ClusterRoles. These Roles grant 
+permissions over cluster wide Crossplane resources. 
 
-The RBAC manager pod supports two policy options:
-* `All` - The default policy, configures and manages all RBAC capabilities.
-* `Basic` - Doesn't enable namespaced RBAC roles.
+<!-- vale Google.Headings = NO -->
+<!-- disable heading checking for the role names -->
+<!-- vale Google.WordList = NO -->
+<!-- allow "admin" -->
+##### crossplane-admin
+<!-- vale Google.WordList = YES -->
 
-Using the `Basic` policy option prevents creating namespaced Crossplane RBAC
-roles and attaching RBAC annotations to Claims. 
+The `crossplane-admin` ClusterRole has the following permissions:
+  * full access to all Crossplane types
+  * full access to all secrets and namespaces (even those unrelated to Crossplane)
+  * read-only access to all cluster RBAC roles, CustomResourceDefinitions and
+    events
+  * ability to bind RBAC roles to other entities. 
 
-Change from the default `All` to the `Basic` policy with the `-m Basic` pod 
-argument.
+View the full RBAC policy with 
 
-To control all RBAC roles and policies related to Crossplane resources and
-providers, you can disable the RBAC manager. 
+```shell
+kubectl describe clusterrole crossplane:aggregate-to-admin
+```
 
-Disable the RBAC manager after installation by deleting the
-`crossplane-rbac-manager` deployment from the `crossplane-system` namespace.
+##### crossplane-edit
 
-Disable the RBAC manager before installation by editing the Helm `values.yaml`
-file, setting `rbacManager.deploy` to `false`.
+The `crossplane-edit` ClusterRole has the following permissions:
+
+  * full access to all Crossplane types
+  * full access to all secrets (even those unrelated to Crossplane)
+  * read-only access to all namespaces and events (even those unrelated to Crossplane).
+
+View the full RBAC policy with 
+
+```shell
+kubectl describe clusterrole crossplane:aggregate-to-edit
+```
+
+##### crossplane-view
+
+The `crossplane-view` ClusterRole has the following permissions:
+
+  * read-only access to all Crossplane types
+  * read-only access to all namespaces and events (even those unrelated to Crossplane).
+
+View the full RBAC policy with 
+
+```shell
+kubectl describe clusterrole crossplane:aggregate-to-view
+```
+
+##### crossplane-browse
+
+The `crossplane-browse` ClusterRole has the following permissions:
+
+  * read-only access to all Crossplane types
+  * read-only access to all namespaces and events (even those unrelated to Crossplane).
+
+View the full RBAC policy with 
+
+```shell
+kubectl describe clusterrole crossplane:aggregate-to-browse
+```
+
+#### Crossplane Roles
+By default the RBAC manager creates three Roles in every namespace. These Roles 
+grant access to Claims in a specific namespace. The RBAC manager dynamically 
+updates the Roles to access the specific API endpoints in a Claim. 
+
+{{< hint "note" >}}
+The specific details of the namespace Roles are beyond this guide. For more
+information read the [Composite Resource ClusterRole
+Mechanics](https://github.com/crossplane/crossplane/blob/master/design/design-doc-rbac-manager.md#composite-resource-clusterrole-mechanics)
+section of the RBAC Manager design document.
+{{< /hint >}}
+
+You can disable these namespace specific roles by configuring the RBAC manager
+with `--manage=Basic`.
+
 
 {{< hint "note" >}}
 
--- a/content/master/concepts/pods.md
+++ b/content/master/concepts/pods.md
@@ -88,6 +88,17 @@ For objects that Kubernetes can't provide a watch for, Crossplane
 periodically poll the resource to find it's state. The default polling rate is
 one minute. Change the polling rate with the `---poll-interval` pod argument.
 
+Reducing the poll-interval value causes Crossplane to poll resources more
+frequently. This increases the load of the Crossplane pod and
+results in more frequent provider API calls.
+
+<!-- vale write-good.TooWordy = NO -->
+<!-- allow "maximum" -->
+Increasing the poll-interval causes Crossplane to poll resources less
+frequently. This increases the maximum time until Crossplane
+discovers changes in the cloud provider that require updating. 
+<!-- vale write-good.TooWordy = YES -->
+
 Managed resources use polling.
 
 Crossplane checks all resources, both watched and polled resources, to
--- a/content/master/concepts/pods.md
+++ b/content/master/concepts/pods.md
@@ -1,0 +1,231 @@
+---
+title: Crossplane Pods
+weight: 1
+description: Background on the components installed with Crossplane and their functions.
+---
+
+The base Crossplane installation consists of two Pods, the `crossplane` pod and
+the `rbac-manager` pod. Both pods install in the `crossplane-system`
+namespace by default. 
+
+
+## Crossplane pod
+
+### Init pod
+Before starting the core Crossplane pod an _init_ Pod runs. The init pod is
+responsible for setting parameters related to the core Crossplane pod and
+installing the core Crossplane 
+[Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)
+(`CRDs`).
+
+The settings the init pod sets include installing Provider or Configuration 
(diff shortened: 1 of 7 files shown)


content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+
+The base Crossplane installation consists of two Pods, the `crossplane` pod and
+the `rbac-manager` pod. Both pods install in the `crossplane-system`
+namespace by default. 
+
+
+## Crossplane pod
+
+### Init pod
+Before starting the core Crossplane pod an _init_ Pod runs. The init pod is
```
jeanduplessis, line 15: Why is _init_ `P`od written with a capital here but in lowercase almost everywhere else?

jeanduplessis, line 15: Why is _init_ `P`od written with a capital here but in lowercase almost everywhere else?

jeanduplessis, line 15: Why is _init_ `P`od written with a capital here but in lowercase almost everywhere else?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
+---
+title: Crossplane Pods
+weight: 1
+description: Background on the components installed with Crossplane and their functions.
+---
+
+The base Crossplane installation consists of two Pods, the `crossplane` pod and
```
jeanduplessis, line 7: Should two `P`ods be capitalized here?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+* `storeconfigs` for connecting external secret stores like 
+[Hashicorp Vault](https://www.vaultproject.io/)
+
+{{< hint "note" >}}
+
+The [Install Crossplane]({{< ref "../software/install" >}}) section has more
+information about customizing the Crossplane install.
+{{< /hint >}}
+
+The init pod completes and starts the Crossplane core pod.
```
jeanduplessis, line 38: What happens to the init pod after completion? Does the end user have to care about it at all after completion? Does it also run in the `crossplane-system` namespace?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+
+The [Install Crossplane]({{< ref "../software/install" >}}) section has more
+information about customizing the Crossplane install.
+{{< /hint >}}
+
+The init pod completes and starts the Crossplane core pod. 
+
+### Core pod
+
+The main Crossplane pod, called the _core_ pod, enforces
```
jeanduplessis, line 42: It's getting confusing with the naming here. We refer to the main Crossplane pod as "core" pod, but its name is `crossplane` in the `crossplane-system` namespace, right? Not sure we can do any better than what's written but just calling it out.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+2. Periodic polling
+
+Crossplane requests that Kubernetes notifies Crossplane of any changes on
+objects. This notification tool is a _watch_. 
+
+Watched objects include Providers and Composite Resource Definitions.
+
+For objects that Kubernetes can't provide a watch for, Crossplane
+periodically poll the resource to find it's state. The default polling rate is
+one minute. Change the polling rate with the `---poll-interval` pod argument.
```
jeanduplessis, line 69: Where should one change the polling rate? Can we be more specific than just stating to add it as a pod argument. Also is there any considerations that need to be taken into account when changing the poll-interval?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+The RBAC manager pod preforms the following tasks:
+* creating and applying RBAC policies to Provider Service Accounts, allowing 
+  them to control their managed resources
+* allowing the `crossplane` Service Account to create managed resources
+* creating and applying policies for Crossplane Composite Resources
+
+When installing a Provider their service account needs permissions to create and 
+manage managed resources. 
+
+When creating Composite Resources the Crossplane service account needs
```
jeanduplessis, line 126: Why do we capitalize Composite Resources here, yet above we write managed resources in lowercase?

jeanduplessis, line 126: Why do we capitalize Composite Resources here, yet above we write managed resources in lowercase?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+permissions for the newly created API endpoints. 
+
+The RBAC manager pod manages these functions. 
+
+The RBAC manager pod supports two policy options:
+* `All` - The default policy, configures and manages all RBAC capabilities.
+* `Basic` - Only manage policies related to Composite resources. 
+
+Change from the default `All` to the `Basic` policy with the `-m Basic` pod
```
jeanduplessis, line 135: Why/when would one want to switch from All to Basic policy?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+* `Basic` - Only manage policies related to Composite resources. 
+
+Change from the default `All` to the `Basic` policy with the `-m Basic` pod 
+argument.
+
+Disable the RBAC manager after installation by deleting the
+`crossplane-rbac-manager` deployment from the `crossplane-system` namespace.
+
+Disable the RBAC manager before installation by editing the Helm `values.yaml`
+file, setting `rbacManager.deploy` to `false`.
```
jeanduplessis, line 142: Under what circumstances would you want to disable the RBAC manager?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+
+{{< hint "note" >}}
+
+Instructions for changing Crossplane pod settings during installation are in the
+[Crossplane Install]({{<ref "../software/install">}}) section. 
+{{< /hint >}}
+
+## Leader election
+
+By default only a single Crossplane pod runs in a cluster. If more than one
```
jeanduplessis, line 152: Why/how would you run more than one Crossplane pod in a cluster?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+Crossplane pod runs both pods try to manage Crossplane resources. To prevent
+conflicts Crossplane uses a _leader election_ to have a single pod in control at
+a time. Other Crossplane pods standby until the leader fails. 
+
+{{<hint "important" >}}
+Leader election applies to both the Crossplane pod and the RBAC manager pod.
+{{< /hint >}}
+
+Enable leader elections with the `--leader-election` pod argument.
```
jeanduplessis, line 161: Would this argument set which XP pod becomes the leader?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+<!-- allow "is unsupported" --> 
+Running multiple Crossplane pods without leader election is unsupported.
+<!-- vale write-good.Passive = YES -->
+<!-- vale write-good.TooWordy = YES -->
+{{< /hint >}}
+
+
+## Change pod settings
+
+Changing Crossplane pod settings either before installing Crossplane by editing
```
jeanduplessis, line 176: "Changing Crossplane pod..." should be "Change Crossplane pod..."

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+deployment in the `crossplane-system` namespace with the command
+
+`kubectl edit deployment crossplane --namespace crossplane-system`
+
+{{< hint "warning" >}}
+Updating the Crossplane deployment restarts the Crossplane pod.
+{{< /hint >}}
+
+Add Crossplane pod arguments to the 
+{{<hover label="args" line="7" >}}spec.template.spec.containers[].args{{< /hover >}}
```
jeanduplessis, line 204: Line="7" doesn't seem like the right line here.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+section of the deployment.
+
+For example, to change the `sync-interval` add 
+{{<hover label="args" line="10" >}}--sync-interval=30m{{< /hover >}}.
```
jeanduplessis, line 208: Line="10" doesn't seem like the right line here.

jeanduplessis (changes requested):
@plumbis Thanks for starting this concept doc. I found it very valuable. Thanks for your patience with some of the noob questions on this content.

content/master/concepts/pods.md
plumbis: More accurately it's the init _container_ and core _container_. Both of these containers are in the `crossplane` _pod_ inside the `crossplane-system` namespace. The name "core" is what Crossplane calls it in source. 

Would being more specific that they are two _containers_ inside a single _pod_ help? Otherwise we're stuck with

content/master/concepts/pods.md
bobh66: The polling interval is an input to the helm chart for Crossplane deployment

content/master/concepts/pods.md
bobh66: This should be "the init container" runs.  The Pod has "initContainers" that run at startup and "containers" that run after startup.

content/master/concepts/pods.md
bobh66: No - it is random based on timing.  It is even possible that the "failed" pod could regain the leader role after it restarts if the timer for the "standby" pod(s) has not expired yet.

plumbis: No, "who" is the leader doesn't matter since it's just another pod in the deployment. Just like the specific node a pod runs on doesn't matter. 

The election itself is driven by kubernetes. 

Should I add more context?

content/master/concepts/pods.md
plumbis: fixed

content/master/concepts/pods.md
plumbis: fixed.

content/master/concepts/pods.md
plumbis: I added more information about viewing the init container in `k get pods` and that starts the core container automatically.

content/master/concepts/pods.md
plumbis: fixed

content/master/concepts/pods.md
plumbis: Added a note.

content/master/concepts/pods.md
plumbis: Added a note that you can run more than one but most people don't need it. (Feedback from Nic)

content/master/concepts/pods.md
plumbis: fixed

content/master/concepts/pods.md
plumbis: fixed

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+---
+
+The base Crossplane installation consists of two Pods, the `crossplane` pod and
+the `rbac-manager` pod. Both pods install in the `crossplane-system`
+namespace by default. 
+
+
+## Crossplane pod
+
+### Init pod
```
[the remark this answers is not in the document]
plumbis: Fixed all pod/container references.

content/master/concepts/pods.md
plumbis: I added some text on the impact of changing the poll-interval. 

@jeanduplessis are you okay with the note on "Read the Change Pod Settings section...." for information on applying the change or do you want more here?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+Crossplane requests that Kubernetes notifies Crossplane of any changes on
+objects. This notification tool is a _watch_. 
+
+Watched objects include Providers and Composite Resource Definitions.
+
+For objects that Kubernetes can't provide a watch for, Crossplane
+periodically poll the resource to find it's state. The default polling rate is
+one minute. Change the polling rate with the `---poll-interval` pod argument.
+
+Managed resources use polling.
```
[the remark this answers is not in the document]
plumbis: What would trigger an MR watch? I figured that MR changes would only happen based on external APIs, so it wouldn't be a notification. 

I wasn't able to figure out what is and isn't assigned a watch in source.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,231 @@
...
+
+Crossplane checks all resources, both watched and polled resources, to
+confirm they're in the desired state. Crossplane does this every one hour by
+default. Use the `--sync-interval` Crossplane pod argument to change this
+interval. 
+
+The most that Crossplane acts on objects is the 
+`--max-reconcile-rate` rate. This rate is 10 times per second by default.
```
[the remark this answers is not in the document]
plumbis: Do you mean that it's 10 API calls per second AND 10 goroutines?  Meaning the value is two different settings?

content/master/concepts/pods.md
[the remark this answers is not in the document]
bobh66: The MR controller is by definition "watching" all MRs of that type - it's the same for all Crossplane resource controllers, they all "watch" the specific composite/managed/etc resources that they are responsible for.  The synch interval is presumably intended to account for any "missed events" and verify that the resource state is what it should be regardless of whether the controller missed an event.  (This is my understanding - if Nic tells you something different, believe him.)

content/master/concepts/pods.md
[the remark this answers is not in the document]
bobh66: Yes - as shown here:  https://github.com/crossplane/crossplane/blob/master/cmd/crossplane/core/core.go#L178

it is used for both MaxConcurrentReconciles (number of goroutines) and GlobalRateLimiter

content/master/concepts/pods.md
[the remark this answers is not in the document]
plumbis: Insightful! Do you have an example of a MR event that would be caught by a watch?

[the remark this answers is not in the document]
bobh66: Probably the simplest case is deletion - when someone or something deletes an MR kubernetes will add the `deletionTimestamp` and `deletionGracePeriodSeconds` attributes to the `metadata` section which triggers an event to the controller, the reconciler sees that the MR has been deleted and calls the external Delete() function to trigger resource deletion via the provider.  Another case would if the someone or something changed a value in the `spec` of the resource - that would also trigger an event to the controller which would detect the difference and trigger an update via the provider.

Note that polling/syncing and watched events are used together - the events are kubernetes-detected changes in (usually desired) state as above, where polling and syncing are looking for drift on the remote side (actual state).  Polling and syncing are time-based (check for differences every 5m/1h) where events are immediate (hey something changed you might care about - or not)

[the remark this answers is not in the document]
plumbis: Thank you! I've added a MRs to the list of resources that uses watches as well as a notice under "MRs use polling" to explain when an MR would rely on the watch vs poll. 

I'm not trying to get too deep but just create awareness in the user.

content/master/concepts/pods.md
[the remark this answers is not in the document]
plumbis: Thanks again. I added a bunch of stuff here to expand on it. I'm not 100% sure I got it right and accurate so I'd love a second review when you have a chance.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+The full list of configuration options are available in the 
+[Crossplane Install]({{<ref "../software/install">}}) section. 
+
+{{< hint "note" >}}
+
+Instructions for changing Crossplane pod settings during installation are in the
+[Crossplane Install]({{<ref "../software/install">}}) section. 
+{{< /hint >}}
+
+### Edit the deployment
```
negz, line 278: Could we link to upstream Kubernetes documentation for this? This feels like we're teaching folks how to use Kubernetes.

negz (commented):
I'm seeing a bunch of changes that don't seem to be related to the PR description - i.e. updates to content under the v1.12 and v1.9 directories. Is this intentional?

plumbis:
> I'm seeing a bunch of changes that don't seem to be related to the PR description - i.e. updates to content under the v1.12 and v1.9 directories. Is this intentional?

I pulled in the v1.12 release either on accident or purpose. I've rebased and removed that commit. Only the pods.md file and vale updates are in the PR now.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+
+The settings the init container sets include installing Provider or Configuration 
+packages with Crossplane, customizing the namespace Crossplane installs in and 
+defining webhook configurations. 
+
+The core CRDs installed by the init container include: 
+* Composite Resource Definitions, Compositions, Configurations and Providers
+* `locks` to manage package dependencies
+* `controllerconfigs` to apply settings to installed Providers
+* `storeconfigs` for connecting external secret stores like
```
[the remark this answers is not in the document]
plumbis: Do you mean why are some of them in code ticks?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+
+The main Crossplane container, called the _core_ container, enforces
+the desired state of Crossplane resources, manages leader elections and process
+webhooks. 
+
+#### Reconcile loop
+
+The core container operates on a _reconcile loop_, constantly checking the 
+status of deployed resources and correcting any "drift." After checking a 
+resource Crossplane waits some time and checks again.
```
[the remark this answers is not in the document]
plumbis: I don't think this should be moved, since the page flow is about the pods and their interactions/responsibilities but I do think the [root concepts page](https://docs.crossplane.io/v1.12/concepts/) cover this at or near the top instead of being a rehash of what's in the [Introduction](https://docs.crossplane.io/v1.12/getting-started/introduction/) page and what's spelled out in the individual concept sections.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+
+Managed resources use polling.
+
+{{< hint "note" >}}
+Managed resources watch for Kubernetes events like deletion or changes to
+their `spec`. Managed resources rely on polling to detect changes in the
+provider environment.
+{{< /hint >}}
+
+Crossplane checks all resources, both watched and polled resources, to
```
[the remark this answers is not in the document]
plumbis: I added a line earlier "Some resources may be both watched and polled"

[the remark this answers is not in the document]
bobh66: From the user's perspective it's essentially all resources that they care about/create are both watched and polled.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+
+* The global retry rate for checking failed objects.
+* The number of threads created to reconcile resources.
+
+Reducing the `--max-reconcile-rate`, or making it smaller, increases the
+CPU resources Crossplane uses.
+
+Increasing the `--max-reconcile-rate`, or making it larger, reduces CPU 
+resources Crossplane uses, but increases the amount of time until changed 
+resources are fully synced.
```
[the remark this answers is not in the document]
plumbis: Had those switched. Fixed.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+
+When Crossplane checks a resource and it's not in the desired
+state they 100 times per second. If the resource is still not in the desired 
+state Crossplane checks the `--max-reconcile-rate` times per second. By default,
+the `--max-reconcile-rate` value of 10 means checking 10 times per second. 
+
+{{< hint "note" >}}
+Kubernetes defines the initial rate of 100 times per second. This isn't
+configurable.
+{{< /hint >}}
```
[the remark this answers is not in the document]
plumbis: Gladly willing to change this. It was from trying to understand how the rate limiter works and ended up on Dan's [blog about client-go](https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/#the-default-controller-rate-limiter) describing the token bucket rate limiter it uses.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+
+Crossplane checks all resources, both watched and polled resources, to
+confirm they're in the desired state. Crossplane does this every one hour by
+default. Use the `--sync-interval` Crossplane pod argument to change this
+interval. 
+
+The `--max-reconcile-rate` rate defines two different settings at the same time:
+
+* The global retry rate for checking failed objects.
+* The number of threads created to reconcile resources.
```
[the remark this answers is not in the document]
plumbis: I've attempted to simplify this section.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+<!-- vale Microsoft.HeadingAcronyms = NO -->
+<!-- allow 'RBAC' since that's the name -->
+### RBAC manager container
+<!-- vale Microsoft.HeadingAcronyms = YES -->
+
+The RBAC manager container preforms the following tasks:
+* creating and applying RBAC policies to Provider Service Accounts, allowing 
+  them to control their managed resources
+* allowing the `crossplane` Service Account to create managed resources
+* creating and applying policies for Crossplane composite resources
```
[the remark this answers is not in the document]
plumbis: >It's not clear to me what this means.

What isn't clear? I'm trying to indicate that the RBAC manager creates "some policies" that Crossplane needs to operate.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+
+The RBAC manager container manages these functions. 
+
+The RBAC manager pod supports two policy options:
+* `All` - The default policy, configures and manages all RBAC capabilities.
+* `Basic` - Only manage policies related to Composite resources. 
+
+Use the `Basic` policy option when manually assigning RBAC policies to
+Crossplane providers and provider resources. The RBAC manager still creates
+cluster roles for composite resources.
```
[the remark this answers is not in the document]
plumbis: I edited this section to hint at that. I'm trying to avoid fully describing every ClusterRole/Role that's created and the policies around it.

content/master/concepts/pods.md
plumbis: The amount of "teaching kubernetes" is light and just provides practical context to applying the settings discussed earlier.

content/master/concepts/pods.md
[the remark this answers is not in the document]
negz: In part, yes. Everything you've listed on the highlighted lines is both a Crossplane concept and a concrete API type. You could write any of them in three ways:

1. Composite Resource Definitions, Providers, Locks, Store Configs, etc.(Typical prose)
2. `CompositeResourceDefinition`, `Provider`, `Lock`, `StoreConfig`, etc. (`KindName` style)
3. `compositeresourcedefinitions`, `providers` `locks`, `storeconfigs`, etc. (`pluralname` style)

I would recommend using one of the first two styles consistently. The `pluralname` style where a type is written as lower case, plural, and in a code block is very uncommon relative to the `KindName` style.

content/master/concepts/pods.md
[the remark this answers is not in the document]
negz: Yeah, I agree. It's not so much that I don't want it mentioned here insomuch as I think it should appear somewhere more prominent. If that were the case I could see this being a note on this page rather than an entire heading.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+{{< hint "important" >}}
+Most providers use their own `--max-reconcile-rate`. This determines the
+same settings for Providers and their managed resources. 
+{{< /hint >}}
+##### Reconcile retry rate
+
+When Crossplane checks a resource and it's not in the desired
+state they 100 times per second. If the resource is still not in the desired 
+state Crossplane checks the `--max-reconcile-rate` times per second. By default,
+the `--max-reconcile-rate` value of 10 means checking 10 times per second.
```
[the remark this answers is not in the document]
negz: Re-opening this thread as there's still some strange grammar in here, e.g.:

> When Crossplane checks a resource and it's not in the desired state __they__ 100 times per second.

Does my comment above help communicate what `--max-reconcile-rate` does? I'd like to make sure we remove any mention of anything happening 100 times per second - it's wrong and I think it would be worrying to readers. Let me know if I can help better explain how this works. Happy to take a stab at writing this section myself if necessary.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+When creating composite resources the Crossplane service account needs
+permissions for the newly created API endpoints. 
+
+The RBAC manager container manages these functions. 
+
+The RBAC manager pod supports two policy options:
+* `All` - The default policy, configures and manages all RBAC capabilities.
+* `Basic` - Doesn't enable namespaced RBAC roles.
+
+Using the `Basic` policy option prevents creating namespaced Crossplane RBAC
```
negz, line 213: I know you're attempting to be succinct here, but if you could briefly mention what the namespaced Crossplane RBAC roles are for I think it would help contextualize why folks would potentially use `Basic`.

Something like:

> When the RBAC manager uses the `All` policy it will create a series of opinionated RBAC roles in each namespace. You can bind these RBAC roles to users to grant them access to Crossplane types within a namespace.

content/master/concepts/pods.md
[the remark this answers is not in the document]
negz: > What isn't clear? 

I think it's the missing "what for". The four broad things the RBAC manager does is:

1. Create and bind a role to Crossplane's service account, so it can control managed resources.
2. Create and bind roles to provider service accounts, so they can reconcile their managed resources.
3. Create cluster roles that you can bind to grant users access to Crossplane resources in all namespaces.
4. Create roles that you can bind to grant users access to Crossplane resources in a specific namespace.

Noting that 4 is disabled when you run the RBAC manager in `Basic` mode.

At the moment you capture the why of the first two well - these roles enable Crossplane and provider pods do what they need to do. What's throwing me off is that 3 and 4 are a bit over-simplified:

* I would briefly mention. what the roles are for.
* The roles grant access to claims, MRs, secrets, etc in addition to XRs. So I would suggest being non-specific and saying "Crossplane resources". Calling out XRs specifically implies to me they grant only that.

content/master/concepts/pods.md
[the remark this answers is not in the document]
negz: Understood. I think it's worth documenting these policies in detail somewhere, but the "pods" page probably isn't it. Let's discuss in the thread at https://github.com/crossplane/docs/pull/412#discussion_r1183121100.

content/master/concepts/pods.md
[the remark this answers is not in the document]
plumbis: After digging through the [k8s style guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects), the [PR that pushed that guideline](https://github.com/kubernetes/website/pull/23497) and some [related discussion](https://github.com/kubernetes/website/issues/20862) the best answer seems to be the k8s style "[UpperCamelCase](https://en.wikipedia.org/wiki/Camel_case)"

The only problem I have with this is that it's confusing in how k8s describes when to use sentence-style capitalization and their docs are inconsistent in this styling so there aren't great examples to follow.

That being said I've updated this to use UpperCamelCase in the called out section and tried to update any other references in the rest of the doc.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,321 @@
...
+
+The core container operates on a _reconcile loop_, constantly checking the 
+status of deployed resources and correcting any "drift." After checking a 
+resource Crossplane waits some time and checks again.
+
+{{<hint "note" >}}
+The Crossplane pod only reconciles core Crossplane components, including claims
+and composite resources. Providers are responsible for reconciling their managed
+resources. 
+{{< /hint >}}
```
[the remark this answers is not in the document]
plumbis: moved.

content/master/concepts/pods.md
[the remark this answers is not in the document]
plumbis: I understand. Making it just a note doesn't feel like it's enough. I think the solution is we need to give reconciliation more attention in the parent /concepts page.

content/master/concepts/pods.md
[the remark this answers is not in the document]
plumbis: I've simplified the wording here and removed the reference to 100x

content/master/concepts/pods.md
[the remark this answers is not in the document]
plumbis: Removed

content/master/concepts/pods.md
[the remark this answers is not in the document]
plumbis: I was confused in your earlier comments on the `Roles` vs `ClusterRoles`. I've rewritten the section and I chose not to detail the `Roles`. Maybe in the future we can create an RBAC section that would go into details.

content/master/concepts/pods.md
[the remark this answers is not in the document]
plumbis: I was confused in your earlier comments on the `Roles` vs `ClusterRoles`. I've rewritten the section and I chose not to detail the Roles. Maybe in the future we can create an RBAC section that would go into details.

content/master/concepts/pods.md
```diff
@@ -0,0 +1,384 @@
...
+The `--max-reconcile-rate` setting configures the number of times per second
+Crossplane or a provider attempts to correct a resource. The default value is 
+10 times per second.
+
+All core Crossplane components share the reconcile rate. Each Provider
+implements their own max reconcile rate setting. 
+
+##### Number of reconcilers
+
+The second value `--max-reconcile-rate` defines is the number of threads
```
negz, line 144: I'd still prefer not to talk about threads. It's not accurate. Could we find some other approachable way to phrase this? i.e. How many resources can be reconciled concurrently?

content/master/concepts/pods.md
```diff
@@ -0,0 +1,384 @@
...
+  * full access to all Crossplane types
+  * full access to all secrets and namespaces (even those unrelated to Crossplane)
+  * read-only access to all cluster RBAC roles, CustomResourceDefinitions and
+    events
+  * ability to bind RBAC roles to other entities. 
+
+View the full RBAC policy with 
+
+```shell
+kubectl describe clusterrole crossplane:aggregate-to-admin
```
negz, line 232:
```suggestion
kubectl describe clusterrole crossplane-admin
```

`crossplane:aggregate-to-admin` isn't the full role, it's just part of it. `crossplane-admin` is the full role.

(Ditto for the edit, view, and browse roles below.)

negz (commented):
I've only reviewed the content in the `master` directory, not `v1.11` or `v1.12` which I presume are identical.

content/master/concepts/pods.md
plumbis: I'm struggling with this. I understand why you want to avoid `thread` but it seems like we're hinting at something without actually naming it. How is this edit? 
---
The second value `--max-reconcile-rate` defines is the number of
resources that can reconcile at once. If there are more resources than
configured `--max-reconcile-rate` the remaining resources must wait until
Crossplane reconciles a an existing resource.

Read the [Change Pod Settings]({{<ref "#change-pod-settings">}}) section for
instructions on applying these settings.

negz: That sounds great to me.

negz (approved):
This looks great, thanks for all the effort!

Issue #336, opened 2023-02-03
Instructions on how to compile a Crossplane Provider with just a subset of the available CRDs


nicolas-g:
### What's Missing?
As we all know, the number of CRDs significantly impacts the Kubernetes cluster performance.

For deployments that only require a small number of CRDs, it will be good to have instructions on compiling the Crossplnae provider with just a tiny number or CRDs.

jbw976:
Related to https://github.com/crossplane/crossplane/issues/2869

negz:
I don't think we should document how to do this. It's not a practice we support.

As soon as we have bandwidth on the maintainer team we'll be starting work on a design document to address https://github.com/crossplane/crossplane/issues/2869.

jbw976:
Here's some related discussion for further context @nicolas-g: https://github.com/crossplane/crossplane/issues/2895#issuecomment-1379781246

plumbis:
I'm going to close this as "won't fix" and out of scope for the user docs. The Provider section update in #420 now has a mention and link to the contrib guide on building your own providers.

PR #420, opened 2023-04-26
Concepts: Provider rewrite

### LINKED ISSUE #235: Document the officially supported conditions in MR,XR,XRD,Provider
### Expected

As a crossplane user,
in order to rely on resource conditions
I need the conditions to be documented

E.g, is it sufficient to wait for condition type=Ready on a MR to ensure its success

```
apiVersion: database.gcp.crossplane.io/v1beta1
kind: CloudSQLInstance
[...]
status:
  atProvider:
[...]
  conditions:
  - lastTransitionTime: "2022-11-24T17:20:31Z"
    reason: Available
    status: "True"
    type: Ready
  - lastTransitionTime: "2022-12-14T19:29:37Z"
    message: 'update failed: cannot update the CloudSQL instance: googleapi: Error
      400: Invalid request: Project [...]
    reason: ReconcileError
    status: "False"
    type: Synced
```

```
apiVersion: sql.gcp.upbound.io/v1beta1
kind: DatabaseInstance
[...]
conditions:
  - lastTransitionTime: "2022-12-12T17:50:49Z"
    reason: Available
    status: "True"
    type: Ready
  - lastTransitionTime: "2022-12-12T16:02:17Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
  - lastTransitionTime: "2022-12-12T17:50:26Z"
    reason: Finished
    status: "True"
    type: AsyncOperation
  - lastTransitionTime: "2022-12-12T16:06:16Z"
    reason: Success
    status: "True"
    type: LastAsyncOperation
  - lastTransitionTime: "2022-12-14T11:10:51Z"
    reason: UpToDate
    status: "True"
    type: Test
```

### Observed

I did not find sufficient documentation so far

https://docs.crossplane.io/master/reference/troubleshoot/#resource-status-and-conditions
> Most Crossplane resources set the Ready condition. Ready represents the availability of the resource - whether it is creating, deleting, available, unavailable, binding, etc.

https://github.com/crossplane/crossplane/blob/e6b1acc6932c107c77a2ede3701eac31e017facc/design/one-pager-managed-resource-api-design.md?plain=1#L423
* Illustrates a Synced condition but does not detail it

https://github.com/upbound/upjet/blob/15411ad4889daad82962abc8063dbfbc50156d92/pkg/resource/conditions.go#L17-L64

does not provide much details over conditions

### LINKED ISSUE #278: Provider upgrade procedure
### What problem are you facing?

I am trying without sucess to find a document that describes what happens if provider such as AWS introduced new resource version for example v1beta1 and deprecates v1alpha1 totally. If we have EKS Cluster deployed will provider itself migrate the existing resources, or this should be done manually, If the answers is manually, then what is the common procedure for such update 

If we create a new resource based on new definition, will this be linked to existing resource, or it will try to provision new one over provider? Do we have an option to create a new resource and link it to exisitng deployed resource?

### How could Crossplane help solve your problem?

Section in documentation describing responsibilities and procedures would help a lot.

plumbis:
Major update to the Providers "concepts' chapter. 

"Resolves" #336 by pointing to the Provider contrib instructions.

Partially resolves #235 by detailing the possible provider conditions.

Resolves #278 by adding Provider upgrade instructions.

### DIFF
--- a/content/master/concepts/providers.md
+++ b/content/master/concepts/providers.md
@@ -3,8 +3,6 @@ title: Providers
 weight: 101
 ---
 
-<!-- REsolves #336, #235, #278 -->
-
 Providers enable Crossplane to provision infrastructure on an
 external service. Providers create new Kubernetes APIs and map them to external 
 APIs.
--- a/content/master/concepts/providers.md
+++ b/content/master/concepts/providers.md
@@ -3,52 +3,308 @@ title: Providers
 weight: 101
 ---
 
-Providers are Crossplane packages that bundle a set of 
-[Managed Resources]({{<ref "managed-resources" >}}) and their respective controllers to allow
-Crossplane to provision the respective infrastructure resource.
+<!-- REsolves #336, #235, #278 -->
 
-## Installing Providers
+Providers enable Crossplane to provision infrastructure on an
+external service. Providers create new Kubernetes APIs and map them to external 
+APIs.
 
-The core Crossplane controller can install provider controllers and CRDs for you
-through its own provider packaging mechanism, which is triggered by the
-application of a `Provider` resource. For example, in order to request
-installation of the `provider-aws` package, apply the following resource to the
-cluster where Crossplane is running:
+Providers are responsible for all aspects of connecting to non-Kubernetes
+resources. This includes authentication, making external API calls and
+providing
+[Kubernetes Controller](https://kubernetes.io/docs/concepts/architecture/controller/)
+logic for any external resources.
 
-```yaml
+Examples of providers include:
+* [Provider AWS](https://github.com/upbound/provider-aws)
+* [Provider Azure](https://github.com/upbound/provider-azure)
+* [Provider GCP](https://github.com/upbound/provider-gcp)
+* [Provider Kubernetes](https://github.com/crossplane-contrib/provider-kubernetes)
+
+{{< hint "tip" >}}
+Find more providers in the [Upbound Marketplace](https://marketplace.upbound.io).
+{{< /hint >}}
+
+<!-- vale write-good.Passive = NO -->
+<!-- "are Managed" isn't passive in this context -->
+Providers define every external resources they can create in Kubernetes as a
+Kubernetes API endpoint. These endpoints are 
+[_Managed Resources_]({{<ref "managed-resources" >}}).
+<!-- vale write-good.Passive = YES -->
+
+Providers are cluster scoped and any namespace can access them.
+
+{{< hint "note" >}}
+Instructions on building your own Provider are outside of the scope of this
+document. Read the Crossplane contributing [Provider Development
+Guide](https://github.com/crossplane/crossplane/blob/master/contributing/guide-provider-development.md)
+for more information.
+{{< /hint >}}
+
+## Install a Provider
+
+Installing a provider creates a Provider pod that's responsible for installing
+the Provider's APIs into the Kubernetes cluster. Providers constantly watch the
+state of the desired managed resources and create any external resources that
+are missing. 
+
+Install a Provider with a Crossplane
+{{<hover label="install" line="2">}}Provider{{</hover >}} object setting the
+{{<hover label="install" line="6">}}spec.package{{</hover >}} value to the
+location of the provider package.
+
+For example, to install the 
+[AWS Community Provider](https://github.com/crossplane-contrib/provider-aws),
+
+```yaml {label="install"}
 apiVersion: pkg.crossplane.io/v1
 kind: Provider
 metadata:
   name: provider-aws
 spec:
-  package: "xpkg.upbound.io/crossplane-contrib/provider-aws:v0.33.0"
+  package: xpkg.upbound.io/crossplane-contrib/provider-aws:v0.39.0
 ```
 
-The field `spec.package` is where you refer to the container image of the
-provider. Crossplane Package Manager will unpack that container, register CRDs
-and set up necessary RBAC rules and then start the controllers.
+{{< hint "tip" >}}
+Providers are Crossplane Packages. Read more about Packages in the
+[Packages documentation]({{<ref "packages" >}}).
+{{< /hint >}}
 
-There are a few other ways to to trigger the installation of provider packages:
+By default, the Provider pod installs in the same namespace as Crossplane
+(`crossplane-system`).
 
-* As part of Crossplane Helm chart by adding the following statement to your
-  `helm install` command: `--set
-  provider.packages={xpkg.upbound.io/crossplane-contrib/provider-aws:v0.33.0}`.
-* Using the Crossplane CLI: `kubectl crossplane install provider
-  xpkg.upbound.io/crossplane-contrib/provider-aws:v0.33.0`
+### Install with Helm
 
-You can uninstall a provider by deleting the `Provider` resource
-you've created.
+Crossplane supports installing Providers during an initial Crossplane 
+installation with the Crossplane Helm chart.
 
-## Configuring Providers
+Use the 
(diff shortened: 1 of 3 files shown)


jeanduplessis (changes requested):
@plumbis this PR contains commits from #412. Could you remove the commits from this PR so it only includes the changes related to this PR.

plumbis:
I dropped the two commits and rebased. I think this fixed it?

content/master/concepts/providers.md
```diff
@@ -3,52 +3,306 @@ title: Providers
...
+* [Provider Kubernetes](https://github.com/crossplane-contrib/provider-kubernetes)
 
-```yaml
+{{< hint "tip" >}}
+Find more providers in the [Upbound Marketplace](https://marketplace.upbound.io).
+{{< /hint >}}
+
+<!-- vale write-good.Passive = NO -->
+<!-- "are Managed" isn't passive in this context -->
+Providers define every external resources they can create in Kubernetes as a
```
negz, line 28:
```suggestion
Providers define every external resource they can create in Kubernetes as a
```

This looks like a typo.

content/master/concepts/providers.md
```diff
@@ -3,52 +3,306 @@ title: Providers
...
+Kubernetes API endpoint. These endpoints are 
+[_Managed Resources_]({{<ref "managed-resources" >}}).
+<!-- vale write-good.Passive = YES -->
+
+Providers are cluster scoped and any namespace can access them.
```
negz, line 33: What are you intending to communicate here?

* If you mean the `Provider` API type that is used to install providers, it's more accurate to say they exist outside of any namespace. Namespaces don't really "access" them, and callers aren't accessing them "from" a namespace.
* If you mean the provider in the more general sense, i.e. the running process that reconciles stuff, they can't be accessed from any namespace.

content/master/concepts/providers.md
```diff
@@ -3,52 +3,306 @@ title: Providers
...
+By default, the Provider pod installs in the same namespace as Crossplane
+(`crossplane-system`).
+
+### Install with Helm
+
+Crossplane supports installing Providers during an initial Crossplane 
+installation with the Crossplane Helm chart.
+
+Use the 
+{{<hover label="helm" line="5" >}}--set provider.pacakges{{</hover >}}
```
negz, line 80:
```suggestion
{{<hover label="helm" line="5" >}}--set provider.packages{{</hover >}} 
```

content/master/concepts/providers.md
```diff
@@ -3,52 +3,306 @@ title: Providers
...
+
+```yaml {label="pps"}
+apiVersion: pkg.crossplane.io/v1
+kind: Provider
+metadata:
+  name: private-provider
+spec:
+  package: private-repo.example.org/providers/my-provider
+  packagePullSecrets:
+    - name: my-secret
```
negz, line 107: If I recall correctly these secrets must exist in the same namespace as Crossplane (i.e. `crossplane-system`). It's probably worth pointing that out.

content/master/concepts/providers.md
```diff
@@ -3,52 +3,306 @@ title: Providers
...
+```
+## Upgrade a Provider
+
+To upgrade an existing Provider edit the installed Provider Package by either
+applying a new Provider manifest or with `kubectl edit providers`.
+
+Update the version number in the Provider's `spec.package` and apply the change. 
+Crossplane installs the new image and creates a new `ProviderRevision`.
```
negz, line 115: You could also completely update the OCI reference (e.g. switch to the same provider or a new version of the same provider, but hosted in a different OCI registry). That may be TMI here though.

content/master/concepts/providers.md
```diff
@@ -3,52 +3,306 @@ title: Providers
...
+
+Remove a Provider by deleting the Provider object with `kubectl delete
+provider`. 
+
+{{< hint "warning" >}}
+Removing a Provider without first removing the Provider's managed resources 
+may abandon the resources. The external resources aren't deleted.
+
+You must manually delete external resources through your cloud provider. Managed
+resources must be manually deleted by removing their finalizers.
```
negz, line 127: Perhaps prefix with "in this case" or similar. Otherwise this could be read to mean that you must do this in happy path, i.e. that you must always delete external resources via your cloud provider before uninstalling a provider.

content/master/concepts/providers.md
```diff
@@ -3,52 +3,306 @@ title: Providers
...
+View the conditions of a provider under their `Status` with 
+`kubectl describe provider`.
 
-In order to authenticate with the external provider API, the provider
-controllers need to have access to credentials. It could be an IAM User for AWS,
-a Service Account for GCP or a Service Principal for Azure. Every provider has a
-type called `ProviderConfig` that has information about how to authenticate to
-the provider API. An example `ProviderConfig` resource for AWS looks like the
-following:
+Providers have the following possible conditions:
```
negz, line 173: Nit: Technically the condition is the type (i.e.) `Installed`. Things like `ActivePackageRevision` are the reason that the condition is true or false.

content/master/concepts/providers.md
```diff
@@ -62,34 +316,87 @@ spec:
...
+Provider can authenticate with multiple users or accounts through 
+ProviderConfigs.
+<!-- vale write-good.TooWordy = YES -->
+
+Each account's credentials tie to a unique ProviderConfig. When creating a
+managed resource, attach the desired ProviderConfig. 
+
+For example, two AWS ProviderConfigs, named 
+{{<hover label="user" line="4">}}user-keys{{</hover >}} and
+{{<hover label="admin" line="4">}}admin-keys{{</hover >}}
```
negz, line 338: I'm not really suggesting a change here, but FWIW two ProviderConfigs _could_ share the same secret/credentials. It would be pretty uncommon, but if the ProviderConfig exposed other configuration options you could imagine those options (not credentials) being where two ProviderConfigs varied.

utils/vale/styles/Crossplane/spelling-exceptions.txt
```diff
@@ -9,9 +9,11 @@ Datastore
 editCode
 Enum
 Env
+finalizers
 GCP's
 Geekdocs
 Grammarly
+Hashicorp
```
negz, line 16: I believe the brand/trademark uses a capital C.

```suggestion
HashiCorp
```

content/master/concepts/providers.md
plumbis: added a note.

content/master/concepts/providers.md
plumbis: Added "If you remove the Provider first, you must manually delete..."

content/master/concepts/providers.md
plumbis: I'm using "condition" to refer to the set of {Type, Status, Reason}. Specifically `Condition` coming from the k8s output. 

Yes, the headings are the `Reason` but I wanted to key on the thing that would be unique to make it easier for readers to find what they're looking for.

content/master/concepts/providers.md
plumbis: Good to know. My goal with this example is to draw attention to ProviderConfigs being applied through the managed resource, meaning that one provider can use multiple auth methods.

negz (approved):

jbw976:
Is there a possibility of backporting this to v1.12 docs, so it will show up when folks come to the docs site and see "latest" docs by default?

plumbis:
Great point @jbw976, oversight on my part. Opened PR #436 to do this.

plumbis:
Resolves #452

PR #421, opened 2023-04-27
add search to the home page. 

### LINKED ISSUE #202: Improve Search Results (include version)
Algolia can solve this but it's more complicated than I originally thought. Today's search a basic implementation of [DocSearch](https://docsearch.algolia.com/). 

We need to build custom [Algolia widgets](https://www.algolia.com/doc/guides/building-search-ui/widgets/showcase/js/) to expand the functionality. 

The goals for this issue are: 
* Provide the ability to search only on a specific version or versions
* Return version numbers with the search results
* Add future support to filter on "knowledge base" or "docs" 

Nice to haves:
* Results and previews similar to how [Algolia docs results](https://www.algolia.com/doc/?q=test) are returned.
* Support for [Algolia metrics](https://unomena.atlassian.net/browse/UP-100#:~:text=Add%20Algolia-,click/convert,-metrics) to see what's being searched for and what is missing.

### LINKED ISSUE #419: Add search box to main docs landing page
### What's Wrong?

When I go to the main docs landing page of https://docs.crossplane.io/, we are presented with the high level choices of user docs, knowledge base, and contributor guide.  However, I often know the content I'm looking for but am not sure under which section it would live. I want to start using search right away, but there is no search on this page.

Is it possible to include the search box directly on this main landing page, so users don't have to first pick a subsection before they can start using search?

Surfacing this as a top level issue from its first mention in https://github.com/crossplane/docs/issues/202#issuecomment-1421352852.

plumbis:
Adds a search button to the index/home page. Resolves #419

### DIFF
--- a/themes/geekboot/assets/scss/_home.scss
+++ b/themes/geekboot/assets/scss/_home.scss
@@ -20,6 +20,13 @@
     color: var(--body-font-color);
 }
 
+@include media-breakpoint-up(lg) {
+.home-search {
+    display: flex !important;
+
+}
+}
+
 .list-group-item {
     border-color: var(--dropdown-border-color);
     background: var(--body-background);
--- a/themes/geekboot/layouts/_default/home.html
+++ b/themes/geekboot/layouts/_default/home.html
@@ -8,8 +8,28 @@
   (dict "Contributing Guide" "/img/multi-crane.svg")
 }}
 
-<div class="container-fluid py-5 bd-content" id="home">
-  <div class="row mx-5 justify-content-evenly my-5">
+
+<div class="container-fluid bd-content" id="home">
+
+
+  <aside class="bd-sidebar">
+    <div class="offcanvas-lg offcanvas-start bd-sidebar-container" tabindex="-1" id="bdSidebar" aria-labelledby="bdSidebarOffcanvasLabel">
+      <div class="offcanvas-header pb-4 border-bottom">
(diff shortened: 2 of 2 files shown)


jbw976 (commented):
Nice @plumbis, thank you for adding this feature I've been wanting for awhile! I know others will benefit from this as well 💪 

> Resolves #202

Is it more accurate to say that it resolves #419 instead, since its only focused on adding search to the homepage?  Looks like #202 has more functionality like searching specific versions etc.

The search box width is wide enough to display the entire "Search" prompt and Cmd+K keyboard shortcut, but do you think it should be any wider than that? It's wider (across the entire width of the TOC) on https://docs.crossplane.io/v1.12/.
Screenshot 2023-04-27 at 11 34 32 PM

Does mobile layout work OK for you? I didn't test on a mobile phone, but just made the browser window very narrow on my laptop and the search box looks like it disappeared.  do you see differently?

plumbis:
>Is it more accurate to say that it resolves https://github.com/crossplane/docs/issues/419 instead

Yes, good catch. Fixed in the description.


>The search box width is wide enough to display the entire "Search" prompt and Cmd+K keyboard shortcut, but do you think it should be any wider than that? 

Sadly a limitation of the current implementation. The search box and popup are "as a service" so we don't have any control over their display or style. Fixing them falls under the work needed for #202


> Does mobile layout work OK for you? 

Testing on my phone with Firefox I see a small search button/magnifying glass. It's not great but same problem as why we can't resize it in normal view.

jeanduplessis (approved):

Issue #422, opened 2023-04-27
Google search results mostly point to old versions


teich:
Search google for a typical thing like `crossplane managed resource` and the results usually point to an older version. Then I need to click in, then realize, then go to the latest version.

<img width="701" alt="Screenshot 2023-04-27 at 2 56 18 PM" src="https://user-images.githubusercontent.com/47476/234999719-191df18f-b970-4aec-a3ea-1510acdfc4ef.png">

PR #424, opened 2023-04-28
create a custom 404 to list sections of available content

### LINKED ISSUE #416: Improve 404 for old docs versions
We remove EOL'd versions of docs when publishing new releases. This makes overall docs maintenance easier and improves search over time.

The downside is that until the search engines update we can have bad search results and customers may have old bookmarks.

## Suggested Fix
Probably keep the old version folder and redirect <old version>/** to a page indicating it's an EOL version and to look for new stuff along with a message on how to find the old release tags.

plumbis:
We delete versions of docs that are EOL'd. The down side is that old bookmarks or bad search results (i.e. #422) can lead to a Netlify default 404 page that leaves you stranded. 

This creates a 404 that provides the search box and lists the available sections.

Screenshot 2023-04-28 at 4 18 53 PM

Resolves #416

### DIFF
--- a/themes/geekboot/layouts/404.html
+++ b/themes/geekboot/layouts/404.html
@@ -1,0 +1,53 @@
+{{ define "main" }}
+<div class="bd-layout docs-container">
+    <aside class="bd-sidebar">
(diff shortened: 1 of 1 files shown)


jeanduplessis (approved):
Thanks @plumbis looks good.

In a future enhancement we could parse the URL for keywords and maybe suggest some content or auto search to show some possible results.

PR #426, opened 2023-04-30
Document patching a composed resource from itself

mamachanko:


plumbis (approved):
Thanks for the contribution!! Approved.

PR #437, opened 2023-05-09
add information about supported k8s versions

plumbis:
[c/c issue #4049](https://github.com/crossplane/crossplane/issues/4049) discusses the fact docs currently list super old versions of k8s as supported. The decision there was we should track to the currently supported versions in the k8s community. 

This adds a notice that we track to the community.

### DIFF
--- a/content/master/software/install.md
+++ b/content/master/software/install.md
@@ -11,8 +11,14 @@ If you don't have a Kubernetes cluster create one locally with [Kind](https://ki
 {{< /hint >}}
 
 ## Prerequisites
-* [Kubernetes](https://kubernetes.io/releases/) version `v1.16.0` or later
+* [Kubernetes](https://kubernetes.io/releases/) version `v1.23.0` or later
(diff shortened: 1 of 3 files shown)


plumbis:
Vale failures are to unrelated content and can be ignored.

jeanduplessis (approved):

content/master/software/install.md
```diff
@@ -11,9 +11,15 @@ If you don't have a Kubernetes cluster create one locally with [Kind](https://ki
 {{< /hint >}}
 
 ## Prerequisites
-* [Kubernetes](https://kubernetes.io/releases/) version `v1.16.0` or later
+* [Kubernetes](https://kubernetes.io/releases/) version `v1.23.0` or later
```
negz, line 14: How good do you feel about keeping up with Kubernetes releases and bumping this version? Would the statement that we support what Kubernetes supports (with a link to their supported versions) be enough?

plumbis: Great point. Opened a new PR for that. https://github.com/crossplane/docs/pull/438

PR #445, opened 2023-05-17
Corrects statement about UXP to Up CLI. Resolves #331

### LINKED ISSUE #331: "Building a configuration package" inaccuracies
### URL

https://docs.crossplane.io/v1.11/getting-started/provider-aws-part-3/#build-a-configuration-package

### What's Wrong?

Screenshot 2023-02-02 at 6 46 38 PM

There's two inaccuracies in the above screenshot:

* I'm pretty sure we mean the [`up`](https://github.com/upbound/up) CLI can be used to build images, not UXP.
* Most config packages contain those three files, but I think the only hard requirement is that they have a `crossplane.yaml` and at least one XRD or Composition.

You could have a config package with 20 XRDs and no Compositions, or vice versa, and I believe the package parser will find them regardless of their filenames, whether they're in a nested directory etc. There's actually a specification that describes all of this but I think it was accidentally deleted when the docs got moved. @hasheddan is working on resurrecting it.

plumbis:
QS incorrectly states you can use UXP for creating packages. Should be `up` CLI. 

Resolves #331

### DIFF
--- a/content/master/getting-started/provider-aws-part-3.md
+++ b/content/master/getting-started/provider-aws-part-3.md
@@ -424,7 +424,7 @@ tool.
 {{< hint "note" >}}
 You can use any software that builds OCI images. This includes
 [Docker](https://www.docker.com/) or 
-[Upbound's Universal Crossplane (UXP)](https://github.com/upbound/universal-crossplane)
+[Upbound's Up CLI)](https://github.com/upbound/up).
 {{< /hint >}}
 
 A configuration package includes three files:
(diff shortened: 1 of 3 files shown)


plumbis:
Vale errors are fixed with style updates in [`99307b0` (#444)](https://github.com/crossplane/docs/pull/444/commits/99307b0cbe1a1d282bf4d7c7eb4422d8fb40bd39)

phisco (approved):

jeanduplessis (approved):

PR #458, opened 2023-05-31
Add more details and organize provider Conditions

plumbis:
Adds a bit more context to the Provider Conditions and puts the descriptions above the output.

### DIFF
--- a/content/v1.12/concepts/providers.md
+++ b/content/v1.12/concepts/providers.md
@@ -197,7 +197,7 @@ Provider `Conditions` support two `Types`:
 * `Type: Healthy` - The Provider package is ready to use. 
 
 #### Reasons
-Each `Reason` is related to a specific `Type` and `Status`. Crossplane uses the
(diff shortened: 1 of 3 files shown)


content/v1.12/concepts/providers.md
```diff
@@ -172,31 +172,49 @@ The Crossplane community has more
...
 ### Provider conditions
 
+Crossplane uses a standard set of `Conditions` for Providers.  
 View the conditions of a provider under their `Status` with 
-`kubectl describe provider`.
+`kubectl describe provider`. 
 
-Providers have the following possible conditions:
+```yaml
+kubectl describe provider
```
jbw976, line 180: this is a truncated output right? is it important to say that some of the output is ommitted for brevity, so as to not confuse readers why their output may be much longer than this sample output here?

i've always wondered about that...

jbw976 (approved):
Looks good to me overall @plumbis! just one question, but it's not blocking so hitting approve now 💪

PR #461, opened 2023-06-02
Add redirects for top 404 pages

plumbis:
A few pages for the v1.9 docs are in existing press about Crossplane that direct to pages that are no longer live.

This creates Netlify redirects for those pages.


jeanduplessis (approved):
Approving and merging to fix current broken links, however, I believe we should update these to 301 status redirects as these are permanent, not temporary redirects.

PR #463, opened 2023-06-12
add an example for claiming existing xr

nabuskey:
This adds an example of claiming existing composite resources. Currently, the documentation instructs readers to update the `spec.resourceRef` field but does not give an example.
This PR does not introduce new vale errors or warnings.

### DIFF
--- a/content/master/concepts/composition.md
+++ b/content/master/concepts/composition.md
@@ -1288,6 +1288,18 @@ an existing Composite Resource as long as its a type of XR that offers a claim
 and no one else has already claimed it. To do so:
 
 1. Set the `spec.resourceRef` of your claim to reference the existing XR.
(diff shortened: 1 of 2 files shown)


jeanduplessis (approved):
Thanks for adding the example @nabuskey

jbw976 (commented):
This looks accurate/correct to me, thanks for making this update @nabuskey!! 💪 

do you want to also make this same edit under `content/v1.12` so the "latest" docs (which defaults to the last release) will show this too?

nabuskey:
I've added the example for v1.12 as well. I have tested this yesterday because a need for this came up and noticed there wasn't an example.

nabuskey:
DCO always gets me :(. Fixed now.

PR #468, opened 2023-06-20
Hugo and other dependency upgrades

jeanduplessis:
This PR proposed bumping the Hugo version to the latest (0.114.0) and updating the PostCSS dependencies to their latest versions.

It further removes the unused Makefile (not referenced in docs or installation steps) so that it doesn't confuse contributors.
Finally it updates the infrastructure page to more accurately reflect the current state.


plumbis (approved):
testing in the deploy preview looks good to me

PR #470, opened 2023-06-20
Update editCode marker to $@

### LINKED ISSUE #370: Update editCode shortcode delimiter
Currently the `editCode` shortcode uses `$$` to mark what can be edited.

The markdown render engine will wrap `$$` in a span tag, but not the content inside. This breaks the shortcode.

Testing shows that `@@` will work as expected.

This is a simple fix but needs to be done with no open PRs so that existing uses of editCode can be updated.

plumbis:
Updates the use of the `editCode` marker. Resolves #370 


jeanduplessis (approved):
Verified by testing the docs preview site.

Issue #450, opened 2023-05-25
Epic: Revamp "Concepts" 


plumbis:
The existing /Concepts  section is dense and doesn't pass Vale linting. The Composition page is too large to index in Algolia. 

This epic is to track the work to rewrite/edit the concepts.

plumbis:
With the merge of #666 I'm considering this epic complete.

PR #474, opened 2023-06-21
Fix copy button to only copy the command and not the output

### LINKED ISSUE #464: [Web Bug] - AWS Quickstart
URL: https://docs.crossplane.io/v1.12/getting-started/provider-aws/

Clicking the copy button copies the output also instead of just copying `kubectl get buckets` in this section: https://docs.crossplane.io/v1.12/getting-started/provider-aws/#create-a-managed-resource

plumbis:
A command in the AWS quickstart didn't limit the copy lines and copies the command and output. 

Resolves #464.

### DIFF
--- a/content/master/getting-started/provider-aws.md
+++ b/content/master/getting-started/provider-aws.md
@@ -1190,7 +1190,7 @@ Crossplane created the bucket when the values `READY` and `SYNCED` are `True`.
 This may take up to 5 minutes.  
 {{< /hint >}}
 
-```shell
+```shell {copy-lines="1"}
 kubectl get buckets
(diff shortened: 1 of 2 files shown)


phisco:
There were a few other occurrences actually, I tried automating that but I still need to go through all the ones found by running the following in the docs repo:
```shell
ag '\`\`\`shell\n(kubectl|helm|sh)[^\`\`\`]*\n([^\`\`\`]+\n)+\`\`\`'
```

plumbis:
@phisco, yeah, there are a lot that are from legacy outputs. As docs are being rewritten and edited I'm fixing these up. It's similar to the existing vale errors.

With the work in #450 we should clean up a large majority of these.

phisco (approved):

Issue #448, opened 2023-05-22
Document XRD Conditions


plumbis:
### What's Missing?
Capture the full list of XRD conditions similar to what's in the [Provider docs](https://docs.crossplane.io/latest/concepts/providers/#provider-conditions)

Issue #432, opened 2023-05-03
Best-practices for deterministic composed resource names


negz:
### What's Missing?

I believe one of the problems https://github.com/crossplane/crossplane/pull/3287 intends to fix is ensuring the _external_ names of composed resources are deterministic. The idea being that if a claim was deleted and then recreated, it would result in the same set of external resources (with the same names, in the external system) being recreated.

This is useful for migrating a claim from one control plane to another. It should be possible to:

1. Ensure all composed MRs have `deletionPolicy: Orphan` so that deleting the MR does not delete the external resource.
2. Delete the claim from the 'old' control plane.
3. Create an identical claim in the 'new' control plane.
4. Have Crossplane find and adopt all of the external resources that were orphaned in step 1, as composed resources of the new claim and XR.

In theory this should be possible today, but it requires some forethought to make it work. Specifically you must ensure all composed resources have a `crossplane.io/external-name` annotation that is deterministically derived from the claim. One way to do this is by patching from the XR's `crossplane.io/claim-name` annotation as suggested in https://github.com/crossplane/crossplane/issues/3120#issuecomment-1178377398.

I believe we should document this pattern as a Composition best-practice. It doesn't have to specifically be patching from `claim-name`, but rather being thoughtful and explicit about how your `external-name` annotations are specified. If you leave this up to Crossplane it will just copy the `metadata.name` of your composed resources, which will have a random suffix and thus be non-deterministic.

Issue #288, opened 2022-11-07
Restart XR and claim controllers when XRDs are updated


muvaf:
### What problem are you facing?

I might be the only one who knows about this since I did the initial implementation so we need to document it. The reason is that we spin up a new controller for every CRD we produce but those controllers work on the cached schema, so the schema cache needs to be refreshed and that doesn't happen unless you restart the pod (the last I checked).

### How could Crossplane help solve your problem?

Add this to composition reference doc.

plumbis:
Fixed in #475

PR #475, opened 2023-06-22
Update Composition Docs

### LINKED ISSUE #281: Documentation bug on label crossplane.io/claim-name*
### What happened?

The documentation explains the existence of a annotation crossplane.io/claim-name and crossplane.io/claim-namespace.
These are labels and not as described in the docs annotations.

### What environment did it happen in?
Crossplane version: 1.8

For docs v1.8
* https://github.com/crossplane/crossplane.github.io/blob/master/docs/v1.8/reference/composition.md?plain=1#L413
* https://github.com/crossplane/crossplane.github.io/blob/master/docs/v1.8/reference/composition.md?plain=1#L420

For docs v1.9
* https://github.com/crossplane/crossplane.github.io/blob/master/docs/v1.9/reference/composition.md?plain=1#L441
* https://github.com/crossplane/crossplane.github.io/blob/master/docs/v1.9/reference/composition.md?plain=1#L448

### LINKED ISSUE #305: User guide for EnvironmentConfig
The ability to use configuration values from the environment is a great new feature in v1.11: https://github.com/crossplane/crossplane/issues/2099.

We have some guidance on how to use them from a composition patching perspective, e.g. `*Environment*` in https://docs.crossplane.io/master/reference/composition/#patch-types, but it would also be helpful to have a complete end to end user guide for how to create and use EnvironmentConfig.

There was a small yaml only (no written prose or context) end to end example in the old docs snippets (see links below), but those are being cleaned up and removed.  We could use those as a starting point to author a complete end to end guide on how to use EnvironmentConfig.

* Discussion and context for this issue: [discussion link](https://github.com/crossplane/docs/pull/271#discussion_r1085960840)
* Commit with example YAML: https://github.com/crossplane/docs/commit/4c35456406dd282f0688d83f063ad55b8630f1f2#diff-d9549ed3cf00d1a20eb5fc1c51194ead6078f9e48cc62c62c8e3284c3306b72f

### LINKED ISSUE #380: [Web Bug] - Composite Resources
The example [XPostgreSQLInstance ](https://docs.crossplane.io/v1.11/concepts/composition/#composite-resources-and-claims) contains the following yaml snippet which should rather belong to the Composition than the Composite Resource. The XR only contains the `environmentConfigRefs` if I understand this correctly?

In addition some example of `EnvironmentConfig` CR would be useful (like this [design doc](https://github.com/crossplane/crossplane/blob/master/design/one-pager-composition-environment.md)) with more clarity on whether the environment `xxxFieldPath` includes `data.x.y.z` or simply `x.y.z` (the examples provided: `tier.name`, `location`... seem to point to the latter).

IMHO EnvironmentConfig might deserve its own section (a shortened version of the design doc) to complement the information available in the [patch types](https://docs.crossplane.io/v1.11/concepts/composition/#patch-types) section.
```
environment:
    # EnvironmentConfigs is a list of object references that is made up of
    # name references and label selectors
    environmentConfigs:
      - type: Reference # this is the default
        ref:
          name: example-environment
      - type: Selector
        selector:
          matchLabels:
          - key: stage
            type: FromCompositeFieldPath # this is the default
            valueFromFieldPath: spec.parameters.stage
          - key: provider
            type: Value
            value: "gcp"
```
URL: /v1.11/concepts/composition/

### LINKED ISSUE #455: Update concepts/composite resources
Update the concepts/composite resources doc.

Likely breaking into component pages. 

Link to epic #450

### LINKED ISSUE #456: Clearly state that connectionSecretKeys is immutable
There is confusion around `connectionSecretKeys`. There are two spots where it's discussed today:
https://docs.crossplane.io/v1.12/concepts/composition/#hl-6-22
https://docs.crossplane.io/v1.12/concepts/composition/#composite-resource-connection-secrets

Make sure to clarify that this is immutable in the Composite Resources update.

plumbis:
This PR rewrites the docs related to Composition.
 
The new [/concepts/compositions](https://deploy-preview-475--crossplane.netlify.app/master/concepts/compositions/) page streamlines the details and focuses on the fields available in a Composition and what they do.

New [XRD](https://deploy-preview-475--crossplane.netlify.app/master/concepts/composite-resource-definitions/) page breaks out info specific to writing XRDs and XRD status. Resolves #456, #448, #288, 

New [XR](https://deploy-preview-475--crossplane.netlify.app/master/concepts/composite-resources/) page details XR options.  Resolves #455, (partially) #370, #432.

New [Claims](https://deploy-preview-475--crossplane.netlify.app/master/concepts/claims) page details creating a Claim and claiming an existing XR. Resolves #281.
 
New [Patch and Transform](https://deploy-preview-475--crossplane.netlify.app/master/concepts/patch-and-transform) page covering patching options and configurations in detail. Resolves #380

New documentation on [EnvironmentConfigs](https://deploy-preview-475--crossplane.netlify.app/master/concepts/environment-configs).  Resolves #305

Additionally composite functions has been moved out of the KB and into `/concepts`

New KB on [connection details](https://deploy-preview-475--crossplane.netlify.app/knowledge-base/guides/connection-details/).

Reviewed:
| Page | Reviewer1 | Reviewer2 | 
| --- | --- | --- | 
| [Compositions](https://deploy-preview-475--crossplane.netlify.app/master/concepts/compositions/)  |  @negz | | 
| [XRD](https://deploy-preview-475--crossplane.netlify.app/master/concepts/composite-resource-definitions/) | @negz  | | 
| [XR](https://deploy-preview-475--crossplane.netlify.app/master/concepts/composite-resources/) | @negz | | 
| [Claims](https://deploy-preview-475--crossplane.netlify.app/master/concepts/claims) | @negz | | 
| [Patch and Transform](https://deploy-preview-475--crossplane.netlify.app/master/concepts/patch-and-transform) | @jbw976 | | 
|  [EnvironmentConfigs](https://deploy-preview-475--crossplane.netlify.app/master/concepts/environment-configs) | @jbw976  | |
| [connection details](https://deploy-preview-475--crossplane.netlify.app/knowledge-base/guides/connection-details/) | @jbw976  | | 


negz:
@plumbis Should I set this to ready for review? I think it is, right?

plumbis:
The published pages mentioned in the first comment are all ready for review. There are still at least two pages to write/edit before this would be ready to merge.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+with all the provided 
+{{<hover label="resources" line="17">}}spec.forProvider{{</hover>}} settings.
+
+The {{<hover label="resources" line="17">}}spec.forProvider{{</hover>}}
+supports any settings used in a managed resource including
+applying `annotations` and `labels` or using a specific 
+`providerConfigRef`.
+
+{{<hint "note" >}}
+Compositions allow applying a `metadata.name` but ignores it. The
```
negz, line 121: Are you sure this is true? I thought they'd let you specify it.

Either way, I'd consider specifying your own `metadata.name` to be an anti-pattern. If it is actually possible you'll need to make sure they're somehow unique, so that three XRs using the same Composition don't try to make three composed MRs with the same name.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+            namespace: crossplane-system
+            name: aws-secret
+            key: creds
+```
+
+
+#### Template another Composite Resource
+
+Compositions can use existing Composite Resources to define more complicated
+templates.
```
negz, line 158: "Existing" reads to me like you already have to have _created_ the XR you want to compose. Is it worth finding another name? "Defined" seems correct to me but maybe a bit academic.

Maybe just "other kinds of Composite Resources"?

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+
+
+#### Template another Composite Resource
+
+Compositions can use existing Composite Resources to define more complicated
+templates. 
+
+A common use case is a Composition that uses other
+Compositions. For example, creating a Composition to create a standard set of 
+networking resources that other Compositions reference.
```
negz, line 162: Technically it's a Composition that composes XRs. Each of the XRs then has a Composition. I think the reason this is important to note is that each composed XR could be compatible with more than one Composition.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+  compositeTypeRef:
+    apiVersion: aws.platformref.upbound.io/v1alpha1
+    kind: XNetwork
+```
+
+The {{<hover label="xnetwork" line="20">}}compositeTypeRef{{</hover>}} gives
+this Composition a 
+{{<hover label="xnetwork" line="21">}}apiVersion{{</hover>}} and
+{{<hover label="xnetwork" line="22">}}kind{{</hover>}} to reference in another
+Composition.
```
negz, line 199: This is an interesting way to frame it. Not wrong, though. Maybe clarify that it specifies a type of XR another Composition can reference?

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+
+{{<hint "note" >}}
+This abbreviated example is from the Upbound [AWS Reference
+Platform](https://github.com/upbound/platform-ref-aws).
+
+View the complete Compositions in the reference platform's 
+[package directory](https://github.com/upbound/platform-ref-aws/blob/main/package/cluster/composition.yaml).
+{{</hint >}}
+
+#### Cross-reference resources
```
negz, line 241:
```suggestion
#### Cross-resource references
```

😄

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+
+Inside a Composition some resources use identifiers or names of other resources.
+For example, creating a new `network` and applying the network identifier to a
+virtual machine. 
+
+Resources inside a Composition can cross-reference other resources by matching
+a label or a _controller reference_.
```
negz, line 248: Exposing the concept of controller references in this API is one of my many regrets. 😬

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+
+Matching only a controller reference simplifies the matching process without
+requiring labels or more information. 
+
+For example, creating an AWS 
+{{<hover label="controller1" line="14">}}InternetGateway{{</hover>}} requires a
+{{<hover label="controller1" line="7">}}VPC{{</hover>}}.
+
+The {{<hover label="controller1" line="14">}}InternetGateway{{</hover>}} could
+match a label, but every VPC created by this Composition share the same label.
```
negz, line 315:
```suggestion
match a label, but every VPC created by this Composition will share the same label. 
```

Looks like a missing word here.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+          toFieldPath: vpcId
+        - type: FromEnvironmentFieldPath
+          fromFieldPath: tags
+          toFieldPath: spec.forProvider.tags
+```
+
+The [EnvironmentConfigs]({{<ref "./environment-configs" >}}) page has 
+more information on EnvironmentConfigs options and usage.
+
+### Composite functions
```
negz, line 607:
```suggestion
### Composition functions
```

Here and elsewhere. 🙂

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+
+Composite functions (`XFNs`) are containers executing any code you define.
```
negz, line 609: FWIW I'm not 100% sure we want to brand these as "xfns" - that has been kind of organic. If we do I think we'd go with lowercase `XFn` or `xfn`.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+Composite functions can read and write to any resource in the Composition
+they're attached to. This allows composite functions to preform complex
```
negz, line 611:
```suggestion
they're attached to. This allows composite functions to perform complex
```

content/master/concepts/compositions.md
negz: Also maybe gloss over the whole "are containers" thing given that we're planning to tweak that for beta.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+Read the
+[Managed Resources]({{<ref "./managed-resources#writeconnectionsecrettoref" >}})
+page for more information about connection secrets.
+{{</hint >}}
+
+The Composition defines where to save the details generated by any resources.
+#### Save connection details to a namespace
+By default Crossplane saves connection details in the `default` namespace. A
+Composition can save connection details to a specific namespace with
+`writeConnectionSecretsToNamespace`.
```
negz, line 664: I believe what's this is configuring is a little more niche.

`writeConnectionSecretsToNamespace` specifies what namespace an XR (only - not its composed MRs) should write its connection details to if that XR is automatically created by Crossplane because someone created a claim.

If you create an XR directly/explicitly you must set `spec.writeConnectionDetailsTo.namespace` on the XR. If you're writing a Composition you must specify `spec.writeConnectionDetailTo.namespace` for any composed MR.

It's only when you create a claim that we don't really know what namespace the XR should use to store its secrets, so we derive it from the Composition.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+Composition. 
+{{< /hint >}}
+
+## Composition validation
+
+When creating a Composition Crossplane automatically validates specific
+parameters in the Composition.
+
+* All resources either use a `name` or don't. Compositions can't use both named
+  and unnamed resources.
```
negz, line 936: This is another best practice - always use names. I regret allowing resources to _not_ be named per https://github.com/crossplane/crossplane/issues/4130.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+validation doesn't validate the fields of a specific resource.
+<!-- vale write-good.TooWordy = YES -->
+{{< /hint >}}
+
+The default validations are still checked with schema validation enabled. 
+#### Validation modes
+
+Schema validation supports two modes:
+* `loose` (default) - Sends an warning for each schema error and installs the 
+  Composition if the default validations pass.
```
negz, line 974: Where would we see the warnings?

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,524 @@
...
+external system, for example the actual name of a `CloudSQLInstance` in the GCP
+API. Some managed resources don't let you specify an external name - in those
+cases Crossplane will set it for you to whatever the external system requires.
+
+If you add the `crossplane.io/external-name` annotation to a claim Crossplane
+will automatically propagate it when it creates an XR. It's good practice to
+have your `Composition` further propagate the annotation to one or more composed
+resources, but it's not required.
+
+where `MyResource` is a Composite Resource Claim kind.
```
negz, line 19: Looks like something might have gone wrong here?

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,524 @@
...
+non-`true` value, reconciliations on the Claim will now resume, and thus the
+annotation will now be propagated to the associated Composite Resource
+with a non-`true` value. An implication of the described behavior is that
+pausing reconciliations on the Claim will not inherently pause reconciliations
+on the associated Composite Resource.
+
+<!-- This also related to XRs -->
+### Connection Details
+
+Connection details secret of XR is an aggregated sum of the connection details
```
negz, line 41: Should this be "The connection details secret"?

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,524 @@
...
+`FromConnectionSecretKey`. Derives an XR connection detail from a connection
+secret key of a composed resource.
+
+```yaml
+# Derive the XR's 'user' connection detail from the 'username' key of the
+# composed resource's connection secret.
+- type: FromConnectionSecretKey
+  name: user
+  fromConnectionSecretKey: username
+```
```
negz, line 62: Might be worth showing the larger context for one of these examples so its more clear where you put this configuration.

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,524 @@
...
+- type: FromValue
+  name: user
+  value: admin
+```
+
+<!-- break --> 
+
+You'll encounter a lot of 'field paths' when reading or writing a `Composition`.
+Field paths reference a field within a Kubernetes object via a simple string
+'path'. [API conventions][field-paths] describe the syntax as:
```
negz, line 88: This link isn't rendering right.

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,524 @@
...
+* `spec.containers[]` - Empty brackets.
+* `spec.containers.[0].name` - Period before open bracket.
+
+### Patch Types
+
+You can use the following types of patch in a `Composition`:
+
+`FromCompositeFieldPath`. The default if the `type` is omitted. This type
+patches from a field within the XR to a field within the composed resource. It's
+commonly used to expose a composed resource spec field as an XR spec field.
```
negz, line 115: Interesting framing. Not wrong, but FWIW I think of it as "copying a field value from the XR to the composed resource".

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,524 @@
...
+
+```yaml
+# Patch from the XR's spec.parameters.size field to the composed resource's
+# spec.forProvider.settings.tier field.
+- type: FromCompositeFieldPath
+  fromFieldPath: spec.parameters.size
+  toFieldPath: spec.forProvider.settings.tier
+```
```
negz, line 123: Note to self: I reviewed every file but this one, but only got this far down this one before I had to run.

negz (commented):
Thanks @plumbis! I like where this is headed.

None of my comments are major or blocking - happy for someone else to approve this once you're ready.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+    container:
+      image: xpkg.upbound.io/provider-aws-xfns/random-rds-password:v0.1.0
+```
+
+Read the [composition functions]({{<ref "./composition-functions">}}) page for
+details on building and using composite functions.
+
+
+
+### Storing connection details
```
jbw976, line 648: I would expect in this section that we are talking about `connectionDetails` of each child composed resource - that field is important because it specifies what/how to derive connection details to send up the chain from each child.  Is it mentioned anywhere? This section feels like the right place for it...

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,808 @@
...
+  versions:
+  # Removed for brevity
+```
+
+{{<hint "important" >}}
+You can't change the `claimNames` after they're defined. You must delete and
+recreate the XRD to change the `claimNames`.
+{{</hint >}}
+
+### Manage connection secrets
```
jbw976, line 538: From what I've seen so far, it looks like we are breaking up the steps/details of how to specify connection information into each specific component (XRDs, XRs, Compositions, etc.)

What I really like about https://github.com/crossplane/docs/pull/485 is that it has an overall summary of the entire flow across all components, which gives the reader a single place to understand how it all comes together.

Do we have that single summary in this PR? of if so, can we make it more discoverable/findable? 🤔

content/master/concepts/composite-resources.md
```diff
@@ -0,0 +1,470 @@
...
+spec:
+  compositionRevisionSelector:
+    matchLabels:
+      channel: dev
+  # Removed for brevity
+```
+
+### Manage connection secrets 
+
+When a composite resource creates managed resources, Crossplane provides any
```
jbw976, line 276: i was using the terminology "child composed resource" in https://github.com/crossplane/docs/pull/485 and https://blog.crossplane.io/faq-2-claim-connection-details/, because the composition could be creating MRs but also nested XRs as well.  Do you want to make that same distinction or use similar terminology?

jbw976 (commented):
Just took a pass specifically through the connection secret details in the context of https://github.com/crossplane/docs/pull/485, that's all I looked at right now 🙇‍♂️

content/master/concepts/claims.md
```diff
@@ -0,0 +1,169 @@
...
+    kind: xMyDatabase
+    plural: xmydatabases
+  claimNames:
+    kind: Database
+    plural: databases
+  # Removed for brevity
+```
+
+The Claim uses the XRD's 
+{{<hover label="xrd1" line="10">}}claimNames{{</hover>}} to request resources.
```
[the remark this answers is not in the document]
plumbis: I changed this to "uses the XRD's `kind` API endpoint to request resources"

I'm purposefully vague with "request resources" to avoid unnecessary details about the underlying XR

content/master/concepts/claims.md
```diff
@@ -0,0 +1,169 @@
...
+apiVersion: example.org/v1alpha1
+kind: database
+metadata:
+  name: my-claimed-database
+spec:
+  # Removed for brevity
+```
+
+When Crossplane creates a Claim in a namespace it also creates a composite
+resource.
```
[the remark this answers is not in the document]
plumbis: The problem with "you" here is that it's not necessarily the reader.  Flipping between terms (you, user, developer, etc) makes the reader have to "keep state" against who is doing what.

utils/vale/styles/Google/Headings.yml
```diff
@@ -10,18 +10,24 @@ exceptions:
...
   - BigQuery
   - Claim
+  - Claims
   - CLI
   - ClusterRoles
   - Code
+  - Composite Resource
   - Composite Resource Definitions
   - Composite Resources
+  - CompositeFunctions
```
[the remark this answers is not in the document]
plumbis: This was an error I made a bunch and realized much later (after adding the spelling exception) that it was wrong. Good catch, thanks! 

No where in the docs do we reference `CompositeFunctions`. Vale should still pass with this fix.

content/master/concepts/environment-configs.md
```diff
@@ -0,0 +1,67 @@
...
+https://github.com/crossplane/crossplane/issues/3723
+
+Updating the Environment in a Composition won't push down to an XR
+https://github.com/crossplane/crossplane/issues/3941
+
+Can't add to a package
+https://github.com/crossplane/crossplane/issues/3865
+discussion: https://github.com/crossplane/crossplane/pull/3955#pullrequestreview-1377419076
+
+Propagate status from MR to Environmetn
```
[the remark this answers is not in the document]
plumbis: This file hasn't been worked yet. Vale catches this so no risk of pushing this error.

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,808 @@
...
+connection secrets. 
+
+## Creating a CompositeResourceDefinition
+
+Creating a CompositeResourceDefinintion consists of:
+* [Defining a custom API group](#xrd-groups).
+* [Defining a custom API name](#xrd-names).
+* [Defining a custom API schema and version](#xrd-versions).
+  
+Optionally, CompositeResourceDefinintions also support:
```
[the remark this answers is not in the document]
plumbis: Yes it should. I don't have a good answer for why it's not matching this while other spelling errors are caught. Opened #490 to track.

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,808 @@
...
+```
+
+### XRD groups
+
+Groups define a collection of related API endpoints. The `group` can be any
+value, but common convention is to map to a fully qualified domain name.
+
+<!-- vale write-good.Weasel = NO -->
+Many XRDs may use the same `group` to create a logical collection of APIs. 
+<!-- vale write-good.Weasel = YES -->
```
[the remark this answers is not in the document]
plumbis: Added the line

>For example a `database` group may have a `relational` and `nosql` kinds.

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,808 @@
...
+resources and claims. 
+
+Using a `defaultCompositeDeletePolicy: Background` policy deletes 
+the composite resource or Claim and relies on Kubernetes to delete the remaining
+dependent objects, like managed resources or secrets. 
+
+Using `defaultCompositeDeletePolicy: Foreground` causes Kubernetes to attach a
+`foregroundDeletion` finalizer to the composite resource or Claim. Kubernetes
+deletes all the dependent objects before deleting the composite resource or
+Claim.
```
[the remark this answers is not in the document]
plumbis: I left that out on purpose as hand-wavy Kube stuff. Not important to know the details but now they know what to look for if they want to go find more.

content/master/concepts/composite-resources.md
```diff
@@ -0,0 +1,470 @@
...
+    API Version:  dynamodb.aws.upbound.io/v1beta1
+    Kind:         Table
+    Name:         my-composite-resource-wnr9t
+# Removed for brevity
+```
+
+### Composite resource conditions
+
+The conditions of composite resources match the conditions of their managed 
+resources.
```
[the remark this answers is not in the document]
plumbis: I couldn't find it in code and didn't see them differ in testing. Where does the XR condition come from?

content/master/concepts/composite-resources.md
```diff
@@ -0,0 +1,470 @@
...
+Read the 
+[conditions section]({{<ref "./managed-resources#conditions">}}) of the
+managed resources documentation for details.
+
+## Composite resource labels
+
+Crossplane adds labels to composite resources to indicates their relationship to
+other Crossplane components.
+
+### Composite label
```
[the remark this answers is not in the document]
plumbis: There is planned work on style improvements.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+  Composition template.
+
+Optionally, Compositions also support:
+* [Modifying and patching](#changing-resource-fields) resource settings.
+* [Storing connection details](#storing-connection-details) and secrets
+  generated by the managed resources.
+* Using [Composite functions](#composite-functions) to allow custom programs to
+  run alongside the Composition. 
+* Creating a 
+  [custom definition of when a resource is ready](#resource-readiness-checks) to use.
```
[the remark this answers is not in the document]
plumbis: changed to "custom check"

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+  run alongside the Composition. 
+* Creating a 
+  [custom definition of when a resource is ready](#resource-readiness-checks) to use. 
+ 
+###  Resource templates
+The 
+{{<hover label="resources" line="4">}}resources{{</hover>}} field of a 
+Composition's 
+{{<hover label="resources" line="3">}}spec{{</hover>}} 
+defines the set of things that a Composite Resource can create.
```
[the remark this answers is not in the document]
plumbis: will is future tense and isn't used in the docs.
"should" implies "maybe it will, maybe it won't, let's hope it works 🤞"

Using "can" puts it in the context of expected, present tense, behavior.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,993 @@
...
+{{<hover label="resources" line="8">}}kind{{</hover>}}, define the 
+{{<hover label="resources" line="10">}}spec.forProvider{{</hover>}} fields
+defining the settings of the resource.
+
+```yaml {copy-lines="none",label="resources"}
+apiVersion: apiextensions.crossplane.io/v1
+kind: Composition
+spec:
+  resources:
+    - name: StorageBucket
```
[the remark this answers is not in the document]
plumbis: names as kebab case and kinds/types as Pascal? 

I have to say trying to track all the camel vs pascal vs kebab conventions sucks. 

For example, half of the patch type values are Pascal Case, i.e., `string.type.TrimPrefix` while others are all lowercase, i.e., `match.pattern.type: literal`

content/master/concepts/compositions.md
plumbis: In testing and what I could see from code that it's ignored.

content/master/concepts/compositions.md
plumbis: No, it's just present tense. What is the state when the VPCs are created? They share a label.

plumbis: https://developers.google.com/style/tense

content/master/concepts/compositions.md
plumbis: ctrl + f'd away.

content/master/concepts/patch-and-transform.md
plumbis: page wasn't ready for review yet.

content/master/concepts/patch-and-transform.md
plumbis: Page not ready for review yet.

content/master/concepts/patch-and-transform.md
plumbis: Page not ready for review yet. 

The new updates provide more comprehensive examples with full example XRDs, Compositions and Claims provided.

content/master/concepts/patch-and-transform.md
plumbis: Page not ready for review yet.

content/master/concepts/patch-and-transform.md
plumbis: Page not ready for review yet. 

This section actually comes from the [existing documentation](https://docs.crossplane.io/v1.12/concepts/composition/#hl-7-60).

content/master/concepts/compositions.md
plumbis: What do you have in mind? I avoid specifics here and redirect to the connection secrets part of the MRs pages since all of the fields come from the managed resource.

content/master/concepts/composite-resource-definitions.md
plumbis: I think there is a TON of value in what you're spelling out, but I think it's a "thing to do next" item.

By breaking up all of the concepts like this mega PR does, we lose that "all in one place" reference. I 100% believe we need to create a more comprehensive deployment example that can pull a thread through all of the components.

content/master/concepts/composite-resources.md
plumbis: I tried to avoid calling out "managed resources" for that reason but since we're talking about connection secrets only the MR can generate them, right? 

If I'm nesting XRs can the XR generate secrets?

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,808 @@
...
+
+
+#### Multiple schema versions
+
+{{<hint "warning" >}}
+Crossplane supports defining multiple `versions`, but the schema of each version
+can't change any existing fields, also called "making a breaking change."
+
+Breaking schema changes between versions requires the use of [conversion
+webhooks](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion).
```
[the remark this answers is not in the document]
plumbis: I didn't want to really employ scare tactics. I'm hoping looking at that k8s docs page will be enough for a user to go "oh, I don't want to do this"

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,808 @@
...
+
+New versions may define new optional parameters, but new required fields are
+a "breaking change."
+
+Crossplane XRDs use Kubernetes custom resource definitions for versioning. 
+Read the Kubernetes documentation on [versions in
+CustomResourceDefinitions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/)
+for more background on versions and breaking changes. 
+
+Crossplane recommends implementing breaking schema changes as brand new XRDs.
```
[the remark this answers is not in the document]
plumbis: I wanted to take a more opinionated stance. If you are sophisticated enough to build a conversion webhook you know you're in the backcountry anyway.

content/master/concepts/composite-resources.md
```diff
@@ -0,0 +1,470 @@
...
+The {{<hover label="comp" line="11">}}fromFieldPath{{</hover>}} patch copies the
+{{<hover label="comp" line="11">}}metadata.annotations{{</hover>}} field from
+the composite resource to the 
+{{<hover label="comp" line="12">}}metadata.annotations{{</hover>}} inside the
+managed resource. 
+
+{{<hint "note" >}}
+If a managed resource has the `crossplane.io/external-name` annotation
+Crossplane uses the annotation value to name the external resource.
+{{</hint >}}
```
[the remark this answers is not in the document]
plumbis: Ack, I'll cover that in a future doc/revision. Tracked in #432

content/master/concepts/composite-resources.md
```diff
@@ -0,0 +1,470 @@
...
+    name: my-other-composition
+  # Removed for brevity
+```
+
+A composite resource can also select a Composition based on labels instead of 
+the exact name with a
+{{<hover label="complabel" line="6">}}compositionSelector{{</hover>}}.
+
+Inside the {{<hover label="complabel" line="7">}}matchLabels{{</hover>}} section
+provide one or more Composition labels to match.
```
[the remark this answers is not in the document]
plumbis: https://github.com/crossplane/docs/issues/492

content/master/concepts/composite-resources.md
```diff
@@ -0,0 +1,470 @@
...
+    matchLabels:
+      environment: production
+  # Removed for brevity
+```
+
+### Composition revision policy
+
+If Crossplane uses [Composition revisions]({{<ref "/knowledge-base/guides/composition-revisions">}}) a composite resource can use
+a {{<hover label="comprev" line="6">}}compositionUpdatePolicy{{</hover>}} to
+manually or automatically reference newer Composition revisions.
```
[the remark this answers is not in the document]
plumbis: Updated.

content/master/concepts/composite-resources.md
jbw976: yes the nested XRs can generate/propagate their own connection details up the chain too!

here's a tangible example in platform-ref-aws where the XEKS composed resource is delcaring/propagating connection details:
https://github.com/upbound/platform-ref-aws/blob/main/package/cluster/composition.yaml#L28-L32

content/master/concepts/compositions.md
jbw976: mainly the `connectionDetails` field for composed resources is what i'm thinking of here - that isn't a general MR thing, it's very specific for child composed resources, and it's critical to getting the connection secrets wired up from top to bottom 😉 - feels like it needs to be somewhere, but i can't find it anywhere when searching around now.

See https://docs.crossplane.io/v1.12/concepts/composition/#composite-resource-connection-secrets as a reference for how it is all wired together.  Each of those parts should be mentioned in their relevant section for the whole thing to work.

content/master/concepts/composite-resource-definitions.md
jbw976: do you want to track it with a new issue then? I think that after the effort to pull all the steps together to get connection details working in https://docs.crossplane.io/v1.12/concepts/composition/#composite-resource-connection-secrets, it's a bit of a step backwards to lose that top to bottom summary, so at least having a tracking issue would be my request 🙇

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+{{<hint "important" >}}
+A PatchSet can't contain other PatchSets.  
+
+Crossplane ignores any [transforms](#transform-a-patch) or
+[policies](#patch-policies) in a PatchSet.
+{{< /hint >}}
+
+## Patching between resources
+
+Compositions can't patch between resources in the same Composition. For example,
```
jbw976, line 216: This scenario actually can be done using bi-directional patching with the composed resources and a status field on the parent composite resource. I'll send you a link to a training deck that goes through that scenario ✅ 

There isn't a need to use 3 compositions here, this is done in platform-ref-aws specifically for modularity - the network resource can stand on its own without a cluster, so it needs to be its owns separate XR/Composition.

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+
+## Types of patches
+Crossplane supports multiple patch types, each using a different source for data
+and applying the patch to a different location. 
+
+Summary of Crossplane patches
+{{< table "table table-hover" >}}
+| Patch Type | Data Source | Data Destination | 
+| ---  | --- | --- | 
+| [FromCompositeFieldPath](#fromcompositefieldpath) | A field in the composite resource. | A field patched managed resource. |
```
jbw976, line 349:
```suggestion
| [FromCompositeFieldPath](#fromcompositefieldpath) | A field in the composite resource. | A field in the patched managed resource. | 
```

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+<!-- vale Google.Headings = NO -->
+### FromCompositeFieldPath
+<!-- vale Google.Headings = YES -->
+
+The 
+{{<hover label="fromComposite" line="12">}}FromCompositeFieldPath{{</hover>}}
+patch takes a value in a composite resource and applies it to a field in the 
+managed resource. 
+
+{{< hint "tip" >}}
```
jbw976, line 488: this is great context to the user about the practical scenario this type of patch is used for

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+
+For example, use the Claim value 
+{{<hover label="combineFromComp" line="15">}}desiredRegion{{</hover>}} and 
+{{<hover label="combineFromComp" line="16">}}field2{{</hover>}} to generate the
+managed resource's
+{{<hover label="combineFromComp" line="20">}}name{{</hover>}}
+
+The 
+{{<hover label="combineFromComp" line="12">}}CombineFromComposite{{</hover>}}
+patch only supports the
```
jbw976, line 624: for a few of these patch types, we say that it only supports X option, then further down after another paragraph we say that it only supports Y strategy.  Should these two sentences be combined into the same paragraph, so the context of "limitations" for this patch type is all grouped together?

e.g., The CombineToComposite patch only supports the `combine` option and the only supported `strategy` is `strategy: string`.

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+          fromFieldPath: spec.forProvider.region
+          toFieldPath: key1
+```
+
+Because the environment is in-memory, there is no command to confirm the patch
+wrote the value to the environment.
+
+{{<hint "important">}}
+The 
+{{<hover label="toEnvField" line="12">}}ToEnvironmentFieldPath{{</hover>}}
```
jbw976, line 872: this is a bit surprising to me - did you get a sense of if this was intentional, or is it a bug in the behavior of `EnvironmentConfig`?

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+The {{<hover label="map" line="6">}}map{{</hover>}} transform is useful for
+translating generic region names like `US` or `EU` to provider specific region
+names. 
+{{< /hint >}}
+
+The {{<hover label="map" line="6">}}map{{</hover>}} transform compares the value
+from the {{<hover label="map" line="3">}}fromFieldPath{{</hover>}} to the
+options listed in the {{<hover label="map" line="6">}}map{{</hover>}}.  
+
+If Crossplane finds the value, Crossplane puts
```
jbw976, line 1208: What happens if Crossplane doesn't find the value? 🤔

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+#### clampMax
+<!-- vale Google.Headings = YES -->
+
+The {{<hover label="clampMax" line="8">}}type: clampMax{{</hover>}} uses a defined 
+minimum value if an input is larger than the 
+{{<hover label="clampMax" line="8">}}type: clampMax{{</hover>}} value.
+
+For example, this 
+{{<hover label="clampMax" line="8">}}type: clampMax{{</hover>}} requires an
+input to be greater than
```
jbw976, line 1465: i think this is supposed to be "less than" 🤔

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+
+<!-- vale Google.Headings = NO -->
+### fromFieldPath policy
+<!-- vale Google.Headings = YES -->
+
+Using a `fromFieldPath: Required` policy on a patch requires the
+`fromFieldPath` to exist in the composite resource.
+
+{{<hint "tip" >}}
+If a resource patch isn't working applying the `fromFieldPath: Required` policy
```
jbw976, line 1649: should we be explaining `fromFieldPath: Optional` here also? (that's a thing right? 😂)

jbw976 (commented):
Took a full pass through `patch-and-transform.md` today and left some comments, looking very good @plumbis!

content/master/concepts/patch-and-transform.md
plumbis: MATH IS HARD, OKAY!? (also fixed)

content/master/concepts/patch-and-transform.md
plumbis: It seem intentional? It's mentioned in the existing docs
https://github.com/crossplane/crossplane/issues/3926#issuecomment-1578886896

But I also think it's a poor user experience and drove me a little crazy on trying to figure out what can and can't be patched.

content/master/concepts/patch-and-transform.md
plumbis: Can you give me another example? For this one (and likely the others) I keep them on unique lines to try and make it easier to find the one line/argument/whatever when you're skimming or to be able to understand each line (almost like an API doc) if you're building it for the first time. Open to changing.

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+Field selectors can be a string matching a field name or an array index, in
+brackets. Field names may use a `.` character to select child elements.
+
+
+{{< expand "Example field selectors" >}}
+Here are some example selectors from a composite resource object.
+{{<table "table table-hover" >}}
+| Selector | Selected element | 
+| --- | --- |
+| `kind` | {{<hover label="select" line="3">}}kind{{</hover>}} |
```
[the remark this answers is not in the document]
plumbis: (1) This is going to require some gross hacking since the table and table highlighting is generated by our CSS library (bootstrap) while the code highlight is a hugo injected thing. I don't think we could do this without making some custom table codes or something. For now I turned off the row highlighting to make it more likely the user will hover the keyword.

(2) ✅

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1692 @@
...
+weight: 70
+description: "Crossplane Compositions use patches and transforms to modify inputs from claims and composite resources before creating managed resources"
+---
+
+Crossplane Compositions allow for "patch and transform" operations. With patches
+a Composition can apply changes to a composite resource or managed resource. 
+
+When users create Claims, Crossplane passes the settings in the Claim to
+the associated composite resource. Patches can use these settings to change the
+associated composite resource or managed resources.
```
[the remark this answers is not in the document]
plumbis: In this specific paragraph I'm trying to say "a patch can write to an XR or MR". I think it's important to be explicit that it's writing to an MR and if you run `kubectl describe rds.instance` or whatever you'll see that patch is applied.

content/master/concepts/patch-and-transform.md
plumbis: This section has been rewritten using To/FromCompositeFieldPath patches.

content/master/concepts/environment-configs.md
```diff
@@ -0,0 +1,291 @@
+---
+title: Environment Configurations
+weight: 90
+state: alpha
+alphaVersion: "1.11"
+description: "Environment Configurations or EnvironmentConfigs are an in-memory datastore used in patching Compositions"
+---
+
+A Crossplane EnvironmentConfig is an in-memory data store. Composition 
+patches can read from and write to a environment.
```
jbw976, line 10:
```suggestion
patches can read from and write to an environment.
```

content/master/concepts/environment-configs.md
```diff
@@ -0,0 +1,291 @@
...
+The Composition sorts all the results by the name of the environments and
+only uses the first environment in the sorted list. 
+
+Set the {{<hover label="selectResults" line="10">}}mode{{</hover>}} as 
+{{<hover label="selectResults" line="10">}}mode: Multiple{{</hover>}} to return
+all matched environments. Use 
+{{<hover label="selectResults" line="19">}}mode: Single{{</hover>}} to 
+return a single environment.
+
+{{<hint "note" >}}
```
jbw976, line 213: this hint is realllly helpful to explain the complexity of selecting and dealing with multiple results

content/master/concepts/environment-configs.md
```diff
@@ -0,0 +1,291 @@
...
+        sortByFieldPath: metadata.annotations[sort.by/weight]
+        matchLabels:
+          - key: my-label-key
+            type: Value
+            value: my-label-value
+          - key: my-label-key
+            type: FromCompositeFieldPath
+            valueFromFieldPath: spec.parameters.deploy
+```
+
```
jbw976, line 287: Do you think this would be a good place to reinforce for the reader again that now that multiple environments have been selected here, they will now all be merged with the top level list of environments?

jbw976 (commented):
Finished reading/reviewing the Environment Configurations page! Not too many comments here, mostly it was all helpful/useful 🤓

content/master/concepts/compositions.md
plumbis: This section has been updated and a new KB article added to clear up secret stuff. 
[connection details](https://deploy-preview-475--crossplane.netlify.app/knowledge-base/guides/connection-details/)

content/master/concepts/composite-resource-definitions.md
plumbis: KB added- 
https://deploy-preview-475--crossplane.netlify.app/knowledge-base/guides/connection-details/

jbw976 (commented):
Read through the new connection details KB page and the storing connection details section of the compositions page.

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,571 @@
+---
+title: Understanding Connection Details
+weight: 11
+description: "How to create and manage connection details across Crossplane managed resources, composite resources, Compositions and Claims"
+---
+
+Using connection details in Crossplane requires the following components:
+* Defining the `writeConnectionSecretToRef.name` in a [Claim]({{<ref "/master/concepts/claims#claim-connection-secrets">}}).
+* Defining the `writeConnectionSecretToRef` value in the [Composition]({{<ref "/master/concepts/compositions#save-connection-details-to-a-namespace">}}).
```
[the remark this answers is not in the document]
plumbis: Updated the headings in the Composition doc and linked to the specific sections covering these topics.

content/master/concepts/patch-and-transform.md
```diff
@@ -0,0 +1,1648 @@
...
+<!-- vale Google.Headings = YES -->
+
+The {{<hover label="multiply" line="8">}}type: multiply{{</hover>}} multiplies
+the input by the {{<hover label="multiply" line="9">}}multiply{{</hover>}} 
+value.
+
+For example, this 
+{{<hover label="multiply" line="8">}}type: multiply{{</hover>}} multiplies the
+value from the {{<hover label="multiply" line="3">}}fromFieldPath{{</hover>}}
+value by {{<hover label="multiply" line="9">}}2{{</hover>}}
```
AaronME, line 1449: Does multiply support floats? If so, we should call out that the lack of a "Divide" function is covered by multiplying decimals.

AaronME (commented):
Thanks, @plumbis !

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+with an external secret store. 
+{{</hint >}}
+
+## Background
+When a [Provider]({{<ref "/master/concepts/providers">}}) creates a managed
+resource, the resource may generate resource-specific details. These details can include 
+usernames, passwords or connection details like an IP address.  
+
+Crossplane refers to this information as the _connection details_ or 
+_connection secrets_.
```
sttts, line 31: These two paragraphs should better be right at the top to set the stage what this topic is actually about.

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+* Define the list of secret keys produced by each composed resource with `connectionDetails` in the
+  [Composition]({{<ref "/master/concepts/compositions#define-secret-keys">}}).
+* Optionally, define the `connectionSecretKeys` in a 
+  [CompositeResourceDefinintion]({{<ref "/master/concepts/composite-resource-definitions#manage-connection-secrets">}}).
+
+{{<hint "note">}}
+This guide discusses creating Kubernetes secrets.  
+Crossplane also supports using external secret stores like [HashiCorp Vault](https://www.vaultproject.io/). 
+
+Read the [external secrets store guide]({{<ref "/knowledge-base/integrations/vault-as-secret-store">}}) for more information on using Crossplane
```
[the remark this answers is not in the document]
plumbis: I haven't tested it but according to [code comments](https://github.com/crossplane/crossplane-runtime/blob/8e87b2c47ee705f88b1224bff7b3eb8d1a071760/apis/common/v1/resource.go#L180-L182) you can do both.

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+Data
+====
+attribute.secret:                40 bytes
+attribute.ses_smtp_password_v4:  44 bytes
+password:                        40 bytes
+username:                        20 bytes
+```
+
+Compositions and CompositeResourceDefinitions require the exact names of the
+secrets generated by a resource.
```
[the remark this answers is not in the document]
plumbis: I'm struggling to express "hey, this is how you find the names of the keys a resource produces and you'll need them when defining the `connectiondetails` in a Composition" 

I'm wide open to suggestions.

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+username:                        20 bytes
+```
+
+Compositions and CompositeResourceDefinitions require the exact names of the
+secrets generated by a resource. 
+
+## Connection secrets in Compositions
+
+Resources in a Composition that create connection details still create a
+secret object containing their connection details.
```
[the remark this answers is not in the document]
plumbis: Are you trying to say "an MR creates a secret. This always happens, even when it's an XR?"

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+secrets generated by a resource. 
+
+## Connection secrets in Compositions
+
+Resources in a Composition that create connection details still create a
+secret object containing their connection details.  
+Crossplane also generates
+another secret object for each composite resource, 
+containing the secrets from all the defined resources.
+
```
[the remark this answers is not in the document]
plumbis: I don't think that's necessarily true. it's possible that the composed resource secret isn't used/needed and only the composite secret is used.

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+Namespace:    other-namespace
+
+Type:  connection.crossplane.io/v1alpha1
+
+Data
+====
+username:                        20 bytes
+attribute.secret:                40 bytes
+attribute.ses_smtp_password_v4:  44 bytes
+password:                        40 bytes
```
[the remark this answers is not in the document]
plumbis: Ask kubernetes 😆 
that's the way they were in the output

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+{{<hover label="comp2" line="16">}}connectionDetails{{</hover>}} object under
+each resource and define the secret keys the resource creates.  
+
+
+{{<hint "warning">}}
+You can't change the 
+{{<hover label="comp2" line="16">}}connectionDetails{{</hover>}} 
+of a Composition.  
+You must delete and
+recreate the Composition to change the
```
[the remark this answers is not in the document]
plumbis: If you remove the resource creating the secret then no secret is created. 

If you remove a different resource then there's no change.

I don't know the _why_ but the defined connectiondetails applied when the Composition controller(? or whatever it is) is created are immutable. Changing the defined connectiondetails won't be reflected on any new composed resources. They'll have that originally defined set of secrets.

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+```
+
+{{<hint "important">}}
+If a key isn't listed in the 
+{{<hover label="comp2" line="16">}}connectionDetails{{</hover>}}
+it isn't stored in the secret object.
+{{< /hint >}}
+
+### Managing conflicting secret keys 
+If resources produce conflicting keys, create a unique name with a connection
```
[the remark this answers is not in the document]
plumbis: Last to write wins. I thought about putting that in there but figured it was easier to just say "don't do that"

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+# Removed for brevity.
+```
+
+## Connection secrets in Composite Resource Definitions
+
+The CompositeResourceDefinintion (`XRD`), can restrict which secrets keys are 
+put in the combined secret and provided to a Claim. 
+
+By default an XRD writes all secret keys from the composed resources to the 
+combined secret object.
```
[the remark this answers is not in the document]
plumbis: Edited:

>By default an XRD writes all secret keys listed in the composed resource `connectionDetails` to the combined secret object.

content/knowledge-base/guides/connection-details.md
```diff
@@ -0,0 +1,593 @@
...
+
+Inside the {{<hover label="xrd" line="4">}}connectionSecretKeys{{</hover>}} list
+the secret key names to create. Crossplane only adds the keys listed to the
+combined secret.
+
+{{<hint "warning">}}
+You can't change the 
+{{<hover label="xrd" line="4">}}connectionSecretKeys{{</hover>}} of an XRD. 
+You must delete and
+recreate the XRD to change the
```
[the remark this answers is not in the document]
plumbis: Yep, applies to deletes as well.

content/master/concepts/claims.md
```diff
@@ -0,0 +1,204 @@
...
+Kubernetes object, inside a namespace. 
+
+Crossplane creates Claims when users access a
+custom API, defined in the CompositeResourceDefinition. 
+
+{{< hint "tip" >}}
+
+Claims are like [composite resources]({{<ref "./composite-resources">}}). The
+difference between Claims and composite resources is Crossplane can create 
+Claims in a namespace, while composite resources are cluster scoped.
```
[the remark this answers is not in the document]
plumbis: Users could create XRs too. This wording makes it sound like you can't create XRs directly.

content/master/concepts/claims.md
```diff
@@ -0,0 +1,204 @@
...
+{{<hover label="claimSec" line="7">}}name: my-claim-secret{{</hover>}}.
+```yaml {label="claimSec"}
+apiVersion: example.org/v1alpha1
+kind: database
+metadata:
+  name: my-claimed-database
+spec:
+  writeConnectionSecretToRef:
+    name: my-claim-secret
+```
```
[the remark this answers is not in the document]
plumbis: This is called out in a note a few lines above.

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,817 @@
...
+kind: CompositeResourceDefinition
+metadata: 
+  name: xmydatabases.example.org
+spec:
+  group: example.org
+  names:
+    kind: xMyDatabase
+    plural: xmydatabases
+  versions:
+  - name: v1alpha1
```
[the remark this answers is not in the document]
plumbis: I've thought about it and haven't come up with a good answer that isn't Clippy-esque. I'm open to suggestions

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,817 @@
...
+  claimNames:
+    kind: Database
+    plural: databases
+  versions:
+  # Removed for brevity
+```
+
+{{<hint "important" >}}
+You can't change the `claimNames` after they're defined. You must delete and
+recreate the XRD to change the `claimNames`.
```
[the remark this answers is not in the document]
plumbis: 👍 added a warning earlier in the doc that those fields are also immutable.

content/master/concepts/composite-resource-definitions.md
```diff
@@ -0,0 +1,817 @@
...
+
+{{<hint "important" >}}
+The keys listed in the
+{{<hover label="key" line="10">}}connectionSecretKeys{{</hover>}} must match the 
+key names listed in the Composition's `connectionDetails`.  
+
+An XRD ignores any keys listed that aren't created by a managed resource.
+
+For more information read the 
+[Composition documentation]({{<ref "./composite-resource-definitions#manage-connection-secrets">}}).
```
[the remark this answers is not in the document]
plumbis: I definitely didn't mean to link back to this same section. Copy/Pasta fail.

This specific link I want to go to the Compositions docs to provide the direct reference. At the end of this section we link to the KB guide for a more complete example. 

The duplication of content is on purpose to help make it more obvious and discoverable.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,1204 @@
...
+supports any settings used in a managed resource including
+applying `annotations` and `labels` or using a specific 
+`providerConfigRef`.
+
+{{<hint "note" >}}
+Compositions allow applying a `metadata.name` to a resource's 
+{{<hover label="resources" line="16">}}spec{{</hover>}} but ignores it. The
+`metadata.name` field doesn't influence the name
+of the managed resource in Crossplane or the external resource inside the 
+Provider.
```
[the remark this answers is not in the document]
plumbis: Yes. when _creating_ a composed resource `metadata.name` is ignored and Crossplane assigns "<claim_name>-<hash>". When the patch is applied the composite resource now exists with a metadata.name that can be patched.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,1204 @@
...
+                type: internet
+# Removed for brevity
+```
+### Enabling Composite Resources
+
+A Composition is only a template defining how to create managed 
+resources. A Composition limits which Composite Resources can use this
+template. 
+
+A Composition's {{<hover label="typeref" line="6">}}compositeTypeRef{{</hover>}}
```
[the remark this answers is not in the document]
plumbis: The plural is referring to multiple XRs (of the same type) using the template. You're saying a single XRD mapping.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,1204 @@
...
+| {{<hover label="conDeet" line="16">}}fromConnectionSecretKey{{</hover>}} | Create a secret key matching the key of a secret generated by the resource. | 
+| {{<hover label="conDeet" line="18">}}fromFieldPath{{</hover>}}  | Create a secret key matching a field path of the resource. |
+| {{<hover label="conDeet" line="20">}}value{{</hover>}}  | Create a secret key with a pre-defined value. |
+{{< /table >}}
+
+{{<hint "note">}}
+The {{<hover label="conDeet" line="20">}}value{{</hover>}} type must use a
+string value.  
+
+The {{<hover label="conDeet" line="20">}}value{{</hover>}} isn't added to the
```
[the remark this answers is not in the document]
plumbis: You mean about using `value` or more generally in the section?

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,1204 @@
...
+====
+myUsername:      20 bytes
+myFieldSecret:   24 bytes
+myStaticSecret:  18 bytes
+```
+
+{{<hint "note" >}}
+The CompositeResourceDefinition can also limit which keys Crossplane stores from
+the composite resources.  
+By default CompositeResourceDefinitions store all resource
```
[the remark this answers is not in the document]
plumbis: Fixed the wording here and in the KB. 

> By default an XRD writes all secret keys listed in the composed resources `connectionDetails` to the combined secret object.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,1204 @@
...
+Stores by default.
+{{< /hint >}}
+
+Use 
+{{<hover label="gcp-storeconfig"
+line="11">}}publishConnectionDetailsWithStoreConfigRef{{</hover>}}
+in place of 
+`writeConnectionSecretsToNamespace` to define the 
+{{<hover label="gcp-storeconfig" line="2">}}StoreConfig{{</hover>}} 
+to save connection details to.
```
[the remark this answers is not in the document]
plumbis: Mentioned earlier, but is that true? Code comments in crossplane-runtime imply you can use both.

content/master/concepts/environment-configs.md
```diff
@@ -0,0 +1,295 @@
...
+        matchLabels:
+          - key: my-label-key
+            type: Value
+            value: my-label-value
+          - key: my-label-key
+            type: FromCompositeFieldPath
+            valueFromFieldPath: spec.parameters.deploy
+  resources:
+  # Removed for brevity
+```
```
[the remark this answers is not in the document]
plumbis: No rule here but I thought it'd be easier than creating two very small sections. I'm willing to change it if you think that'd be better.

content/master/concepts/patch-and-transform.md
plumbis: Good point, I didn't think about that. It only supports integers. Note added.

content/knowledge-base/guides/connection-details.md
plumbis: The reason it's not is because the current top is intended to be a quick reference for anyone dealing with secrets and wanting a checklist to see what they did right/wrong.

content/master/concepts/compositions.md
```diff
@@ -0,0 +1,1204 @@
...
+
+Use the `crossplane.io/external-name` annotation on the resource to influence
+the external resource name.
+{{< /hint >}}
+
+#### Template a ProviderConfig
+
+Compositions can define a ProviderConfig like it defines managed resources.
+Generating a ProviderConfig may be useful in providing unique credentials to
+each deployment.
```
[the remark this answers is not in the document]
plumbis: What would be a resource that doesn't honor the spec?

jbw976 (approved):
Thanks for all the collaboration and syncing together to review and discuss all these changes. They will be a great improvement for the v1.13 docs!! 💪

content/master/concepts/compositions.md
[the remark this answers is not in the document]
AaronME: You can't put a Pod or Configmap directly into a composition. You have to wrap them in provider-kubernetes resources.

content/master/concepts/compositions.md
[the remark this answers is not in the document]
AaronME: I can run a few more tests. When I tried to mix and match, Crossplane  "lost" secrets.

content/master/concepts/compositions.md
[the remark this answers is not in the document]
plumbis: Good points. 
In this case I'd prefer to stick with "the docs say what _is_ supported, not what _isnt_" without introducing the complexity of needing to know what the XRM is.

PR #485, opened 2023-07-15
Clarify the specifics of connection details in compositions

jbw976:
This PR attempts to clarify the nuances/specifics that are necessary to correctly specify the connection details that will be aggregated from composed resources and published for a claim.

This wording was put together in our latest FAQ blog post, so I wanted to bring the docs closer in line to what we felt worked in that post: https://blog.crossplane.io/faq-2-claim-connection-details/

Most of the changes are just rewordings and clarifications, e.g. to point to specific field paths, but there is 1 line I found in my testing that appears to be incorrect in our current docs:

1. The claim and XR must both specify a `writeConnectionSecretToRef`.

I believe this is inaccurate because XR/XRDs do not have a `writeConnectionSecretToRef` field.  Instead, a Composition has a `writeConnectionSecretsToNamespace` field and that was missing from our current docs.

### DIFF
--- a/content/master/concepts/composition.md
+++ b/content/master/concepts/composition.md
@@ -1270,12 +1270,20 @@ Claim and Composite Resource connection secrets are often derived from the
 connection secrets of the managed resources they compose. This is a common
 source of confusion because several things need to align for it to work:
 
-1. The XR/claim's connection secret keys must be declared by the XRD.
-1. The `Composition` must specify how to derive connection details from each
-   composed resource.
-1. If connection details are derived from a composed resource's connection
-   secret that composed resource must specify its `writeConnectionSecretToRef`.
-1. The claim and XR must both specify a `writeConnectionSecretToRef`.
+1. The **claim** must specify the secret where the aggregated connection details
+   should be written
+    * This is the `spec.writeConnectionSecretToRef` field in a claim
+1. The **composite resource** must define which connection details to aggregate
+   from its children and publish to the claim
+    * This is the `spec.connectionSecretKeys` field in a composite resource
+1. The **composition** must define where to write its aggregated connection
+   details
+    * This is the `spec.writeConnectionSecretsToNamespace` field in the
+       composition
+1. Each child **composed resource** must define the connection details it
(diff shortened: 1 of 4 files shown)


plumbis:
Can you take a look at the changes in PR #475 and give feedback on this there?

[XRD Secrets](https://64ac3c152a16d9000817d0ca--crossplane.netlify.app/master/concepts/composite-resource-definitions/#manage-connection-secrets)

[XR Secrets](https://64ac3c152a16d9000817d0ca--crossplane.netlify.app/master/concepts/composite-resources/#manage-connection-secrets)

[Composition Secrets](https://64ac3c152a16d9000817d0ca--crossplane.netlify.app/master/concepts/compositions/#storing-connection-details)

Looks like I may have missed covering secrets for Claims.

jbw976:
Just took a pass now @plumbis: https://github.com/crossplane/docs/pull/475#pullrequestreview-1533918330

jbw976:
> there is 1 line I found in my testing that appears to be incorrect in our current docs:
>   1. The claim and XR must both specify a writeConnectionSecretToRef.
> 
> I believe this is inaccurate because XR/XRDs do not have a writeConnectionSecretToRef field

btw, I think I goofed this 😓 - I think I was reading the original documentation from the perspective of the XRD having to specify a `writeConnectionSecretToRef` field.  What I now believe I overlooked is the scenario of someone directly creating a XR instead of indirectly having the XR created via a claim.

I'll try to update both this PR and the [blog post](https://blog.crossplane.io/faq-2-claim-connection-details/) to get this detail correct 😩

jbw976:
@plumbis I updated the guidance in these docs to be actually accurate in https://github.com/crossplane/docs/pull/485/commits/ec97ef21b7e88089d46aac30a5d5eac372472faf, can you take another look?

I also published that update to https://blog.crossplane.io/faq-2-claim-connection-details/ already so the correction is out there for folks. We can tweak if needed from feedback in this PR 😅

plumbis (approved):

PR #491, opened 2023-07-20
fix(docs): fix poll-interval -

haarchri:
remove a leading - for `poll-interval`


jeanduplessis (approved):

jeanduplessis:
Thanks @haarchri for the fix.

PR #500, opened 2023-07-27
Improve Vale spellcheck

### LINKED ISSUE #490: Vale doesn't catch CompositeResourceDefinintions as an error
### What's Wrong?
Noticed in #475, Vale doesn't flag the string "CompositeResourceDefinintions" as a spelling error. (should be "CompositeResourceDefinitions")

plumbis:
In #475 it was noticed that typos of `CompositeResourceDefinition` aren't flagged by Vale.

As described in [Vale #527](https://github.com/errata-ai/vale/issues/527) the default spellcheck ignores words with multiple caps in the word, assuming it's covered by something else.

We don't have an easy way to check for _any_ misspelling of a `kind` like CompositeResourceDefinition.

**Good News:**
This PR fixes it and makes the spell checker more sensitive and it catches any error like this (and some others we had previously missed)

**Bad News**
It makes the spell checker more sensitive. It not picks up errors with links in line breaks and a few other conditions described in the Contributor guide updates that are part of this PR. Going forward I expect a  small uptick in the work we have to do in order to keep the spelling exceptions updated.

Resolves #490 


jeanduplessis (approved):
LGTM

PR #501, opened 2023-07-27
Update tags for new GA version

plumbis:
updates the google analytics tag for the new google analytics 4 requirements.

### DIFF
--- a/themes/geekboot/layouts/partials/google-analytics.html
+++ b/themes/geekboot/layouts/partials/google-analytics.html
@@ -1,9 +1,7 @@
-<!-- Global site tag (gtag.js) - Google Analytics -->
-<script async src="https://www.googletagmanager.com/gtag/js?id=G-SFCPQYSLHY"></script>
-<script>
-  window.dataLayer = window.dataLayer || [];
-  function gtag(){dataLayer.push(arguments);}
-  gtag('js', new Date());
+<!-- Google Tag Manager -->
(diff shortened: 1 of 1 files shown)


themes/geekboot/layouts/partials/google-analytics.html
```diff
@@ -1,9 +1,7 @@
...
-  gtag('js', new Date());
+<!-- Google Tag Manager -->
+<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-WFF2NQHG');</script>
+<!-- End Google Tag Manager -->
 
-  gtag('config', 'G-SFCPQYSLHY');
-</script>
\ No newline at end of file
+<!-- Google Tag Manager (noscript) -->
+<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-WFF2NQHG" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
```
jbw976, line 6: i think this 2nd section is supposed to go in `<body>` - will this partial layout do that? or will this get incorrectly included into the `<head>` still instead?

plumbis: We do things weird-

This file is loaded as part of the [footer](https://github.com/crossplane/docs/blob/2583a00224bedc530300e9096fec9431285919eb/themes/geekboot/layouts/partials/scripts.html#L7) at the [bottom of the body](https://github.com/crossplane/docs/blob/master/themes/geekboot/layouts/_default/baseof.html#L16) of the page.

The reason for this is to try and load the page a few ms faster without having the non-essential stuff in the way. We may lose a few hits on folks that bail on the page before load, but I'm okay with that if it means faster load times for everyone.

jbw976: oh okay well then if that's the way it's done on this site and we expect this to work correctly still then all good 😊 👍

jbw976 (approved):

Issue #502, opened 2023-07-27
More elaborate description of composition function's desired resources


tenstad:
### What's Missing?

Location: https://github.com/crossplane/docs/blob/40c435e85062a8d3d7bb3d8854b07bee730de368/content/master/concepts/composition-functions.md?plain=1#L348-354

Would be nice to explicitly state that the initial desired state passed to the function array does NOT include any resources created by composition functions. These resources are not copied from the observed list. Could also include that later functions are passed the desired resources of their previous functions.

I have formulated a suggestion based on my current understanding. Quite new to this, please correct me if I'm wrong.

```diff
- * A copy of the observed state of the XR.
+ * A copy of the observed state of the composite resource (XR).
- * A copy of the observed state of any existing composed resources.
+ * A copy of the observed state of any existing composed resources produced
+   from the `resources` array.
  * Any new composed resources or modifications to observed resources produced
    from the `resources` array.
+ 
+ The initial desired state does NOT include any copies of observed resources
+ produced by the function pipeline. When using multiple functions, each function's
+ desired resources output is passed as input to the next pipeline function.
```

plumbis:
This sounds good to me. My only suggestion would be to put the added paragraph in a [hint box](https://docs.crossplane.io/contribute/features/#hints-and-alert-boxes).

Also a minor edit to pass [Vale linting](https://docs.crossplane.io/contribute/vale/).

> The initial desired state doesn't include any copies of observed resources produced by the function pipeline.  
> When using multiple functions each function passes their desired resources output as input to the next
pipeline function.

Do you want to make the contribution and send a PR?

PR #506, opened 2023-07-28
docs:  clarify content of desired resources in composition functions

tenstad:
Resolves https://github.com/crossplane/docs/issues/502

Clarifies content of `desired.resources` and when resources produced by function pipeline may be present. It is now less likely that someone thinks resources produced by a function pipeline is also copied from `observed` to `desired` during _bootstrapping_ of initial desired state.

- I think it is good actually use the word _composite_, as it refers to the `composite` field within `desired`
- Clarifies that _existing composed resources_  are produced from the `resources` array (same origin as _new_ and _modified_)
- Explicitly states that initial desired state does not contain resources produced by the function pipeline
- And that the desired resources output of a function is passed as input to the next pipeline function


plumbis (approved):
Thanks for the contribution!!

PR #508, opened 2023-07-29
Revamp Azure quickstart to focus on creating custom API

plumbis:
This updates the Azure Quickstart guide in master and v1.13 to use the Upbound family providers and changes the focus from being about Crossplane components to focusing on building and using a custom API with Crossplane.

The main motivation for the change is the new /concepts topics pushed in #475 gives us much better places to point people to discuss the details of the Crossplane components. We can have the quickstart focus on outcomes instead.

This also deletes part 3 of the quickstart that covered patching and packages. Patching is now covered in part 2 and packaging isn't required for a "quickstart".


plumbis:
Hugo is failing because this requires the "install crossplane" include in #507

PR #509, opened 2023-07-30
fix/aws-getting-started

r0B-O:
Fixes issues in AWS Getting Started Guide

- use provider-aws v0.41.1 instead of 0.27.0
- replace `aws.upbound.io/v1beta1` with `pkg.crossplane.io/v1` in provider config
- replace `s3.aws.upbound.io/v1beta1` with `s3.aws.upbound.io/v1beta1`

The current doc installs provider which does not macth the API in examples.


r0B-O:
closigng to riase another with proper signoff

Issue #497, opened 2023-07-24
update "Import resources automatically" after `Support Granular management policies`


haarchri:
### What's Missing?

after we merging the following PRs we need to update the documentation 

The old
`spec.managmentPolicy: FullControl/ObserveOnly/OrphanOnDelete`
is getting replaced by
`spec.managementPolicies:["*", "Observe", "Create", "Update", "LateInitialize", "Delete"]`

https://github.com/upbound/upjet/pull/224
https://github.com/upbound/provider-aws/pull/785
https://github.com/upbound/provider-gcp/pull/336
https://github.com/upbound/provider-azure/pull/495

haarchri:
I'm contemplating whether it's necessary to clarify whether we should drop the old design document or not...

https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md
vs.
https://github.com/crossplane/crossplane/blob/master/design/one-pager-ignore-changes.md#deletion-policy

lsviben:
> I'm contemplating whether it's necessary to clarify whether we should drop the old design document or not...
> 
> https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md vs. https://github.com/crossplane/crossplane/blob/master/design/one-pager-ignore-changes.md#deletion-policy

Can we use the Status field? ObserveOnly -> Deprecated, IgnoreChanges -> Approved/Implemented/Accepted?

plumbis:
@haarchri are there changes to Crossplane to support these new statuses? Is this something entirely in the scope of providers?

PR #510, opened 2023-07-31
update management policies with GMP changes

lsviben:
Late update for the management policies alpha feature. 

Replaces the old `managementPolicy` to `managementPolicies`, updates the Importing guide and adds a section for `spec.initProvider`. 

Fixes: #497

### DIFF
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -61,8 +61,9 @@ Provider deletes the managed resource but doesn't delete the external resource.
 #### Interaction with management policies
 
 If a resource configures a Crossplane
-[management policy](#managementpolicies), the management policy takes precedence 
-over the `deletionPolicy` setting, unless it's the default management policy.
+[management policy](#managementpolicies) and the related management policy alpha
+feature is enabled, the management policy takes precedence over the
+`deletionPolicy` setting, unless it's the default management policy.
 
 {{< table "table table-sm table-hover">}}
 | managementPolicies          | deletionPolicy   | result  |
@@ -113,13 +114,6 @@ inside a Provider's web console, Crossplane reverts that change back to what's
 configured in the `forProvider` setting. 
 {{< /hint >}}
 
-#### Late initialization
-
-After the external resource creation, providers add some provider defaulted
-settings not manually set to the `forProvider` field of the created managed
-resource object into their respected `forProvider` fields.
-Use `kubectl describe <managed_resource>` to view the applied values. 
- 
 #### Referencing other resources
 
 Some fields in a managed resource may depend on values from other managed
@@ -229,7 +223,7 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
 <!-- vale write-good.Passive = YES -->
 {{< /hint >}}
 
-#### Late Initialization
+#### Late initialization
 
 For some of the optional fields, users rely on the default that the cloud
 provider chooses for them. Since Crossplane treats the managed resource as the
@@ -309,10 +303,10 @@ spec:
 <!-- vale on --> 
 
 {{<hint "important" >}}
-The managed resource `managementPolicies` option is an alpha feature. 
+The managed resource `managementPolicies` option is an alpha feature.
 
-Enable `managementPolicies` in a provider with `--enable-management-policies` 
-in a 
+Enable `managementPolicies` in a provider with `--enable-management-policies`
+in a
 [ControllerConfig]({{<ref "./providers#controller-configuration" >}}).
 {{< /hint >}}
 
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -208,7 +208,19 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
 `delete`.
 <!-- vale write-good.Passive = YES -->
 {{< /hint >}}
- 
+
+#### Late Initialization
+
+For some of the optional fields, users rely on the default that the cloud
+provider chooses for them. Since Crossplane treats the managed resource as the
+source of the truth, values of those fields need to exist in `spec` of the
+managed resource. So, in each reconciliation, Crossplane will fill the value of
+a field that is left empty by the user but is assigned a value by the provider.
+For example, there could be two fields like `region` and `availabilityZone` and
+you might want to give only `region` and leave the availability zone to be
+chosen by the cloud provider. In that case, if the provider assigns an
+availability zone, Crossplane gets that value and fills `availabilityZone`. Note
+that if the field is already filled, the controller won't override its value.
 
 <!-- vale off -->
 ### managementPolicies
@@ -222,33 +234,35 @@ in a
 [ControllerConfig]({{<ref "./providers#controller-configuration" >}}).
 {{< /hint >}}
 
-Granular `managementPolicies` determine what actions Crossplane can take on a 
+Granular `managementPolicies` determine what actions Crossplane can take on a
 managed and its external resource. `managementPolicies` is an array of actions
 (`Observe`, `Create`, `Update`, `Delete`, `LateInitialize`, `*`), and there are
-multiple combinations that Crossplane supports based on the use case. For 
-example, setting `["Observe"]` makes Crossplane treat the resource as an
-`ObserveOnly` resource, importing the external resources not originally created
-by Crossplane. This allows new managed resources to reference the `ObserveOnly`
-resource, for example, a shared database or network. 
-The `["Observe"]` policy can also place existing resources under the control of
-Crossplane.  
+multiple combinations that Crossplane supports based on the use case.
+
+For example, setting just `["Observe"]` makes Crossplane treat the resource as
+a read-only resource, importing the external resources not originally created
+by Crossplane. This allows other managed resources to reference the read-only
+resource, for example, a shared database or network. Importing a resource as
+read-only is convenient since Crossplane will not attempt to take control and
+only identifier parameters are needed to be specified in the managed resource.
 
 {{< hint "tip" >}}
-Read the [Import Existing Resources]({{<ref "/knowledge-base/guides/import-existing-resources" >}}) 
+Read the [Import Existing Resources]({{<ref "/knowledge-base/guides/import-existing-resources" >}})
 guide for more
 information on using the `managementPolicies` to import existing resources.
 {{< /hint >}}
 
-Furthermore, granular `managementPolicies` can be used to skip the late 
-initialization of the managed resource, by not including `LateInitialize` in the
-`managementPolicies`, which is useful when there are conflicting fields between
-the managed resource and the external resource, due to some external resource
-field being controlled externally. 
+Another application of `managementPolicies` is to skip the [Late Initialization]({{<ref "./managed-resources#late-initialization" >}})
+of the managed resource, by not including `LateInitialize` in the
+`managementPolicies`, which is useful when there are fields in the resource,
+that we would not like Crossplane to control, but would usually be filled in
+by Late Initialization.
 
 {{< hint "tip" >}}
 Read the [initProvider]({{<ref "./managed-resources#initprovider" >}}) section
-for more info about how we can use `LateInitialize` and `initProvider` together
-to ignore changes to fields that are controlled externally.
+for more info about how we can use management policies without `LateInitialize`
+and `initProvider` together to ignore changes to fields that are controlled
+externally.
 {{< /hint >}}
 
 The `"Delete"` action replaces the [deletionPolicy]({{<ref "./managed-resources#deletionpolicy" >}})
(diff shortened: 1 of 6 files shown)


lsviben:
Vale seems to complain for `managementPolices`, `deletionPolicy` . Can we add this somewhere for Vale to ignore?

Also btw, I would like to add a guide for using `initProvider` to ignore changes, but as a separate PR.

plumbis:
Re: Vale -
Yes, you can add them to the [crossplane-words.txt](https://github.com/crossplane/docs/blob/master/utils/vale/styles/Crossplane/crossplane-words.txt) file. 
https://docs.crossplane.io/contribute/vale/#spelling-errors-and-exceptions

I have some suggested edits for you but before I make them I want to make sure I understand the feature- 

These changes came in 1.13? 

For `LateInitialize` the resource will set anything defined in `forProvider` and import all other settings? It's like half an observe only resource? 

Are there any restrictions on what can be combined as a policy (even if it seem to make sense)? Is [this chunk from crossplane-runtime](https://github.com/crossplane/crossplane-runtime/blob/0d8cbcef6fac5299f2ce724f3bed0dec2f72f444/pkg/reconciler/managed/policies.go#L46) the supported set of combinations? 

> It's suggested to use a combination of `managementPolicies` without  `LateInitialize` with `initProvider` to avoid late initialization of fields  in `forProvider`.

I'm not following what this is telling me to do/why to do it.

content/master/concepts/managed-resources.md
```diff
@@ -211,37 +211,111 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
-This allows new managed resources to reference 
-the `ObserveOnly` resource, for example, a shared database or network. 
-The `ObserveOnly` policy can also place existing resources under the control of
+Granular `managementPolicies` determine what actions Crossplane can take on a 
+managed and its external resource. `managementPolicies` is an array of actions
+(`Observe`, `Create`, `Update`, `Delete`, `LateInitialize`, `*`), and there are
+multiple combinations that Crossplane supports based on the use case. For 
+example, setting `["Observe"]` makes Crossplane treat the resource as an
+`ObserveOnly` resource, importing the external resources not originally created
+by Crossplane. This allows new managed resources to reference the `ObserveOnly`
```
turkenh, line 231: > importing the external resources not originally created by Crossplane.

Import is not limited to `["Observe"]`. One can still import using `["*"]` as long as they provide all required parameters. Observe only just makes importing more convenient (since you don't have to provide all required params) and safer (since provider doesn't immediately take control).

content/master/concepts/managed-resources.md
```diff
@@ -211,37 +211,111 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
+resource, for example, a shared database or network. 
+The `["Observe"]` policy can also place existing resources under the control of
```
turkenh, line 233: Not sure I got what this sentence means 🤔

content/master/concepts/managed-resources.md
```diff
@@ -211,37 +211,111 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
 
 {{< hint "tip" >}}
 Read the [Import Existing Resources]({{<ref "/knowledge-base/guides/import-existing-resources" >}}) 
 guide for more
-information on using the `managementPolicy` to import existing resources.
+information on using the `managementPolicies` to import existing resources.
 {{< /hint >}}
 
+Furthermore, granular `managementPolicies` can be used to skip the late 
+initialization of the managed resource, by not including `LateInitialize` in the
```
turkenh, line 243: A link to a section about Late initialization would be helpful. 
I remember there was such a section under managed resources page but couldn't find it.

content/master/concepts/managed-resources.md
```diff
@@ -211,37 +211,111 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
+to ignore changes to fields that are controlled externally.
+{{< /hint >}}
+
+The `"Delete"` action replaces the [deletionPolicy]({{<ref "./managed-resources#deletionpolicy" >}})
+field in the managed resource. For now the `deletionPolicy` is still supported,
+but only if it's set to a non default value, which is `Orphan`. If it's set to
+`Orphan`, and the `managementPolicies` is set to `["*"]`, which is default,
+then the external resource will be orphaned when the managed resource is
+deleted. In other cases, non default `managementPolicies` take precedence over
+the `deletionPolicy` field.
```
turkenh, line 260: But only when the feature is enabled, right?
It may be hard to follow, which is especially important since it is controlling a critical behavior. Wondering if a table showing the behavior of different combinations would make it clearer.

content/master/concepts/managed-resources.md
```diff
@@ -211,37 +211,111 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
+
 #### Options
-* `managementPolicy: FullControl` - **Default** - Crossplane can create, change
-  and delete the managed resource. 
-* `managementPolicy: ObserveOnly` - Crossplane only imports the details of the
-  external resource, but doesn't make any changes to the managed resource. 
+* `*` - ** Default ** - Crossplane can observe, create, change, delete the 
+   external resource and update the managed resource object with late init
```
turkenh, line 264:
```suggestion
   external resource and update the managed resource spec with late init
```

content/master/concepts/managed-resources.md
```diff
@@ -211,37 +211,111 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
+   parameters.
+* `Observe` - managed resource `status.atProvider` will be updated with the 
+   external resource state.
+* `Create` - external resource will be created using the managed resource
+   `spec.initProvider` and `spec.forProvider`.
+* `Update` - external resource will be updated using the managed resource
+   `spec.forProvider`.
+* `Delete` - the external resource will be deleted when the managed resource is
+   deleted.
+* `LateInitialize` - select fields of the managed resource
```
turkenh, line 274: What about something like? 

> * `LateInitialize` - Unprovided spec fields are late-initialized to `spec.forProvider` with the default values from the cloud provider. This enables XP to take full control of the external resource, even those values not provided by the user upfront.

Again a link to a more detailed section would be helpful.

lsviben:
> Re: Vale - Yes, you can add them to the [crossplane-words.txt](https://github.com/crossplane/docs/blob/master/utils/vale/styles/Crossplane/crossplane-words.txt) file. https://docs.crossplane.io/contribute/vale/#spelling-errors-and-exceptions
> 

Thanks. Added them.

> I have some suggested edits for you but before I make them I want to make sure I understand the feature-
> 
> These changes came in 1.13?

Yeah, these changes are a part of the Granular managmenent policy / ingore changes that came in 1.13, but are actually implemented on a provider level, so they will be released with the next provider release ( for now just the Upbound official ones).

> 
> For `LateInitialize` the resource will set anything defined in `forProvider` and import all other settings? It's like half an observe only resource?

Late initialization has been around for a while, its not a part of this feature. But its related because this feature now allows us to skip Late Initialization by setting up a granular management policy which does not include it. A big part of the ignore changes feature requests came because Crossplane was late-initializing some fields that users wanted to be left alone. 

I saw that Late Initialize section was removed from the docs, probably unintentionally, so I found an old version from the repo and returned it back. Maybe it could do some rework, to avoid confusion on what it actually does. 

> 
> Are there any restrictions on what can be combined as a policy (even if it seem to make sense)? Is [this chunk from crossplane-runtime](https://github.com/crossplane/crossplane-runtime/blob/0d8cbcef6fac5299f2ce724f3bed0dec2f72f444/pkg/reconciler/managed/policies.go#L46) the supported set of combinations?

yeah we support by default a certain set of policies, which is subject to change if somebody has a good use case. The providers have an option to set their own set of polices to support if they prefer so.

> 
> > It's suggested to use a combination of `managementPolicies` without  `LateInitialize` with `initProvider` to avoid late initialization of fields  in `forProvider`.
> 
> I'm not following what this is telling me to do/why to do it.

Reworded, basically it means that in most cases you should use a mangement policy which does not have `LateInitialize` when using `initProviders` because the whole point of `initProviders` is to avoid setting the fields in `forProvider`. Which can be messed up by Crossplane late-initializing that field in `forProvider`

content/master/concepts/managed-resources.md
lsviben: added the table, and mention that its only when the feature is enabled.

content/master/concepts/managed-resources.md
lsviben: there was a section in the old docs, I found it and brought it back as it was. Maybe it could do with some refurbishment?

content/master/concepts/managed-resources.md
```diff
@@ -208,40 +208,142 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
+{{< /hint >}}
+
+The `spec.initProvider` is a subset of the `spec.forProvider` fields that
+Crossplane merges with `forProvider` to create the external resource, but
+ignores when updating it. This allows users to create a resource with all the
+required fields which may then be controlled externally. For example, a user may
+want to create a `NodeGroup` with the required `desiredSize` field, but then
+let an autoscaler control the field. If the `desiredSize` field is included
+in the`forProvider` field, Crossplane will attempt to update the field, which
+will conflict with the autoscaler.
```
plumbis, line 325: Suggested restructure to improve wording and flow and add `{{<hover>}}` support.

```suggestion
{{<hint "important" >}}
The managed resource `initProvider` option is an alpha feature related to 
[managementPolicies]({{<ref "./managed-resources#managementpolicies" >}}).

Enable the `initProvider` in a provider with `--enable-management-policies`
in a
[ControllerConfig]({{<ref "./providers#controller-configuration" >}}).
{{< /hint >}}

The `spec.initProvider` is a subset of the `spec.forProvider` fields that
Crossplane merges with `forProvider` to create the external resource, but
ignores when updating it. This allows users to create a resource with all the
required fields which may then be controlled externally. For example, a user may
want to create a `NodeGroup` with the required `desiredSize` field, but then
let an autoscaler control the field. If the `desiredSize` field is included
in the`forProvider` field, Crossplane will attempt to update the field, which
will conflict with the autoscaler.
```

```
The 
{{<hover label="initProvider" line="7">}}initProvider{{</hover>}} defines 
settings Crossplane applies only when creating a new managed resource.  
Crossplane ignores settings defined in the 
{{<hover label="initProvider" line="7">}}initProvider{{</hover>}} 
field that change after creation.

{{<hint "note" >}}
Settings in `forProvider` are always enforced by Crossplane. Crossplane reverts
any changes to a `forProvider` field.  

Settings in `initProvider` aren't enforced by Crossplane. Crossplane ignores any
changes to a `initProvider` field.
{{</hint >}}

Using `initProvider` is useful for setting initial values that a Provider may
automatically change, like an auto scaling group.

For example, creating a 
{{<hover label="initProvider" line="2">}}NodeGroup{{</hover>}} 
with an initial 
{{<hover label="initProvider" line="9">}}desiredSize{{</hover>}}.  
Crossplane doesn't change the 
{{<hover label="initProvider" line="9">}}desiredSize{{</hover>}}
setting back when the Provider auto scales the Node Group.

{{< hint "tip" >}}
Crossplane recommends configuring
{{<hover label="initProvider" line="6">}}managementPolicies{{</hover>}} without
`LateInitialize` to avoid conflicts with `initProvider` settings.
{{< /hint >}}
```

content/master/concepts/managed-resources.md
```diff
@@ -208,40 +208,142 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
+  forProvider:
+    region: us-west-1
+    scalingConfig:
+      - maxSize: 4
+        minSize: 1
+```
+
+It's suggested to use a combination of `managementPolicies` without the
+`LateInitialize` action when using `initProvider` to avoid late initialization
+of fields in `forProvider`. Read more about [Late Initialization]({{<ref "./managed-resources#late-initialization" >}})
```
plumbis, line 346: I moved this to a `{{<hint "tip">}}` above the code block.

plumbis (changes requested):
Thanks for all the work Lovro!  There are a lot of suggestions but it's mainly big chunks of moving things around. 

Let me know if you have any questions or objections to my suggestions.

content/master/concepts/managed-resources.md
lsviben: Applied, just added a slight change to

```
Settings in `forProvider` are always enforced by Crossplane. Crossplane reverts
any changes to a `forProvider` field.  

Settings in `initProvider` aren't enforced by Crossplane. Crossplane ignores any
changes to a `initProvider` field.
```
to 
```
Settings in `forProvider` are always enforced by Crossplane. Crossplane reverts
any changes to a `forProvider` field in the external resource.

Settings in `initProvider` aren't enforced by Crossplane. Crossplane ignores any
changes to a `initProvider` field in the external resource.
```

To make a clear distinction in making changes in managed its external resource

content/master/concepts/managed-resources.md
```diff
@@ -208,40 +208,142 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
+managed and its external resource. `managementPolicies` is an array of actions
+(`Observe`, `Create`, `Update`, `Delete`, `LateInitialize`, `*`), and there are
+multiple combinations that Crossplane supports based on the use case.
+
+For example, setting just `["Observe"]` makes Crossplane treat the resource as
+a read-only resource, importing the external resources not originally created
+by Crossplane. This allows other managed resources to reference the read-only
+resource, for example, a shared database or network. Importing a resource as
+read-only is convenient since Crossplane wont attempt to take control and
+only identifier parameters are needed to be specified in the managed resource.
```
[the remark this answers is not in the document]
lsviben: btw this table in dark mode is unreadable, do you know how to fix that?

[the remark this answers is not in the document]
plumbis: Well that's a CSS design bug 😆

Can you add this:
```
    //Striped table font color
    .table-striped-columns > :not(caption) > tr > :nth-child(even){
        color: inherit !important;
    }
```

before the final closing brace `}`

to  `themes/geekboot/assets/scss/dark-mode.scss`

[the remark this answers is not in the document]
lsviben: thanks, looks good now

plumbis (approved):
changes to fix vale and improve table display. Otherwise LGTM. Approving, assuming there's no problems with the suggestions.

content/master/concepts/managed-resources.md
```diff
@@ -208,40 +208,142 @@ resource object is deleted from Kubernetes and the `deletionPolicy` is
...
+For some of the optional fields, users rely on the default that the cloud
+provider chooses for them. Since Crossplane treats the managed resource as the
+source of the truth, values of those fields need to exist in `spec` of the
+managed resource. In each reconciliation, Crossplane will fill the value of
+a field that's left empty by the user but is assigned a value by the provider.
+For example, there could be two fields like `region` and `availabilityZone` and
+you might want to give only `region` and leave the availability zone to be
+chosen by the cloud provider. In that case, if the provider assigns an
+availability zone, Crossplane gets that value and fills `availabilityZone`. Note
+that if the field is already filled, the controller won't override its value.
```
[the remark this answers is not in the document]
turkenh: Pushed a commit to fix that.

turkenh (approved):
Thanks @lsviben, looking good to me!

I just pushed a couple of comments to fix Vale errors and make some improvements in the late initialization section.

@plumbis feel free to have another look, or please let me know if we can hit the merge button.

PR #522, opened 2023-08-08
Minor color improvements

shanecmiller23:
- Improve contrast for accordions/expandable blocks
- Improve coloring for "Join the Crossplane Slack" home page tooltip
- Replaced all old color variables with the new palette variables
- Fix dark mode coloring for vertical striped tables
- Remove table aqua background hover states

### DIFF
--- a/themes/geekboot/assets/scss/dark-mode.scss
+++ b/themes/geekboot/assets/scss/dark-mode.scss
@@ -86,10 +86,14 @@
     }
     .table {
         border: 1px solid #{$fog-700};
+        --bs-table-hover-bg: #{$fog-1000};
     }
-    .table-striped > tbody > tr:nth-of-type(odd) > * {
+    .table-striped > tbody > tr:nth-of-type(odd) > *, .table-striped-columns > :not(caption) > tr > :nth-child(even) {
         --bs-table-accent-bg: #{$fog-950};
         color: #{$fog-0};
+    }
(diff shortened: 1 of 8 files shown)


themes/geekboot/assets/scss/dark-mode.scss
```diff
@@ -51,7 +51,7 @@
 
     // Override Bootstrap accordion colors in the body
     .accordion {
-        --bs-accordion-border-color: #{$fog-800};
+        --bs-accordion-border-color: #{$fog-500};
```
jbw976, line 54: There's something fishy going on with the bottom rounded corners of the accordions when they are in their expanded state - looks like it's still a dark color that doesn't match the straight line edge borders, or they are clipped or something? they look like they are missing essentially - do you see this too?

e.g. on https://deploy-preview-522--crossplane.netlify.app/v1.13/concepts/compositions/ 

Screenshot 2023-08-08 at 10 08 16 AM

plumbis: I see the same thing in dark mode.

shanecmiller23: @jbw976 @plumbis I'll need to dig into that issue to find more. Being that issue is pretty minor and exists today, I may scope/investigate that in a separate PR if that works

plumbis (approved):

PR #524, opened 2023-08-08
add doc for adler32 transform

stevendborrelli:
Add new transform from upstream PR https://github.com/crossplane/crossplane/pull/4417.

PR #526, opened 2023-08-08
Revert "add doc for adler32 transform"

plumbis:
Reverts crossplane/docs#524


phisco:
@plumbis we finally merged https://github.com/crossplane/crossplane/pull/4417, this one can be closed.

PR #544, opened 2023-09-05
feat: optional environment label selector fieldpath

phisco:
Documenting https://github.com/crossplane/crossplane/pull/4547 and https://github.com/crossplane/crossplane/pull/4568.

### DIFF
--- a/content/master/concepts/environment-configs.md
+++ b/content/master/concepts/environment-configs.md
@@ -196,15 +196,75 @@ spec:
     - type: Selector
       selector: 
         matchLabels:
-          - key: my-label-key
+          - key: my-first-label-key
             type: Value
-            value: my-label-value
-          - key: my-label-key
+            value: my-first-label-value
+          - key: my-second-label-key
             type: FromCompositeFieldPath
             valueFromFieldPath: spec.parameters.deploy
   resources:
   # Removed for brevity
 ```
+
+By default, Crossplane errors out if a
+{{<hover label="byLabel" line="16">}}valueFromFieldPath{{</hover>}}
+field doesn't exist in the composite resource at runtime.  
+Set the
+{{<hover label="byLabelOptional" line="17">}}fromFieldPathPolicy{{</hover>}}
+field to `Optional` to ignore fields that don't exist. 
+
+```yaml {label="byLabelOptional",copy-lines="all"}
+apiVersion: apiextensions.crossplane.io/v1
+kind: Composition
+metadata:
+  name: example-composition
+spec:
+  environment:
+    environmentConfigs:
+      - type: Selector
+        selector:
+          matchLabels:
+            - key: my-first-label-key
+              type: Value
+              value: my-first-label-value
+            - key: my-second-label-key
+              type: FromCompositeFieldPath
+              valueFromFieldPath: spec.parameters.deploy
+              fromFieldPathPolicy: Optional
+  resources:
+  # Removed for brevity
+```
+
+Crossplane evaluates label selectors in order, so if a label set as optional
+isn't found, but if you already defined an explicit value for it, it uses that
+{{<hover label="byLabelOptionalDefault" line="16">}}default value{{</hover>}}
+instead.
+
+```yaml {label="byLabelOptionalDefault",copy-lines="all"}
+apiVersion: apiextensions.crossplane.io/v1
+kind: Composition
+metadata:
+  name: example-composition
+spec:
+  environment:
+    environmentConfigs:
+      - type: Selector
+        selector:
+          matchLabels:
+            - key: my-first-label-key
+              type: Value
+              value: my-label-value
+            - key: my-second-label-key
+              type: Value
+              value: my-default-value
+            - key: my-second-label-key
+              type: FromCompositeFieldPath
+              valueFromFieldPath: spec.parameters.deploy
+              fromFieldPathPolicy: Optional
+  resources:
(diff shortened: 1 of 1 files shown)


content/master/concepts/environment-configs.md
```diff
@@ -196,16 +196,76 @@ spec:
...
+            - key: my-second-label-key
+              type: FromCompositeFieldPath
+              valueFromFieldPath: spec.parameters.deploy
+              fromFieldPathPolicy: Optional
+  resources:
+  # Removed for brevity
+```
+
+Crossplane evaluates label selectors in order, so if a label set as optional
+isn't found, but if you already defined an explicit value for it, it uses that
```
plumbis, line 239: "Evaluates in order" and "already defined"

Does the definition have to come first? Does it behave the same if the order is:

```yaml
        selector:
          matchLabels:
            - key: my-first-label-key
              type: Value
              value: my-label-value
            - key: my-second-label-key
              type: FromCompositeFieldPath
              valueFromFieldPath: spec.parameters.deploy
              fromFieldPathPolicy: Optional
            - key: my-second-label-key
              type: Value
              value: my-default-value
```

content/master/concepts/environment-configs.md
```diff
@@ -196,16 +196,76 @@ spec:
...
   resources:
   # Removed for brevity
 ```
 
+By default, Crossplane errors out if a
+{{<hover label="byLabel" line="16">}}valueFromFieldPath{{</hover>}}
+field doesn't exist in the composite resource at runtime. You can change this
+behavior by setting the
+{{<hover label="byLabelOptional" line="17">}}fromFieldPathPolicy{{</hover>}}
+field to `Optional`, ignoring the label altogether, if not found.
```
plumbis, line 214:
```suggestion
By default, Crossplane errors out if a
{{<hover label="byLabel" line="16">}}valueFromFieldPath{{</hover>}}
field doesn't exist in the composite resource at runtime.  
Set the
{{<hover label="byLabelOptional" line="17">}}fromFieldPathPolicy{{</hover>}}
field to `Optional` to ignore fields that don't exist. 
```

Trying to make the word more explicit "Do $x to achieve $y" instead of "hey, you can do this if you want"

plumbis (changes requested):

content/master/concepts/environment-configs.md
phisco: no, it wouldn't behave the same, `my-second-label-key: my-default-value` would always overwrite the value coming from `spec.parameters.deploy` if they were defined this way.

content/master/concepts/environment-configs.md
```diff
@@ -196,16 +196,76 @@ spec:
...
+            - key: my-second-label-key
+              type: Value
+              value: my-default-value
+            - key: my-second-label-key
+              type: FromCompositeFieldPath
+              valueFromFieldPath: spec.parameters.deploy
+              fromFieldPathPolicy: Optional
+  resources:
+  # Removed for brevity
+```
```
plumbis, line 266:
```suggestion
Set a default value for an optional label by setting the 
{{<hover label="byLabelOptionalDefault" line="15">}}value{{</hover>}} for the 
{{<hover label="byLabelOptionalDefault" line="14">}}key{{</hover>}} first, then
define the 
{{<hover label="byLabelOptionalDefault" line="20">}}Optional{{</hover>}} label.

For example, this Composition defines
{{<hover label="byLabelOptionalDefault" line="16">}}value: my-default-value{{</hover>}} 
for the key {{<hover label="byLabelOptionalDefault" line="14">}}my-second-label-key{{</hover>}}.   
If the label
{{<hover label="byLabelOptionalDefault" line="17">}}my-second-label-key{{</hover>}}
exists, Crossplane uses the value from the label instead.

```yaml {label="byLabelOptionalDefault",copy-lines="all"}
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: example-composition
spec:
  environment:
    environmentConfigs:
      - type: Selector
        selector:
          matchLabels:
            - key: my-first-label-key
              type: Value
              value: my-label-value
            - key: my-second-label-key
              type: Value
              value: my-default-value
            - key: my-second-label-key
              type: FromCompositeFieldPath
              valueFromFieldPath: spec.parameters.deploy
              fromFieldPathPolicy: Optional
  resources:
  # Removed for brevity
```

{{<hint "warning" >}}
Crossplane applies values in order. The value of the last key defined always takes precedence.  

Defining the default value _after_ the label always overwrites the label
value. 
{{< /hint >}}
```

This is a suggested edit, you can throw it away if you'd like. I'm happy to approve as-is.

The edit makes it more explicit to say "Define a default like this..." and adds a warning that if they get the order wrong it won't work.

Comment (phisco) [reply content/master/concepts/environment-configs.md]: added manually in a separate commit, as otherwise DCO gets messed up.

File: content/master/concepts/environment-configs.md [hunk 4]
```diff
@@ -196,16 +196,90 @@ spec:
...
+              value: my-first-label-value
+            - key: my-second-label-key
+              type: FromCompositeFieldPath
+              valueFromFieldPath: spec.parameters.deploy
+              fromFieldPathPolicy: Optional
+  resources:
+  # Removed for brevity
+```
+
+Set a default value for an optional label by setting the
```
turkenh, line 238: This is a great use case for `fromFieldPathPolicy` 🙌

PR #546, opened 2023-09-06
Update composition revisions doc to remove alpha notice

recht:


plumbis (approved):
Thanks for fixing this oversight! 🙌

PR #555, opened 2023-09-21
Fixed formatting of annotation table in managed-resources.md.

jamesdobson:
The text "began creating a new managed resource." was in a table row by itself. Also, changed article from "a" to "the" for consistency with other rows, and removed the word "new" as it is redundant.


plumbis:
Awesome, thank you! 

Would you mind adding the same change to the other versions of `managed-resources.md` in the `/content/` folder?

jamesdobson:
Updated the other versions, thanks @plumbis!

plumbis (approved):

Issue #556, opened 2023-09-22
Include information about Crossplane Package Dependency for configurations.


haarchri:
### What's Missing?

https://docs.crossplane.io/knowledge-base/guides/troubleshoot/#handling-crossplane-package-dependency

Today, we have an example for Crossplane package dependency for providers, but there is no information indicating that this is also possible for configurations.

check: 
https://github.com/upbound/crossplane/blob/6221afd3638273aa1f66e736931619ee1153a5ac/design/one-pager-package-format-v2.md?plain=1#L197-L213

spec also not covers this information: https://github.com/crossplane/crossplane/blob/master/contributing/specifications/xpkg.md

new example could look like:

```
apiVersion: meta.pkg.crossplane.io/v1
kind: Configuration
metadata:
  name: test-configuration
  annotations:
    provider: aws
spec:
  crossplane:
    version: ">=v1.12.1-0"
  dependsOn:
    - configuration: xpkg.upbound.io/upbound/configuration-caas
      version: ">=v0.2.0"
    - provider: xpkg.upbound.io/upbound/provider-aws-s3
      version: ">=v0.38.0"
    - provider: xpkg.upbound.io/upbound/provider-aws-kms
      version: ">=v0.38.0"
```

plumbis:
I'm working on updating the concepts/packages chapter right now, I'll make sure to fold some of this info in.

Issue #454, opened 2023-05-25
Update concepts/packages


plumbis:
Update the concepts/packages doc.

Related to epic #450

Issue #277, opened 2022-10-04
Document manual population of package cache


hasheddan:
### What problem are you facing?

Populating Crossplane's package cache manually can be useful to avoid having to push package images to a registry during development. One place we do this today is in CI for providers. However, there is not currently good documentation on how to do it.

### How could Crossplane help solve your problem?

Add documentation for manually populating package cache.

plumbis:
Discussed in the provider chapter

https://docs.crossplane.io/latest/concepts/providers/#install-offline

PR #566, opened 2023-10-10
Rewrite packages content and split into Config and Provider chapters

### LINKED ISSUE #493: Install Crossplane offline
### What's Missing?

How do you install crossplane in an airgapped environment? Include info about installing providers as well. 

Comment from slack

>You can pull the public images, tag them with your private registry and push them. It should work for both Crossplane and providers (edited)

plumbis:
Updates and expands the Packages chapter. The chapter is now only focused on Configuration packages.

As a result the Providers chapter has been expanded to cover package related elements in the context of Providers.

Because the title of the page changed changing the menu from `v1.13` to `master` doesn't work for the new Packages chapter. The fix for this is in #567 and will fix these pages when merged.

Packages: https://deploy-preview-566--crossplane.netlify.app/master/concepts/packages/

Providers: https://deploy-preview-566--crossplane.netlify.app/master/concepts/providers

Note: This references the new Crossplane CLI but doesn't add new docs for the CLI. We'll need to come back and add links after new CLI docs are published.

Resolves #493, #556, #454, #450, #276, #277


plumbis:
Vale error is on the KB page, new content is passing.

jbw976:
> This references the new Crossplane CLI but doesn't add new docs for the CLI.

this is being tracked in https://github.com/crossplane/crossplane/issues/4761, which you are more than welcome to contribute to if you have the time 😄

content/master/concepts/providers.md
```diff
@@ -62,14 +58,35 @@ spec:
...
 (`crossplane-system`).
 
+{{<hint "note" >}}
+Providers are part of the 
+{{<hover label="install" line="1">}}pkg.crossplane.io{{</hover>}} group.  
+
+The {{<hover label="meta-pkg" line="1">}}meta.pkg.crossplane.io{{</hover>}}
+group is for creating Provider packages. 
+
+Instructions on building Providers are outside of the scope of this
```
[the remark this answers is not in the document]
plumbis: Not in this PR. I would assume a simple `crossplane build provider...` cli command reference would be in a related CLI doc. #TODO

content/master/concepts/packages.md
```diff
@@ -1,513 +1,525 @@
...
+platform-ref-aws-3ac761211893   True      1          xpkg.upbound.io/upbound/platform-ref-aws:v0.4.1   Inactive                               5m13s
 ```
 
-If the Provider package is valid, you will see a file with the `.xpkg`
-extension.
+Only a single revision is active at a time. The active revision determines the
+available resources, including Compositions and Composite Resource Definitions. 
+
+By default Crossplane keeps only a single _Inactive_ revision. Change the number
+of older revisions by setting the
```
tr0njavolta, line 125:
```suggestion
By default Crossplane keeps only a single _Inactive_ revision. 
```

content/master/concepts/packages.md
```diff
@@ -1,513 +1,525 @@
...
-```
+Change the number of revisions Crossplane maintains with a Configuration Package 
+{{<hover label="revHistoryLimit" line="6">}}revisionHistoryLimit{{</hover>}}. 
+
+The {{<hover label="revHistoryLimit" line="6">}}revisionHistoryLimit{{</hover>}}
+field is an integer.  
+The default value is `1`.  
+Disable storing revisions by setting 
+{{<hover label="revHistoryLimit" line="6">}}revisionHistoryLimit{{</hover>}} 
+to `0`.
```
tr0njavolta, line 244:
```suggestion
```

content/master/concepts/packages.md
```diff
@@ -1,513 +1,525 @@
...
 
+```yaml {label="revHistoryLimit"}
+apiVersion: pkg.crossplane.io/v1
+kind: Configuration
+metadata:
+  name: platform-ref-aws
+spec:
+  revisionHistoryLimit: 10
+# Removed for brevity
 ```
```
tr0njavolta, line 257:
```suggestion
```

content/master/concepts/packages.md
```diff
@@ -1,513 +1,525 @@
...
+Crossplane API group.
 
-Packaged `Provider` controllers are installed in the form of a `Deployment`.
-Crossplane populates the `Deployment` with default values that may not be
-appropriate for every use-case. In the event that a user wants to override some
-of the defaults that Crossplane has set, they may create and reference a
-`ControllerConfig`.
+Specify any other Configurations, Functions or Providers in the 
+{{<hover label="cfgMeta" line="7">}}dependsOn{{</hover>}} list.  
+Optionally you can require a specific or minimum package version with the
```
tr0njavolta, line 445:
```suggestion
Optionally, you can require a specific or minimum package version with the 
```

content/master/concepts/packages.md
```diff
@@ -1,513 +1,525 @@
...
+{{<hover label="cfgMeta" line="7">}}dependsOn{{</hover>}} list.  
+Optionally you can require a specific or minimum package version with the 
+{{<hover label="cfgMeta" line="9">}}version{{</hover>}} option.
 
-An example of when this may be useful is when a user is running Crossplane on
-EKS and wants to take advantage of [IAM Roles for Service Accounts]. This
-requires setting an `fsGroup` and annotating the `ServiceAccount` that
-Crossplane creates for the controller. This could be accomplished with the
-following `ControllerConfig` and `Provider`:
+You may also define a specific or minimum version of Crossplane for this
```
tr0njavolta, line 448:
```suggestion
You can also define a specific or minimum version of Crossplane for this
```

tr0njavolta (commented):
Some suggestions. Removed the second instance of the revisionHistory example in the providers file since you link to the section containing the same example.

content/master/concepts/packages.md
```diff
@@ -1,513 +1,489 @@
...
-### Provider Packages
-
-A Provider package contains a `crossplane.yaml` with the following format:
-
-```yaml
-apiVersion: meta.pkg.crossplane.io/v1
-kind: Provider
+
+A _Configuration_ package is an 
+[OCI container images](https://opencontainers.org/) containing a collection of
```
turkenh, line 8:
```suggestion
[OCI container image](https://opencontainers.org/) containing a collection of
```

content/master/concepts/packages.md
```diff
@@ -1,513 +1,489 @@
...
+configuring the Crossplane Pod settings in the 
+[Crossplane install documentation]({{<ref "../software/install#customize-the-crossplane-helm-chart">}}).
 
-For an example Provider package, see [provider-gcp].
+Provide the name of the Configuration's `.xpkg` file and set 
+{{<hover label="offline" line="7">}}packagePullPolicy: Never{{</hover>}}.
 
-To build a Provider package, navigate to the package root directory and execute
-the following command:
+For example, to install a locally downloaded version of
```
turkenh, line 84: nit: which I believe also a pretty common use case.

```suggestion
For example, to install a locally downloaded or built version of 
```

content/master/concepts/packages.md
```diff
@@ -1,513 +1,489 @@
...
 ```
-crossplane build provider
+
+### Installation options
+
+Configurations support multiple options to change configuration package related
+settings. 
+
+
+#### Configuration revisions
```
turkenh, line 106:
```suggestion
#### Revision History Limit
```

I found a bit confusing not to have titles the same as the option. Given these are subsections under `Installation Options`, I was expecting to see the configuration as is (e.g. `Revision activation policy` instead of `Upgrade policy` or `Skip Dependency Resolution` instead of `Ignore Dependencies`) as the title.

content/master/concepts/packages.md
```diff
@@ -1,513 +1,489 @@
...
+# Removed for brevity
 ```
 
-If the Provider package is valid, you will see a file with the `.xpkg`
-extension.
+#### Upgrade policy
 
-## Pushing a Package
+Crossplane automatically upgrades a Configuration the to the latest version 
+available in the package cache.
```
turkenh, line 188: This doesn't have to be the case and could be misleading. 

Imagine the following:

1. Created a configuration package with `xpkg.upbound.io/upbound/platform-ref-aws:v1` 
2. XP downloaded `v1` to cache and installed it.
3. Updated `Configuration` spec.package to `xpkg.upbound.io/upbound/platform-ref-aws:v2` 
4. XP downloaded `v2` to the cache and upgraded to it with a new revision.
5. For some reason I decided to downgrade and change `Configuration` spec.package to `xpkg.upbound.io/upbound/platform-ref-aws:v1` 

At this state, while both `v1` and `v2` are in the cache, XP would activate `v1` as it is the desired state instead of _automatically upgrading to the latest version available in the package cache_.

Revision Activation Policy is about automatically marking the latest desired revision as active, and for Configuration packages, this doesn't have much user-facing impact in contrast to Providers where active revisions controller would process MRs. An active configuration revision would only own installed manifests (e.g. XRDs and Compositions`) as `controller=true`, which doesn't matter for the user.

content/master/concepts/packages.md
```diff
@@ -1,513 +1,489 @@
...
-```
-crossplane install provider xpkg.upbound.io/crossplane-contrib/provider-gcp:v0.22.0
-```
+The {{<hover label="pps" line="6">}}packagePullSecrets{{</hover>}} is a list of
+secrets.
 
-To install a Configuration package, execute the following command:
+For example, to use the secret named
+{{<hover label="pps" line="6">}}example-secret{{</hover>}} configure a 
+{{<hover label="pps" line="6">}}packagePullSecrets{{</hover>}}.
```
turkenh, line 243: It is worth adding a note stating that this secret needs to be in the namespace where Crossplane runs.

content/master/concepts/packages.md
```diff
@@ -1,513 +1,489 @@
...
-
-If `skipDependencyResolution: true`, the package manager will install a package
-without considering its dependencies.
 
-### spec.ignoreCrossplaneConstraints
+### Verify a Configuration
 
-Valid values: `true` or `false` (default: `false`)
+Verify a Configuration with 
+{{<hover label="verify" line="1">}}kubectl get configuration{{</hover >}}.
```
turkenh, line 311: nit: The sentence feels a bit incomplete. Extending the sentence about expecting to see `Installed` and `Healthy` as `True` could help.

content/master/concepts/packages.md
```diff
@@ -1,513 +1,489 @@
...
+```
 
-### spec.controllerConfigRef
+### Manage dependencies
 
-{{< hint "warning" >}}
-The `ControllerConfig` API has been deprecated and will be removed in a future
-release when a comparable alternative is available.
+Configuration packages may include dependencies on other packages including
+Providers or other Configurations.
```
turkenh, line 322:
```suggestion
Providers, Functions or other Configurations. 
```

content/master/concepts/packages.md
```diff
@@ -1,513 +1,489 @@
...
+The CLI recursively searches for `.yml` or `.yaml` files in the directory to
+include in the package.
+
+{{<hint "important" >}}
+You must ignore any other YAML files with `--ignore=<file_list>`.  
+For
+example, `crossplane build configuration -f test-directory --ignore=".tmp/*,other-file.yaml"`.
+
+Including YAML files that aren't Compositions or CompositeResourceDefinitions, 
+including Claims isn't supported.
```
turkenh, line 452: nit: The two `including`s made the sentence a bit hard to grasp :)

content/master/concepts/providers.md
```diff
@@ -89,28 +106,277 @@ crossplane-stable/crossplane \
...
+apiVersion: pkg.crossplane.io/v1
+kind: Provider
+metadata:
+  name: offline-provider-aws
+spec:
+  package: provider-aws
+  packagePullPolicy: Never
+```
+
+### Installation options
```
turkenh, line 139: Most of my comments on `Configurations` side also apply here, so I am not duplicating them here.

turkenh (commented):
@plumbis nice work 🙌 

I have a concern regarding the new structure though. There are quite a lot of similar (if not duplicate) content on both `Configurations` and `Providers` pages now. Soon, we will need to have them under `Functions` as well, and for possible future extensions (e.g. Secret Store Plugins). So, I feel like having those content in a single place but referring them from specific package pages could be a more scalable and maintainable approach.

A related example is, I recently started documenting `DeploymentRuntimeConfigs` [here](https://github.com/crossplane/docs/pull/570) under `Providers`. They will also be relevant for Functions and I was planning to refer from `Functions` to `Providers` not to duplicate the same content which is not great. However, I was thinking having a single place, which is higher level to specific packages types could be a better place and was wondering whether I should move under https://docs.crossplane.io/v1.13/concepts/packages/#installing-a-package which seems to be removed with this PR.

So, my suggestion would be to keep `Crossplane Packages` section and add `Configurations` as a new one instead of converting the former to the latter.

plumbis:
@turkenh thank you for your time and feedback on this! 

I must have made three drafts of this going back and forth between duplicating and centralizing the content. I ended up landing on breaking it up and duplicating for a few reasons:

1.) The user workflow is going to be around a task: use a configuration, install and modify a provider, install or modify a function.  

2.) The concept of a "package" is apparent if you're building a package or you're an experienced Crossplane user. If I'm trying to modify my provider it's not obvious that "package" is the right section to look under for information about my configuration options.

3.) The examples, outputs and terms are just different enough to make it confusing when it's centralized. For example, Configurations and Providers can specify different dependencies. Installing a Provider may take minutes, but installing a Configuration shouldn't. Creating a config and a provider are very different processes that just happen to use the same packaging. 

There are authoring options to centralize some parts and import them but this makes the authoring process a bit of a pain in the ass. Honestly I'd rather require us to be mindful of how a feature like `DeploymentRuntimeConfigs` is used and configured in relation to each of the respective top level objects. Your comment on Revision Activation Policy  is a great example where the explanation  and impact changes based on the object it's modifying.

content/master/concepts/packages.md
turkenh: > On the other hand, we will have users wanting to accomplish a task but not knowing what Crossplane term to look for (for example Revision History) and will want to search for something more generic or task based (like "Configuration revisions").

I would see the task as "Limiting revision history"; maybe I am biased as someone who knows internals. However, Crossplane terms are also selected with the best effort to reflect the task/configuration, so feel like they are not that 
different than what they are for.

Having a section talking about `Configuration Revisions` (or Package `Revisions` in general) makes sense to me in general, but I find it confusing to have it as a Subsection under `Installation options`. 

Search or "ctrl + f" would work in any case, so I don't think this is something we should consider while looking for a good structure.

content/master/concepts/packages.md
turkenh: > If I change the XRD APIs between v1 and v2, wouldn't downgrading lose those APIs?

Yes, but I am not sure I understood the relation with the revisionActivationPolicy here 🤔 

You can read more details from the [design](https://github.com/crossplane/crossplane/blob/master/design/design-doc-packages-v2.md), but I couldn't find an easy answer there so let me try to explain:

Revision Activation Policy is all about the Package reconciler automatically marking a revision as `Active` or not. It doesn't impact which version is installed at any instant. `spec.package`  field of the package is authoritative on what version should be installed or not. 

Putting `revisionActivationPolicy` aside for a second, if you have the following `Configuration`, it will install `v1`:

```
apiVersion: pkg.crossplane.io/v1
kind: Configuration
metadata:
  name: platform-ref-aws
spec:
  package: xpkg.upbound.io/upbound/platform-ref-aws:v1
```

If you change it below, it will install `v2`:

```
apiVersion: pkg.crossplane.io/v1
kind: Configuration
metadata:
  name: platform-ref-aws
spec:
  package: xpkg.upbound.io/upbound/platform-ref-aws:v2
```

If you change it back to the previous spec, it will again install `v1`. The available versions in the package cache or in the registry doesn't have any impact here, you may have `v5` here or there, but the package manager will just install what you asked in the `Configuration` spec. It is your responsibility to have backward-compatible upgrades or supporting rollbacks, and package manager will just do whatever it is asked to do.

turkenh: So, what is `revisionActivationPolicy` then?

You can read more details from the [design](https://github.com/crossplane/crossplane/blob/master/design/design-doc-packages-v2.md), but I couldn't find an easy answer there, so let me try to explain:

Revision Activation Policy is all about the Package reconciler automatically marking a revision as `Active` or not. It doesn't impact which version is **installed** at any instant. `spec.package`  field of the package is authoritative on what version should be installed or not, as I explained above. 

So the question is, what is an active revision? A revision being active has the following meanings:

- Despite all package revisions owning the CRD, XRD, or Composition installed by the package, only the active one controls it with `controller: true` in the ownerRef. 
  - [This post](https://stackoverflow.com/a/65825463/4124416) explains the technical details well, but, in our context, we can think it as, which revision would be responsible for acting on the instances of those resources. 
  - As I stated above, this doesn't have a meaning for `Configuration` packages, since no controller is running for them different than `Provider`.
- Only the controller for the `Active` revision will be running to act on the instances. For a given `Provider` with `v1` and `v2` revisions installed where `v2` is marked as active, the provider controller for `v2` will be running and operating on the instances of the types that the provider installed, a.k.a. Managed Resources. 

When revisionActivationPolicy is automatic, the package manager will simply set the desired version (i.e. spec.package` in the `Provider) as active automatically.

As an example, I installed the following:

```
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-helm
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-helm:v0.14.0
  revisionActivationPolicy: Automatic
```

There is only 1 revision, which is for `v0.14.0`, and marked as `Active` by package manager (since we used Automatic policy)

```
❯ kubectl get providerrevisions.pkg.crossplane.io
NAME                         HEALTHY   REVISION   IMAGE                                                      STATE    DEP-FOUND   DEP-INSTALLED   AGE
provider-helm-ce18dd03e6e4   True      1          xpkg.upbound.io/crossplane-contrib/provider-helm:v0.14.0   Active                               50s
```

If I update the `spec.package` of `Provider` to `v0.15.0`:

```
❯ kubectl get providerrevisions.pkg.crossplane.io
NAME                         HEALTHY   REVISION   IMAGE                                                      STATE      DEP-FOUND   DEP-INSTALLED   AGE
provider-helm-503c3591121b   True      2          xpkg.upbound.io/crossplane-contrib/provider-helm:v0.15.0   Active                                 12s
provider-helm-ce18dd03e6e4   True      1          xpkg.upbound.io/crossplane-contrib/provider-helm:v0.14.0   Inactive                               2m40s
``` 

Not `v0.15.0` is marked as `Active` automatically. Meaning, that versions controller will process my MRs and I see the provider controller is running with its version:

```
❯ kubectl -n crossplane-system get pods provider-helm-503c3591121b-77fb968677-p6zp9 -o yaml | grep image
    image: xpkg.upbound.io/crossplane-contrib/provider-helm:v0.15.0
```

If I used `revisionActivationPolicy: Manual` in the above flow:

```
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-helm
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-helm:v0.14.0
  revisionActivationPolicy: Manual
``` 

It will install `v0.14.0` by creating a revision but won't mark it as `Active`, so no provider pod is running at all. No one will control, hence process the MRs of this provider. (For some reason, noticing that they were not initialized as `Inactive` either, but probably a small non-functional bug, we now they are not active)

```
❯ kubectl get providerrevisions.pkg.crossplane.io -w
NAME                         HEALTHY   REVISION   IMAGE                                                      STATE   DEP-FOUND   DEP-INSTALLED   AGE
provider-helm-ce18dd03e6e4   True      1          xpkg.upbound.io/crossplane-contrib/provider-helm:v0.14.0                                       13s

❯ kubectl -n crossplane-system get pods | grep provider-helm

```

Similarly, if I upgrade to `v0.15.0`, it will install it but still not run its controller since it will not be activated either.

We don't expect users to interact with this option much except for some advanced use cases. We only leveraged that during migration from monolith to family providers, to properly transfer ownership.

turkenh:
> The concept of a "package" is apparent if you're building a package or you're an experienced Crossplane user. 

I think I see your point in that a user is usually not aware that they are installing a Crossplane Package when they are installing a Configuration. However, could this be related to how the existing [packages](https://docs.crossplane.io/v1.13/concepts/packages/) page was structured and us not doing a good job introducing the concept clearly 🤔 

> I must have made three drafts of this going back and forth between duplicating and centralizing the content. I ended up breaking it up and duplicating it for a few reasons:

I see your points and agree with them. My primary concern is maintenance; once you have similar (or even the same) content in multiple places, they eventually diverge. I really wish we could find a way to avoid this. Imagine that I want to document the behavior of `revisionActivationPolicy` I detailed [above](https://github.com/crossplane/docs/pull/566#discussion_r1371319663). I would never feel good about putting that amount of information in two (even 3) different places and would probably avoid diving deep in that case.

`DeploymentRuntimeConfigs` is another example. It could be ok to duplicate if I just put a paragraph for it. But when I want to put a good amount of details and examples to clarify the topic, it is really concerning that I need to do it in two different places and keep maintaining. 

The ideal structure in my mind would be something like below:

- Crossplane Packages
   - What is a Crossplane Package? What packages do we support today?
     - Configurations: a short description and link to the dedicated section.
     - Providers: a short description and link to the dedicated section.
     - Functions: a short description and link to the dedicated section.
   - What is a Package Revision?
     - The relation between Packages and PackageRevisions
     - What is an Active Revision
   - Installing or Upgrading a Package
   - Package Installation Options:
     - Revision History Limit
     - Revision Activation Policy
     - ... 
   - Packages with Runtime: Provider and Functions are Crossplane Packages with a runtime, whereas Configuration is not.
     - DeploymentRuntimeConfig: How to configure the runtime of a package.
   - ~Building a Package~ We can move these to a developer-focused page/doc.
   - ~Pushing a Package~ We can move these to a developer-focused page/doc.
- Configurations
  - What are Configurations?
  - Deploying a Configuration
    - Short examples on how to install
    - For more details on package installation options, see Crossplane Packages/Package Spec.
- Providers
  - What is a Provider?
  - Deploying a Provider
    - Short examples on how to install
    - For more details on installation options, see Crossplane Packages/Package Spec.
    - Runtime Configuration, see Crossplane Packages/DeploymentRuntimeConfig
    - ProviderConfigs
- Functions
  - What is a Function?
  - Deploying a Function
    - Short examples on how to install
    - For more details on package installation options, see Crossplane Packages/Package Spec.
    - Runtime Configuration, see Crossplane Packages/DeploymentRuntimeConfig
    - ProviderConfigs

negz:
> I must have made three drafts of this going back and forth between duplicating and centralizing the content.

@plumbis @turkenh I wonder if there's some middle ground here? I came looking at this PR as I started documenting how to install a Function. I started copying what was on the Provider page, then realized if I continued down that path the reader would be scrolling through pages of detailed "how to install a package" content before they ever even saw a Composition that used a Function.

I do agree there's value in repeating the really common things per package type. For example high level examples of how to install, update, and delete each package. In some cases there's extra context that is only relevant to a particular type of package - for example being wary that deleting a provider will delete any MRs it owns.

However some of the detail feels ripe to deduplicate into a "package details" page. For me this would include all the conditions a package can have, and most of the installation options that are common to all packages.

plumbis:
I think both of y'all have great points that should be addressed. 

What I would propose is that if there are no objections to the content, let's merge this. It's duplicating content but it's also a large improvement over the existing packages chapter.

Sometime in the 1.15 time frame I can take on some de-dup and organizational work, is that acceptable?

negz:
> What I would propose is that if there are no objections to the content, let's merge this.

@plumbis Sounds good to me regarding figuring out the duplication later. I haven't actually reviewed this new content in any depth though so I defer to @tr0njavolta and @turkenh WRT whether it's ready to merge.

Note that I can review if it would help, but my assumption is that @turkenh has already gone deep enough to potentially approve faster than waiting for me to review this.

Part of https://github.com/crossplane/docs/pull/583 is going to be updating the packages docs to make sure they mention/cover Functions. I'm hopeful that won't be too hard since it should mostly be a copy of the provider docs, but I'd like to see this merged to unblock that.

turkenh:
> What I would propose is that if there are no objections to the content, let's merge this. It's duplicating content but it's also a large improvement over the existing packages chapter.

I am fine with following up for de-duplication separately. I can approve once the open discussions were resolved, specifically [this](https://github.com/crossplane/docs/pull/566#discussion_r1367879599) one.

plumbis:
@turkenh I simplified the language there. I can expand on it after release.

Let me know if this work.

content/master/concepts/providers.md
```diff
@@ -89,28 +106,280 @@ crossplane-stable/crossplane \
...
+By default Crossplane sets the most recently installed package revision as 
+`Active`.
+
+Control the Provider upgrade behavior with a
+{{<hover label="revision" line="6">}}revisionActivationPolicy{{</hover>}}.
+
+The {{<hover label="revision" line="6">}}revisionActivationPolicy{{</hover>}} 
+options are:
+* `Automatic` - (**default**) Automatically use the latest Provider version
+  available in the cache.
```
turkenh, line 195:
```suggestion
* `Automatic` - (**default**) Automatically activate the most recently installed package revision.
```

content/master/concepts/providers.md
```diff
@@ -89,28 +106,280 @@ crossplane-stable/crossplane \
...
+`Active`.
+
+Control the Provider upgrade behavior with a
+{{<hover label="revision" line="6">}}revisionActivationPolicy{{</hover>}}.
+
+The {{<hover label="revision" line="6">}}revisionActivationPolicy{{</hover>}} 
+options are:
+* `Automatic` - (**default**) Automatically use the latest Provider version
+  available in the cache. 
+* `Manual` - Require the current Provider in use to be manually set.
```
turkenh, line 196:
```suggestion
* `Manual` - Do not activate revisions automatically.
```

content/master/concepts/providers.md
```diff
@@ -89,28 +106,280 @@ crossplane-stable/crossplane \
...
+```
+
+{{<hint "important" >}}
+Crossplane only upgrades a Provider if a newer version is in the package cache.   
+By default the Crossplane [`packagePullPolicy`](#provider-pull-policy) doesn't
+download new Provider versions, even if they're available.
+{{< /hint >}}
+
+Read the [Provider upgrade](#upgrade-a-provider) section for
+more information on the use of package revisions.
```
turkenh, line 218: This section is still inaccurate as detailed https://github.com/crossplane/docs/pull/566#discussion_r1367879599

> Crossplane only upgrades a Provider if a newer version is in the package cache.   
By default the Crossplane [`packagePullPolicy`](#provider-pull-policy) doesn't
download new Provider versions, even if they're available.

For example, the `revisionActivationPolicy` has nothing to do with the upgrades.

turkenh (approved):
> @turkenh I simplified the language there. I can expand on it after release.
> 
> Let me know if this work.

I left some comments for the technically inaccurate part.

PR #567, opened 2023-10-10
add support for matching pages when the page title changes

plumbis:
The dropdown menu with the version numbers builds links to the same page in another version if the Page title is the same. 

Because Hugo doesn't (directly) care about filenames and prefers focusing on file/page metadata it's much more difficult to match on file names than titles. 

This adds support for an optional frontMatter `altTitle` to allow authors to reference the older/original title. 

There are more flexible ways to do this but not without major build performance impact required to iterate over all pages multiple times.

### DIFF
--- a/content/contribute/contribute.md
+++ b/content/contribute/contribute.md
@@ -143,3 +143,18 @@ more.
 To hide a page from the left-hand navigation use `tocHidden: true` in the front
 matter of the page. The docs website skips pages with `tocHidden:true` when
 building the menu.
+
+### Changing page titles
+
+The version dropdown list that links the same page in different versions 
+together looks for pages with a matching title.
+
+If a page title changes use the front matter value `altTitle:` and a value of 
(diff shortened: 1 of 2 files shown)


content/contribute/contribute.md
```diff
@@ -142,4 +142,19 @@ more.
...
-building the menu.
\ No newline at end of file
+building the menu.
+
+### Changing page titles
+
+The version dropdown list that links the same page in different versions 
+together looks for pages with a matching title.
+
+If a page title changes use the front matter value `altTitle:` and a value of
```
jeanduplessis, line 152: Non-blocking suggestion: would `matchTitle` better convey the meaning of the field maybe?

jeanduplessis (approved):

content/contribute/contribute.md
plumbis: that is better! updated.

PR #576, opened 2023-10-19
feat: beta composition schema aware validation

phisco:
Now that https://github.com/crossplane/crossplane/pull/4814 was merged we should update the docs.

### DIFF
--- a/content/master/concepts/compositions.md
+++ b/content/master/concepts/compositions.md
@@ -1308,47 +1308,69 @@ parameters in the Composition.
 * Composition function `type` is `container`.
 * Composition function names are unique.
 
-### Resource schema validation
-<!-- vale write-good.TooWordy = NO -->
-Optionally, Crossplane can also validate the schema of the resources defined
-inside a Composition. This verifies that the resource `apiVersion` and `kinds`
-are valid. 
-<!-- vale write-good.TooWordy = YES -->
+### Composition schema aware validation
 
-Enable "schema validation" with the
-`--enable-composition-webhook-schema-validation` flag on the Crossplane pod.
-The [Crossplane Pods]({{<ref "./pods#edit-the-deployment">}}) page has
-more information on enabling Crossplane flags.
+On top of the preceding logical validation, Crossplane performs schema aware
+validation on the resources in a Composition, checking that `patches`,
+`readinessChecks` and `connectionDetails` are valid according to the involved
+resources' schemas, for example checking that the source/destination field of a
+patch is a valid field according to the source/destination resource's schema and
+that types match, if defined, taking into account transforms.
 
 {{<hint "note" >}}
-<!-- vale write-good.TooWordy = NO -->
-Schema validation only checks the `apiVersion` and `kind` are valid. Schema
-validation doesn't validate the fields of a specific resource.
-<!-- vale write-good.TooWordy = YES -->
+Composition schema aware validation is a beta feature, so Crossplane
+enables it by default. You can disable it by setting the
+`--enable-composition-webhook-schema-validation=false` flag on the Crossplane
+pod.
+The [Crossplane Pods]({{<ref "./pods#edit-the-deployment">}}) page has
+more information on enabling Crossplane flags.
 {{< /hint >}}
 
-The default validations are still checked with schema validation enabled. 
-#### Validation modes
+#### Schema aware validation modes
 
-Schema validation supports two modes:
-* `loose` (default) - Sends an warning for each schema error and installs the 
-  Composition if the default validations pass. 
-* `strict` - Send an error for every schema validation error and rejects the
-  Composition. 
+Logical validation errors always result in a rejection of the Composition.
+
+You can configure how Crossplane should handle both missing involved resources'
+schemas and actual schema aware validation errors.
+
+{{<hint "note" >}}
+In case of any involved resource schema missing, Crossplane skips schema aware
+validation altogether.
+{{< /hint >}}
+
+The following modes are available:
+
+| Mode     | Missing Schema | Schema aware Error | Logical Error |
+| -------- | -------------- |--------------------| ------------- |
+| `warn`   | Warning        | Warning            | Error         |
+| `loose`  | Warning        | Error              | Error         |
+| `strict` | Error          | Error              | Error         |
 
 Change the validation mode for a Composition with the
-{{<hover label="mode" line="5">}}crossplane.io/composition-validation-mode{{</hover>}} 
-annotation. 
+{{<hover label="mode" line="5">}}crossplane.io/composition-schema-aware-validation-mode{{</hover>}} 
+annotation.
 
-For example, to enable `strict` mode checking:
+If not specified, the default mode is `warn`.
+
+For example, to enable `loose` mode checking:
 
 ```yaml {copy-lines="none",label="mode"}
 apiVersion: apiextensions.crossplane.io/v1
 kind: Composition
 metadata:
   annotations:
-    crossplane.io/composition-validation-mode: strict
+    crossplane.io/composition-schema-aware-validation-mode: loose
   # Removed for brevity
 spec:
   # Removed for brevity
 ```
+
+{{<hint "important" >}}
+Modes also apply to Compositions defined by Configuration packages, look at
+Crossplane's logs for warnings in case of issues. The default is just to warn,
+so unless the Configuration explicitly set a different mode, Crossplane creates
+the Composition even in case of errors, but you should still check warnings as
+they can surface bugs in the Configuration or in the schema aware validation
+logic, and either the author of the Configuration or the Crossplane team should
+fix it, respectively.
+{{< /hint >}}
--- a/content/master/concepts/compositions.md
+++ b/content/master/concepts/compositions.md
@@ -1290,8 +1290,8 @@ Composition.
 
 ## Composition validation
 
-When creating a Composition Crossplane automatically validates specific
-parameters in the Composition.
+When creating a Composition, Crossplane automatically validates its integrity,
+checking that the Composition is well formed, for example:
 
 * All resources either use a `name` or don't. Compositions can't use both named
   and unnamed resources.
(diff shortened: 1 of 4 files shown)


content/master/concepts/compositions.md
```diff
@@ -1315,8 +1315,10 @@ inside a Composition. This verifies that the resource `apiVersion` and `kinds`
 are valid. 
 <!-- vale write-good.TooWordy = YES -->
 
-Enable "schema validation" with the
-`--enable-composition-webhook-schema-validation` flag on the Crossplane pod.
+Crossplane performs compositions "schema validation" by default. However, if you
+prefer to bypass this functionality, you can set the
+`--enable-composition-webhook-schema-validation=false` flag on the Crossplane
+pod.
```
plumbis, line 1321:
```suggestion
Crossplane performs compositions "schema validation" by default.  
Disable validation with the 
`--enable-composition-webhook-schema-validation=false` flag on the Crossplane
pod.
```

plumbis: Update to pass Vale

plumbis (changes requested):
I changed the branch to the WIP v1.14-release branch. Can you also copy these changes to the `/content/v1.14` folder in there as well?

phisco:
I noticed the description for the loose and strict modes is wrong, so I'll update them as part of this pr, probably adding the warn mode.

content/master/concepts/compositions.md
phisco: changed, have a look at it, I managed to make vale happy 🎉

content/master/concepts/compositions.md
```diff
@@ -1308,47 +1308,69 @@ parameters in the Composition.
...
+according to the source/destination resource's schema and that types match, if
+defined, taking into account transforms.
+
+{{<hint "note" >}}
+Composition schema aware validation is a beta feature, so Crossplane
+enables it by default. You can disable it by setting the
+`--enable-composition-webhook-schema-validation=false` flag on the Crossplane
+pod.
 The [Crossplane Pods]({{<ref "./pods#edit-the-deployment">}}) page has
 more information on enabling Crossplane flags.
```
plumbis, line 1326:
```suggestion
Composition schema aware validation is a beta feature. Crossplane enables
beta features by default. 

Disable schema aware validation by setting the
`--enable-composition-webhook-schema-validation=false` flag on the Crossplane
pod.  

The [Crossplane Pods]({{<ref "./pods#edit-the-deployment">}}) page has
more information on enabling Crossplane flags.
```

content/master/concepts/compositions.md
```diff
@@ -1308,47 +1308,69 @@ parameters in the Composition.
...
+{{< /hint >}}
+
+#### Schema aware validation modes
+
+Integrity validation errors always result in a rejection of the Composition.
+
+You can configure how Crossplane should handle both missing involved resources'
+schemas and actual schema aware validation errors.
```
plumbis, line 1334:
```suggestion
<!-- vale alex.ProfanityUnlikely = NO -->
Integrity validation errors reject a Composition.
<!-- vale alex.ProfanityUnlikely = Yes  -->

Set the schema validation mode to configure how Crossplane handles both missing 
resource schemas and schema validation errors.
```

The word "reject" is a vale error and shouldn't be but I can't make the change in review so I just disable the check.

content/master/concepts/compositions.md
```diff
@@ -1308,47 +1308,69 @@ parameters in the Composition.
...
 
 {{<hint "note" >}}
-<!-- vale write-good.TooWordy = NO -->
-Schema validation only checks the `apiVersion` and `kind` are valid. Schema
-validation doesn't validate the fields of a specific resource.
-<!-- vale write-good.TooWordy = YES -->
+In case of any involved resource schema missing, Crossplane skips schema aware
+validation altogether.
```
plumbis, line 1338: This is based on the validation mode according to the table below, right?

content/master/concepts/compositions.md
```diff
@@ -1308,47 +1308,69 @@ parameters in the Composition.
...
-Schema validation supports two modes:
-* `loose` (default) - Sends an warning for each schema error and installs the 
-  Composition if the default validations pass. 
-* `strict` - Send an error for every schema validation error and rejects the
-  Composition. 
+| Mode     | Missing Schema | Schema Aware Error | Integrity Error |
+| -------- | -------------- |--------------------|-----------------|
+| `warn`   | Warning        | Warning            | Error           |
+| `loose`  | Warning        | Error              | Error           |
+| `strict` | Error          | Error              | Error           |
```
plumbis, line 1347:
```suggestion
{{< table "table table-sm table-striped" >}}
| Mode     | Missing Schema | Schema Aware Error | Integrity Error |
| -------- | -------------- |--------------------|-----------------|
| `warn`   | Warning        | Warning            | Error           |
| `loose`  | Warning        | Error              | Error           |
| `strict` | Error          | Error              | Error           |
{{< /table >}}
```

table styling.

content/master/concepts/compositions.md
```diff
@@ -1308,47 +1308,69 @@ parameters in the Composition.
...
 Change the validation mode for a Composition with the
-{{<hover label="mode" line="5">}}crossplane.io/composition-validation-mode{{</hover>}} 
-annotation. 
+{{<hover label="mode" line="5">}}crossplane.io/composition-schema-aware-validation-mode{{</hover>}} 
+annotation.
+
+If not specified, the default mode is `warn`.
 
-For example, to enable `strict` mode checking:
+For example, to enable `loose` mode checking:
```
plumbis, line 1355:
```suggestion
For example, to enable `loose` mode checking set the annotation value to 
{{<hover label="mode" line="5">}}loose{{</hover>}}.
```

content/master/concepts/compositions.md
```diff
@@ -1308,47 +1308,69 @@ parameters in the Composition.
...
-    crossplane.io/composition-validation-mode: strict
+    crossplane.io/composition-schema-aware-validation-mode: loose
   # Removed for brevity
 spec:
   # Removed for brevity
 ```
+
+{{<hint "important" >}}
+Modes also apply to Compositions defined by Configuration packages, look at
+Crossplane's logs for warnings in case of issues. The default is just to warn,
```
plumbis, line 1370: Do you mean `kubectl logs` or the messages in `kubectl describe` of something?

plumbis (changes requested):

content/master/concepts/compositions.md
phisco: I get the intention, but I feel it's slightly wrong saying that errors reject the composition... it's crossplane rejecting it because of errors. I'd rather be correct than have vale somewhat happy 😬

content/master/concepts/compositions.md
phisco: That's for all modes, schema aware validation is not performed unless we have all required schemas available. Depending on the mode this could result in a warning or error.

content/master/concepts/compositions.md
phisco: Logs logs, as in kubectl logs, describe ones are events, we log warnings to stdout if we get any back for one of the calls Crossplane does. E.g. the infamous "ControllerConfig has been deprecated" which showed up in the logs

content/master/concepts/compositions.md
plumbis: I'm trying to get away from "result in a rejection". How about 

"Crossplane rejects Compositions with validation errors"

?

The vale error from an imported library that says the word "reject" may be profane, which it isn't in our writing/use.

phisco:
```suggestion
Crossplane always rejects Compositions in case of integrity errors.

Set the schema validation mode to configure how Crossplane handles both missing 
resource schemas and schema aware validation errors.
```

content/master/concepts/compositions.md
```diff
@@ -1308,47 +1308,74 @@ parameters in the Composition.
...
+You can configure how Crossplane should handle both missing involved resources'
+schemas and actual schema aware validation errors.
 
 {{<hint "note" >}}
-<!-- vale write-good.TooWordy = NO -->
-Schema validation only checks the `apiVersion` and `kind` are valid. Schema
-validation doesn't validate the fields of a specific resource.
-<!-- vale write-good.TooWordy = YES -->
+In case of any involved resource schema missing, Crossplane skips schema aware
+validation altogether.
```
plumbis, line 1340:
```suggestion
If a resource is missing a schema, Crossplane skips schema aware and integrity 
checking.

Crossplane only provides "missing schema" warnings or errors.
```

content/master/concepts/compositions.md
```diff
@@ -1308,47 +1308,74 @@ parameters in the Composition.
...
 ```
+
+{{<hint "important" >}}
+Modes also apply to Compositions defined by Configuration packages, look at
+Crossplane's logs for warnings in case of issues. The default is just to warn,
+so unless the Configuration explicitly set a different mode, Crossplane creates
+the Composition even in case of errors, but you should still check warnings as
+they can surface bugs in the Configuration or in the schema aware validation
+logic, and either the author of the Configuration or the Crossplane team should
+fix it, respectively.
```
plumbis, line 1380:
```suggestion
Validation modes also apply to Compositions defined by Configuration packages.

View the Crossplane pod's logs with `kubectl logs` to see warnings or errors. 

By default Composition validation only issues warnings unless the 
Configuration explicitly sets a different mode. 

Crossplane creates the Composition from a Configuration package even if there 
are errors.
```

plumbis (changes requested):

content/master/concepts/compositions.md
phisco: Nope, just schema aware validation is skipped, not integrity checks. See the full logic below 👇 read "valid per se" as integrity check
 image

content/master/concepts/compositions.md
plumbis: LGTM

content/master/concepts/compositions.md
plumbis: updated

content/master/concepts/compositions.md
phisco: Splitting things up this way make it slightly incorrect.

You have to view logs to see warnings only for compositions applied by Crossplane itself, splitting it up makes it look like for all warnings you have to look at the logs, which is false.

The default mode is to only emit warnings for schema aware validation errors, not for all kinds of validation, and unless the Composition defined in the Configuration set a different mode, the default will apply. So, compositions will be successfully applied, but warnings may be printed in the logs. While from your sentence it looks like when creating compositions for packages we always ignore errors.

content/master/concepts/compositions.md
phisco: Not sure I get what you mean with the last sentence in the updated version.

The first sentence is an ok rewriting of the original version, although a bit oversimplified. I would be happier keeping the original version tbh

plumbis: The last sentence is to be explicit about what is still being done. It's implied when you say. "skips schema aware" but this adds in meaning of "it still runs the other two checks"

phisco: `Crossplane only provides "missing schema" and "integrity checking" warnings or errors.` 

The wording and the fact makes it sound like if Crossplane always only provided these two things.

A more correct version would be:
```suggestion
If a resource schema is missing, Crossplane skips schema aware validation altogether, but will still return an error in case of any integrity error and either a warning or an error depending on the mode selected for the missing schemas.
```
But I get Vale would probably not like it and it's a long sentence, which are the reasons why I tried avoiding getting in so much details.

plumbis: I like that. Slight edit for our angry overlord Vale:

> If a resource schema is missing, Crossplane skips schema aware validation 
> but still returns an error for integrity errors and a warning or an error 
> for the missing schemas.

content/master/concepts/compositions.md
plumbis: Updated copy.

phisco: `Unlike applying a Composition, Crossplane doesn’t reject errors in a Composition inside of a Configuration.` is not correct, Crossplane doesn't distinguish between Compositions coming from a user and the ones coming from a Configuration, they all hit the validating webhook and are all validated according to the same above logic, so if the mode set is something else than `warn`, a Composition coming from a Configuration could result in an error.

plumbis: Sorry, I feel like I'm going in a circle here, I apologize.

So if I apply a configuration that has a Composition with `crossplane.io/composition-schema-aware-validation-mode: strict` what happens?

Does it apply the Configuration, reject the Composition and the package doesn't become `HEALTHY`?

phisco: Yes, exactly.

plumbis: one more rewrite. Let me know if this works.

phisco: Not really 😅
A more correct, less vale-pleasing version would be:
```
Crossplane applies schema aware validation to Compositions defined by Configuration packages as well.

Depending on the validation mode set on the Composition, in case of schema aware validation issues,
Crossplane will log any warning to  its pod stdout or, if the Composition is fully rejected, mark the parent Configuration as unhealthy and show the error as condition's status.
```

However what I cared about is letting people know to look at logs to see warning and report them back if needed, they will notice if a composition is unhealthy no need to tell that.

PR #577, opened 2023-10-23
fix links for moved platform ref files

plumbis:
The AWS platform ref was recently refactored moving the location of files referenced in the docs.

This updates the links to point to the new URLs


ytsarev:
it is safer to base docs on some stable tag instead of `main` branch imho

PR #584, opened 2023-10-25
CLI docs

plumbis:
Includes CLI docs covering everything **except** `beta trace` and `beta xpkg init`.


content/v1.14/cli/_index.md
```diff
@@ -0,0 +1,47 @@
...
+[The script](https://raw.githubusercontent.com/crossplane/crossplane/master/install.sh)
+detects your CPU architecture and downloads the latest stable release.
+
+Manually download a binary from the Crossplane releases repository at 
+https://releases.crossplane.io/stable/current/bin
+
+{{<hint "important" >}}
+<!-- vale write-good.Passive = NO -->
+The CLI is named `crank` in the release repository. Download this file. 
+<!-- vale write-good.Passive = YES -->
```
negz, line 42: Could we get away without this, and have everyone use the script? It would help avoid explaining the crank thing.

content/v1.14/cli/_index.md
```diff
@@ -0,0 +1,47 @@
...
+
+The Crossplane CLI is independent of Kubernetes and doesn't require a Crossplane
+pod.
+{{< /hint >}} 
+
+To download the latest version for your CPU architecture with the Crossplane
+install script.
+
+```shell
+curl -sL "https://raw.githubusercontent.com/crossplane/crossplane/master/install.sh" | sh
```
negz, line 30: Should we document the `XP_VERSION` and `XP_CHANNEL` variables that control which version of the CLI is downloaded and installed?

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
+---
+weight: 50
+title: Command Reference
+description: "Command reference for the Crossplane CLI"
+---
+
+<!-- vale Google.Headings = NO -->
+The `crossplane` CLI provides utilities to make using Crossplane easier.
```
negz, line 8: Does "easier" go against the style/contributing guide here?

If so, perhaps "more convenient"?

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+and publishing of Crossplane packages to a Crossplane package registry. 
+
+### xpkg build
+
+Using `crossplane xpkg build` provides automation and simplification to build 
+Crossplane packages.  
+
+The Crossplane CLI creates a 
+single YAML file from a directory of YAML files and combines them as an 
+[OCI container image](https://opencontainers.org/).
```
negz, line 37: Does it matter to the user/reader at this level that it creates a single YAML file? That seems to me like an internal implementation detail. It's covered by linking to the spec for those who really want to know how it works. Saying it combines them into a package as you do below feels like the right level of detail to me.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+[function]({{<ref "../concepts/composition-functions">}}) and 
+[provider]({{<ref "../concepts/providers" >}}) package types. 
+
+
+#### Flags
+{{< table "table table-sm table-striped">}}
+| Short flag   | Long flag                            | Description                    |
+| ------------ | -------------                        | ------------------------------ |
+|              | `--embed-runtime-image-name=NAME`    |  The image name and tag of a container to include in the package. Only for provider and function packages. |
+|              | `--embed-runtime-image-tarball=PATH` |  The filename of a container to include in the package. Only for provider and function packages.                              |
```
negz, line 53: Should we explain to the user why we're including these containers (nit: technically it's an image - containers are running images) in the package?

We include them so that the resulting package can be used both by the package manager to install the package, and by Kubernetes to run the package's "runtime". The runtime could be a provider pod, or a function pod.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+
+Include YAML files demonstrating how to use the package with `--examples-root`. 
+
+[Upbound Marketplace](https://marketplace.upbound.io/) uses files included with 
+`--examples-root` as documentation for published packages.
+
+#### Include a runtime image
+
+Functions and Providers require YAML files describing their dependencies and 
+settings as well as a container image for the function or provider controller.
```
negz, line 95: This could be read as if a function has a controller. It doesn't - that's why we say runtime rather than controller in the flag.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+
+Using `--embed-runtime-image-name` downloads a specified container image and 
+includes the image inside the function or provider package. Use the
```
negz, line 98: Looks like this trails off.

Also, I'm not sure it does actually download the image. I think it connects to the Docker daemon and looks for it there, so you might need to explicitly pull it first.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+
+Store more inactive copies of a package with `--revision-history-limit`. 
+
+Read more about 
+[package revisions]({{< ref "../concepts/packages#configuration-revisions" >}}) 
+in the package documentation. 
+
+### xpkg login
+
+Use `xpkg login` to authenticate to the
```
negz, line 178:
```suggestion
Use `xpkg login` to authenticate to xpkg.upbound.io, the 
```

I think it's worth explicitly including the domain name. It's going to appear in a lot of places (e.g. as a prefix in any package, like xpkg.upbound.io/crossplane/contrib/foo:v1.0.0). I suspect a lot of Crossplane users will end up being exposed to xpkg.upbound.io as a place where packages are pulled/pushed vs "the Upbound Marketplace".

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+By default the `crossplane xpkg install` command uses the Kubernetes 
+configuration defined in `~/.kube/config`.  
+
+Define a custom Kubernetes configuration file location with the environmental 
+variable `KUBECONFIG`.
+
+Specify the package kind, package file and optionally a name to give the package 
+inside Crossplane.
+
+`crossplane xpkg install <package-kind> <package-file> [<optional-name>`
```
negz, line 118: I would not call this a package file. It's not a file, and I think this oversimplification will be confusing to anyone familiar with OCI images and registries. I struggled with naming for this too, and landed on just calling it "a package".

Other commands take flags/arguments like `--package-file` and in those cases they accept an actual `.xpkg` file.

"package tag" or "package reference" could also be good here. The former is more accurate (technically you could give a digest, not a tag, like `crossplane-contrib/provider-foo@sha256:...`) but the latter is probably good enough to be meaningful to most people.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+
+The package file must be an organization, image and tag on the 
+[Upbound Marketplace](https://marketplace.upbound.io/).
```
negz, line 121: This is only the default behaviour. You could specify `index.docker.io/crossplane-contrib/foo:v1.0.0` to install from a different registry.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+`~/.crossplane/config.json` file, but doesn't delete the configuration file.
+{{< /hint >}}
+
+### xpkg push
+
+Push a Crossplane package file to a package registry. 
+
+The Crossplane CLI pushes images to the 
+[Upbound Marketplace](https://marketplace.upbound.io/) at `xpkg.upbound.io` by 
+default.
```
negz, line 248: Should we show how to push to another registry?

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+By default the `crossplane xpkg update` command uses the Kubernetes 
+configuration defined in `~/.kube/config`.  
+
+Define a custom Kubernetes configuration file location with the environmental 
+variable `KUBECONFIG`.
+
+Specify the package kind, package file and optionally the name of the package 
+already installed in Crossplane.
+
+`crossplane xpkg update <package-kind> <package-file> [<optional-name>`
```
negz, line 292: Same note here WRT `<package-file>`

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+releases.
+
+### beta render 
+
+The `crossplane beta render` command previews the output of a 
+[composite resource]({{<ref "../concepts/composite-resources">}}) after applying 
+any 
+[composition patches]({{<ref "../concepts/compositions#changing-resource-fields">}}) 
+and [composition function]({{<ref "../concepts/composition-functions">}}) 
+changes.
```
negz, line 318: It doesn't apply patches. It completely ignores the `resources` field. `crossplane beta render` only works with composition functions - i.e. a Composition with `mode: Pipeline`.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+
+The command doesn't require Kubernetes or access to a Crossplane pod. The CLI
+processes the inputs locally to generate the rendered output.
```
negz, line 321: It does require Docker (or technically, anything that exposes a Docker-compatible socket and API). Probably worth pointing that out.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+```
+{{< /expand >}}
+
+#### Flags
+
+{{< table "table table-sm table-striped">}}
+| Short flag   | Long flag                             | Description                                           |
+| ------------ | -------------                         | ------------------------------                        |
+|              | `--context-files=<key>=<file>;<key>=<file>`    | A semicolon separated list of files to load for function "contexts." |
+|              | `--context-values=<key>=<value>;<key>=<value>` | A semicolon separated list of key-value pairs to load for function "contexts."                                                    |
```
negz, line 369: I made these comma separated as part of the CLI docs PR on the Crossplane end.

content/v1.14/cli/command-reference.md
negz: I also like how this is framed in the provider docs:

> Install a Provider with a Crossplane Provider object setting the spec.package value to **the location of the provider package**.

content/v1.14/cli/_index.md
plumbis: I would but since some people don't like `curl | bash` they would look at the repo and it's not obvious that `crank` is the binary, not `crossplane`. This exact scenario tripped me up.

content/v1.14/cli/command-reference.md
plumbis: No violation. The rule is about saying something is easy (i.e., if you don't understand you're dumb). In this case we're talking about a relative change.

content/v1.14/cli/command-reference.md
plumbis: The reason I call this out is that if you've read the spec it talks about a single yaml file. I'm calling out "hey, this makes things better because you don't have to build the single massive yaml, we'll do it for you"

content/v1.14/cli/command-reference.md
plumbis: Is there another word besides "runtime" I could use? Would "container image for the function or provider binary" work?

content/v1.14/cli/command-reference.md
plumbis: I couldn't tell from the `daemon.image` docs, can you confirm that it won't download a local package?

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,416 @@
...
+includes the image inside the function or provider package. Use the 
+
+The `--embed-runtime-image-tarball` flag includes a local OCI image tarball 
+inside the function or provider package.
+
+
+### xpkg install
+
+Download and install Crossplane packages with  `crossplane xpkg install`.
+Crossplane packages.
```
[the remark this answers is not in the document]
plumbis: I updated the note in the `_index.md`

> {{<hint "note" >}}
> Install the Crossplane CLI on a user's computer. 
> 
> Most Crossplane CLI commands are independent of Kubernetes and 
> don't require access to a Crossplane pod.
> {{< /hint >}} 
 
For this command I updated the referenced sentence:

> Download and install packages into Crossplane with  `crossplane xpkg install`.

content/v1.14/cli/command-reference.md
plumbis: added the domain anywhere there's a Marketplace reference and the domain wasn't there.

content/v1.14/cli/command-reference.md
plumbis: Updated to

`crossplane xpkg install <package-kind> <registry package name and tag> [<optional-name>`

content/v1.14/cli/command-reference.md
plumbis: Testing this it looks like the default is always index.docker.io? I thought it was changed to be xpkg.upbound.io?

content/v1.14/cli/command-reference.md
plumbis: Sure, how? Can I just provide a full URI for the package argument?

content/v1.14/cli/command-reference.md
plumbis: Updated:

> The `crossplane beta render` command previews the output of a 
> [composite resource]({{<ref "../concepts/composite-resources">}}) after applying 
> any [composition functions]({{<ref "../concepts/composition-functions">}}).
> 
> {{< hint "important" >}}
> The `crossplane beta render` command doesn't apply 
> [composition patches]({{<ref "../concepts/compositions#changing-resource-fields">}}.
> {{< /hint >}}

content/v1.14/cli/command-reference.md
negz: I don't think we should be assuming users of `crossplane xpkg` build are familiar with the spec. That feels a bit like assuming `docker build` users have read the OCI specification and are thinking about what it does through that lens.

content/v1.14/cli/command-reference.md
negz: No, I think I wrote this comment before I noticed that section and forgot to remove it. Thanks!

content/v1.14/cli/command-reference.md
negz: I thought so too.

content/v1.14/cli/command-reference.md
negz: Yeah exactly - like `index.docker.io/crossplane-contrib/...`

content/v1.14/cli/command-reference.md
plumbis: Maybe I should open an issue. 

```shell
$ ./crank xpkg install provider upbound/provider-aws-s3/v0.42.0                                                                                           
provider/upbound-provider-aws-s3-v0-42-0 created

$ k describe provider/upbound-provider-aws-s3-v0-42-0 created
#--- snip ---#
  Warning  UnpackPackage  2s (x3 over 4s)  packages/provider.pkg.crossplane.io  cannot unpack package: failed to fetch package digest from remote: failed to fetch package descriptor with a GET request after a previous HEAD request failure: HEAD https://index.docker.io/v2/upbound/provider-aws-s3/v0.42.0/manifests/latest: unexpected status code 401 Unauthorized (HEAD responses have no body, use GET for details): GET https://index.docker.io/v2/upbound/provider-aws-s3/v0.42.0/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:upbound/provider-aws-s3/v0.42.0 Type:repository]]
```

content/v1.14/cli/command-reference.md
plumbis: updated.

content/v1.14/cli/command-reference.md
negz: It would. I'd like to get folks associating "runtime" as meaning that in this context, so I'd prefer to keep that word in there somewhere to make the connection.

plumbis: Good enough reason for me. I'll change it.

content/v1.14/cli/_index.md
phisco: Uploading all binaries to the gh releases too as many other tools do could allow us not to have to point people at the s3 bucket.

content/v1.14/cli/_index.md
```diff
@@ -0,0 +1,60 @@
...
+description: "Documentation for the Crossplane command-line interface"
+---
+
+The Crossplane CLI helps simplify some development and administration aspects of
+Crossplane.
+
+The Crossplane CLI includes:
+* tools to build, install, update and push Crossplane Packages
+* standalone Composition Function testing and rendering without the need to access a Kubernetes cluster running Crossplane
+* visual dependency mapping across Compositions, Composite Resources and Managed Resources
```
jbw976, line 13: maybe mention troubleshooting here, as that is a key focus of this feature area

content/v1.14/cli/_index.md
```diff
@@ -0,0 +1,60 @@
...
+install script.
+
+```shell
+curl -sL "https://raw.githubusercontent.com/crossplane/crossplane/master/install.sh" | sh
+```
+
+[The script](https://raw.githubusercontent.com/crossplane/crossplane/master/install.sh)
+detects your CPU architecture and downloads the latest stable release.
+
+Manually download a binary from the Crossplane releases repository at
```
jbw976, line 36: this seems like an alternate option from running the script, but it's reading like the next step in the flow. do you want to make it more clear that it's either run the script OR manually download the binary, not both?

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+By default the `crossplane xpkg install` command uses the Kubernetes 
+configuration defined in `~/.kube/config`.  
+
+Define a custom Kubernetes configuration file location with the environmental 
+variable `KUBECONFIG`.
+
+Specify the package kind, package file and optionally a name to give the package 
+inside Crossplane.
+
+`crossplane xpkg install <package-kind> <registry package name and tag> [<optional-name>`
```
jbw976, line 116:
```suggestion
`crossplane xpkg install <package-kind> <registry package name and tag> [<optional-name>]`
```

jbw976, line 116: are we saying anywhere what the valid options are for `<package-kind>`? I didn't see it myself. We give `provider` in the example below, but I'm not sure we define what other kinds are allowed.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+
+The package file must be an organization, image and tag on the `xpkg.upbound.io`
```
jbw976, line 118: is this true? i assumed you could install from any registry, e.g. `index.docker.io/acme/package:v0.1.0`

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+| `-r`         | `--revision-history-limit=<number of revisions>` | Set the `revisionHistoryLimit`. Defaults to `1`.                                                |
+| `-w`         | `--wait=<number of seconds>`                     | Number of seconds to wait for a package to install.                                             |
+
+{{< /table >}}
+
+#### Wait for package install
+
+When installing a package the `crossplane xpkg install` command doesn't wait for
+the package to download and install. View any download or installation problems 
+by inspecting the `configuration` with `kubectl describe configuration`.
```
jbw976, line 142: this will only be valid if the package is a configuration i think - if it's a provider or function then this command wouldn't find it.

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+
+Use an Upbound user token instead of a username and password with `--token` or 
+the `UP_TOKEN` environmental variable. 
+
+{{< hint "important" >}}
+The `--token` or `UP_TOKEN` environmental variables take precedence over a 
+username and password.
+{{< /hint >}}
+
+Using `-` as the input for `--password` or `--token` reads the input from stdin.
```
jbw976, line 210: we could also consider saying what happens if you pass no arguments - you get prompted for username and password:
```
❯ crossplane xpkg login
Username: jbw976
Password:
Login successful.
```

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+
+#### Flags
+
+{{< table "table table-sm table-striped">}}
+| Short flag   | Long flag              | Description                                   |
+| ------------ | -------------          | ------------------------------                |
+| `-f`         | `--package-files=PATH` | A comma-separated list of xpkg files to push. |
+{{< /table >}}
+
+For example, to push a package named `my-configuration` use
```
jbw976, line 274: does this example add more value that hasn't already been shown? everything this does is covered by the previous example.

also, would it work OK? there isn't a registry, org, repo, or tag specified 🤔

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+By default the `crossplane xpkg update` command uses the Kubernetes 
+configuration defined in `~/.kube/config`.  
+
+Define a custom Kubernetes configuration file location with the environmental 
+variable `KUBECONFIG`.
+
+Specify the package kind, package file and optionally the name of the package 
+already installed in Crossplane.
+
+`crossplane xpkg update <package-kind> <registry package name and tag> [<optional-name>`
```
jbw976, line 290:
```suggestion
`crossplane xpkg update <package-kind> <registry package name and tag> [<optional-name>]`
```

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+
+The package file must be an organization, image and tag on the `xpkg.upbound.io`
+registry on [Upbound Marketplace](https://marketplace.upbound.io/).
+
+For example, to update to version 0.42.0 of the 
+[AWS S3 provider](https://marketplace.upbound.io/providers/upbound/provider-aws-s3/v0.42.0):
+
+`crossplane xpkg update provider xpkg.upbound.io/upbound/provider-aws-s3:v0.42.0`
```
jbw976, line 298: interesting, i didn't know we had this functionality 😂 - does it look for existing installed packages from the same registry, org, and repo then tries to upgrade what it finds to the new version here? 🤔

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+
+### beta render 
+
+The `crossplane beta render` command previews the output of a 
+[composite resource]({{<ref "../concepts/composite-resources">}}) after applying 
+any [composition functions]({{<ref "../concepts/composition-functions">}}).
+
+{{< hint "important" >}}
+The `crossplane beta render` command doesn't apply 
+[composition patches]({{<ref "../concepts/compositions#changing-resource-fields">}}).
```
jbw976, line 317: would it make more sense to say that it ONLY executes functions? or clarify that these are traditional patch and transforms (you can still use the function based patch and transforms...) - up to you to judge if this would just be more confusing 😂

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+{{< /hint >}}
+
+The `crossplane beta render` command connects to the locally running Docker 
+Engine to pull and run composition functions. 
+
+{{<hint "important">}} 
+Running `crossplane beta render` requires [Docker](https://www.docker.com/).
```
jbw976, line 324: it does require Docker specifically, not any alternate package runtimes/clients?

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+
+#### Include function results
+
+If a function produces Kubernetes events with statuses use the 
+`--include-function-results` to print them along with the managed resource 
+outputs. 
+
+#### Mock existing resources
+
+Provide mocked, or artificial data representing a managed resource with
```
jbw976, line 398: would it help to include the words "existing" and/or "observed" in this sentence to further describe what these mocked resources represent?

content/v1.14/cli/command-reference.md
```diff
@@ -0,0 +1,595 @@
...
+### beta xpkg init
+
+The `crossplane beta xpkg init` command populates the current directory with 
+files to build a package. 
+
+Provide a name to use for the package and the package template to start from 
+with the command  
+`crossplane beta xpkg init <name> <template>`
+
+The `<name>` input isn't used.
```
jbw976, line 578: should this be explained further that it isn't currently used, but is reserved for future usage?

jbw976 (commented):
This is a fantastic effort @plumbis to get our CLI docs going - I really like it and appreciate it!! dropped a few comments that may help with clarifying some of the content here, feel free to push back on anything you think doesn't actually improve things 😇

content/v1.14/cli/command-reference.md
negz: Yeah, confirmed you have to manually pull. For example:

```shell
$ crossplane xpkg build -f package --embed-runtime-image=alpine
crossplane: error: failed to get runtime base image options: failed to pull runtime image: Error response from daemon: No such image: alpine:latest

$ docker pull alpine
Using default tag: latest
latest: Pulling from library/alpine
579b34f0a95b: Pull complete
Digest: sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978
Status: Downloaded newer image for alpine:latest
docker.io/library/alpine:latest

$ crossplane xpkg build -f package --embed-runtime-image=alpine
$ echo $?
0
```

content/v1.14/cli/command-reference.md
negz: Oh, my bad, install and update do default to index.docker.io.

This is because they're really just asking the package manager to install the thing, and the package manager defaults to index.docker.io. We considered changing that with this release so everything is consistent and not-confusing but we thought the change would be too risky/disruptive.

content/v1.14/cli/command-reference.md
negz: I think we should leave this as is. It uses the Docker API to talk to Docker. Some other runtimes such as Podman emulate the Docker API. I feel like if someone is using an alternative runtime they'll know whether or not they should expect it to work as a stand-in for Docker.

content/v1.14/cli/command-reference.md
negz: No, it just defaults the name from the package the same way `xpkg install` does if you omit the name.

content/v1.14/cli/command-reference.md
phisco, line 1: I really appreciate your effort here @plumbis, but are we really sure we are going to be able to maintain this? Looking at other examples such as [istioctl](https://istio.io/latest/docs/reference/commands/istioctl/) and [kubectl](https://kubernetes.io/docs/reference/kubectl/) which then points to https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands. It feels like the common pattern is to provide an high level overview of common flags and functionalities, showcase common scenarios and then point to the generated documentation directly for all the nitty gritty details, which would make sense to me as a user.

utils/vale/styles/Crossplane/spelling-exceptions.txt
```diff
@@ -2,6 +2,7 @@
 /tabs
 call-outs
 ClusterRoles`
+comma-separated
```
phisco, line 5: duplicated

content/v1.14/cli/_index.md
plumbis: Put the manual steps in a collapse section.

content/v1.14/cli/command-reference.md
plumbis: Nope. Text updated.

content/v1.14/cli/command-reference.md
plumbis: If you give the command an invalid Provider package URL it returns immediately and you see 404 errors when you describe the provider.

Is this what you mean or something else?

content/v1.14/cli/command-reference.md
plumbis: I think the concern is valid but if we have enough commands that it's a problem, that's a good problem to have.

For now I'd like to keep it manually updated.

PR #590, opened 2023-10-27
Update providers to not say they install CRDs

### LINKED ISSUE #586: [Web Bug] - Provider pods don't install a provider's APIs
URL: https://docs.crossplane.io/v1.13/concepts/providers/

> Installing a provider creates a Provider pod that’s responsible for installing the Provider’s APIs into the Kubernetes cluster.

If I was a Kubernetes-aware user reading this I would assume the Provider pod created its own CRDs. In reality, Crossplane installs the CRDs "installs the Provider's APIs". The provider pod reconciles them.

plumbis:
Resolves #586.

Changes the wording about what happens when a provider is installed.

### DIFF
--- a/content/master/concepts/providers.md
+++ b/content/master/concepts/providers.md
@@ -40,10 +40,11 @@ for more information.
 
 ## Install a Provider
 
-Installing a provider creates a Provider pod that's responsible for installing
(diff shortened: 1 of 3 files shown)


plumbis:
Vale errors are unrelated to content changed.

negz (approved):
Thanks @plumbis!

PR #628, opened 2023-11-25
Improve Offline Provider Installation Documentation

arielsepton:


plumbis:
It looks like you may have accidentally pulled in a lot of other commits for this PR. Can you rebase or open a new PR that only includes your intended changes?

arielsepton:
Thanks for catching that @plumbis! I've fixed the issue

content/master/concepts/providers.md
```diff
@@ -113,28 +113,112 @@ Crossplane installs packages from a local package cache. By
...
+Follow this example to configure Crossplane using a PersistentVolumeClaim (PVC) for a storage location containing the Provider image.
+
+```shell
+CACHE_PATH="localpackage-cache"
+
+# Extract package and move it to the cache path
+up xpkg xp-extract xpkg.upbound.io/upbound/provider-aws:v0.44.0
+mv out.gz "${CACHE_PATH}/provider-aws.gz"
+chmod 644 "${CACHE_PATH}/provider-aws.gz"
+docker tag xpkg.upbound.io/upbound/provider-aws:v0.44.0 provider-aws
```
plumbis, line 152: `up xpkg xp-extract` is a hidden command (the equiv is an alpha command) is it possible to use the `crossplane` CLI to do this? 

Does the package have to be extracted?

plumbis: _bump_

arielsepton: Yes, the package does need to be extracted.
For Crossplane offline installation, with ‘–set packageCache.pvc=’ option, The extracted packages must be placed in the directory where the PVC is mounted for Crossplane’s inner registry to recognize them.

Regarding the Crossplane CLI command you mentioned, Could you please provide me with the relevant documentation or details about the command? I’ll review it to see if we can replace the up xpkg xp-extract command with a Crossplane CLI equivalent.

plumbis: Documentation for the Crossplane CLI is [here](https://docs.crossplane.io/v1.14/cli/command-reference/). 

I'm not sure if `extract` is an option (even as a hidden command) but can you check to see if the options match what you're doing here? I'd prefer to have the Crossplane CLI in this doc but if Crossplane doesn't provide it and we have to reference the `up` CLI that's okay.

PR #649, opened 2023-12-04
Fix issue with API folder on /latest

plumbis:
The API page doesn't render in /latest. /latest is the entry point for search engines.

The reason is that the existing code looks for a version number that doesn't exist. This changes the logic to use the current directory instead of the version parameter. 

Verified by creating a /latest content dir and removing the version number.


jbw976 (approved):
Seems reasonable, but leaning heavily on your testing notes for verification here 😊

PR #651, opened 2023-12-04
Correct crossplane-browse cluster role description

erikgb:
Update docs to be better aligned with the granted permissions. Tried to write something meaningful based on: 

https://github.com/crossplane/crossplane/blob/90abed63db7b77713ee3b262dc53422d565260b5/cluster/charts/crossplane/templates/rbac-manager-managed-clusterroles.yaml#L183C45-L185C18

but feel free to suggest improvements.

Also correcting some non-working kubectl-commands. At least in my cluster.


plumbis (approved):
This is great @erikgb, thanks!!

PR #652, opened 2023-12-05
doc: fix inconsistency with toFieldPath and the applied value

girdharshubham:


plumbis:
Good catch, thank you! Would mind adding this same fix to the file in `content/master/concepts/patch-and-transform.md`?

girdharshubham:
@plumbis Done.
Thanks for reviewing.!

plumbis (approved):
Woo! Thank you 🎉

Issue #643, opened 2023-12-01
Link to raw YAML in API docs


plumbis:
### What's Missing?

The API docs should have a link to the original YAML file.

PR #656, opened 2023-12-12
Add download YAML and back to top buttons

### LINKED ISSUE #644: Collapse API docs
### What's Missing?
The API docs should have a way to collapse back to the parent when you're at the bottom of a large output

plumbis:
Adds a 'back to top' button at the bottom of a CRD and a 'download YAML' link. Resolves #644 and #643

Also upgrades bootstrap to v5.3 to get new functionality required for this change.


jeanduplessis (approved):
@plumbis I assume we deliberately don't have this functionality on mobile - which I am okay with.

plumbis:
That's correct. Lots of data here makes it hard to add more features to mobile view so I assumed most wouldn't want to view the raw YAML (or they can switch to desktop view)

PR #659, opened 2023-12-14
fix(concepts): correct xpkg build flags

morey-tech:
Updated the build the package examples to use the correct flags for `xpkg build`.
- `--name` is now `--output`
- `-f` is now `--package-root`

I've updated the section in the `master` and `v1.14` docs.


jeanduplessis (approved):
Thank you for the fix @morey-tech

PR #668, opened 2024-01-04
Dark Mode Table Legibility 

tr0njavolta:
Addresses https://github.com/upbound/docs/issues/285


plumbis (approved):
Thanks @tr0njavolta ! this fixes one issue but it made me find a few more with tables in dark mode. I opened #669 to investigate and fix the others.

PR #677, opened 2024-01-17
Argocd health update

nimjor:


content/knowledge-base/integrations/argo-cd-crossplane.md
```diff
@@ -63,6 +63,7 @@ data:
         end
 
         local has_no_status = {
+          "ProviderConfig", --ProviderConfig may have no status, or it may have a status.users field
```
plumbis, line 66: Is `--ProviderConfig may have no status, or it may have a status.users field` a comment about the field?

If so can we move this to a comment outside of the code block so a user doesn't accidentally copy/paste it? 

I'd suggest two changes:

1. Change the code fence on line 43 from ` ```yaml ` to ` ```yaml {label="argocfg"}`

2. Hint at the comment in line 42 by changing it to:

```
To configure it, edit the `argocd-cm` `ConfigMap` in the `argocd` `Namespace`. 
{{<hint "note" >}}
{{<hover label="argocfg" line="22">}} ProviderConfig{{</hover>}} may have no status or a `status.users` field. 
{{</hint >}}
```

plumbis (changes requested):

content/knowledge-base/integrations/argo-cd-crossplane.md
nimjor: I like it--fixed.

plumbis (approved):
Thanks for the update! looks good, Approved!

PR #679, opened 2024-01-18
Remove mention of the RBAC manager management policy

negz:
It's going away per https://github.com/crossplane/crossplane/issues/5227. See also https://github.com/crossplane/crossplane/issues/5227.


plumbis:
If we're doing away with the entire thing can you pull the section from the Crossplane Pods page? 
https://docs.crossplane.io/master/concepts/pods/#rbac-manager-pod

negz:
> If we're doing away with the entire thing

@plumbis We're not removing the whole RBAC manager, just one of its configuration options.

I didn't notice that that page mentioned the option we're removing though. I've added a commit that removes those mentions.

plumbis:
Gotcha! I wasn't entirely sure, thanks for confirming.

plumbis (approved):

PR #673, opened 2024-01-11
Update ArgoCD custom health status snippet

nimjor:

PR #687, opened 2024-01-29
fix: Invalid inequality operator in ArgoCD custom health for ProviderConfig

illrill:
Fixes mistake by #673. In Lua, the inequality operator is `~=` not `!=`. This caused an ArgoCD ComparisonError and sync failure.
```
error setting app health: failed to get resource health for "ProviderConfig" with name "azure" in namespace "": <string> line:27(column:56) near '!': Invalid token
```

### DIFF
--- a/content/knowledge-base/integrations/argo-cd-crossplane.md
+++ b/content/knowledge-base/integrations/argo-cd-crossplane.md
@@ -77,7 +77,7 @@ data:
         end
 
         if obj.status == nil or obj.status.conditions == nil then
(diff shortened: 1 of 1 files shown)


plumbis (approved):

denzhel:
Damn you're fast ... I was having this issue yesterday and I couldn't figure out what is the problem.
Refreshed the page and noticed that I had the `!` missing .. what are the odds that they fixed it that quick .. 
Thank you !!

PR #688, opened 2024-01-30
Document the crossplane.io/external-create-... annotations

negz:
These annotations were introduced in https://github.com/crossplane/crossplane-runtime/pull/283.

Per https://github.com/crossplane/crossplane/issues/3037 folks find these annotations hard to reason about. That's understandable, because they're doing a lot of subtle things.

This section ended up super long, but I think this is an area where folks really need to understand what's happening in order to make good decisions when Crossplane refuses to proceed.

### DIFF
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -678,8 +678,8 @@ the resource doesn't exist. When the provider thinks a resource doesn't exist
 it creates the resource.
 
 Some external systems don't let a provider specify a resource's name when the
-provider creates it. Instead the external system generates an unpredictable name
-and returns it to the provider.
+provider creates it. Instead the external system generates an nondeterministic
+name and returns it to the provider.
 
 When the external system generates the resource's name, it's critical that the
 provider saves it to the managed resource's `crossplane.io/external-name`
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -712,14 +712,6 @@ Events:
   Warning  CannotInitializeManagedResource  29m (x19 over 19h)  managed/queue.sqs.aws.crossplane.io  cannot determine creation result - remove the crossplane.io/external-create-pending annotation if it is safe to proceed
 ```
 
-{{<hint "important">}}
-The safest thing for a provider to do when it detects that it might have leaked
-a resource is to stop and wait for human intervention.
-
-This ensures the provider doesn't create duplicates of the leaked resource.
-Duplicate resources can be costly and dangerous.
-{{</hint>}}
-
 Providers use the creation annotations to detect that they might have leaked a
 resource.
 
@@ -738,6 +730,14 @@ The provider knows it might have leaked a resource because it updates all the
 resource's annotations at the same time. If the provider couldn't update the
 creation annotations after it created the resource, it also couldn't update the
 `crossplane.io/external-name` annotation.
+
+{{<hint "important">}}
+The safest thing for a provider to do when it detects that it might have leaked
+a resource is to stop and wait for human intervention.
+
+This ensures the provider doesn't create duplicates of the leaked resource.
+Duplicate resources can be costly and dangerous.
+{{</hint>}}
 
 {{<hint "tip">}}
 If a resource has a `cannot determine creation result` error, inspect the
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -740,8 +740,8 @@ creation annotations after it created the resource, it also couldn't update the
 `crossplane.io/external-name` annotation.
 
 {{<hint "tip">}}
-Inspect the external system when resources have a `cannot determine creation
-result` error.
+If a resource has a `cannot determine creation result` error, inspect the
+external system.
 
 Use the timestamp from the `crossplane.io/external-create-pending` annotation to
 determine when the provider might have leaked a resource. Look for resources
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -752,7 +752,7 @@ external system.
 
 Remove the `crossplane.io/external-create-pending` annotation from the managed
 resource after you're sure no leaked resource exists. This tells the provider to
-resume reconciliation of the managed resource.
+resume reconciliation of and recreate the managed resource.
 {{</hint>}}
 
 Providers also use the creation annotations to avoid leaking resources.
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -672,9 +672,9 @@ A provider uses the
 {{<hover label="creation" line="7">}}crossplane.io/external-name{{</hover>}}
 annotation to lookup a managed resource in an external system.
 
-If the provider can't find a managed resource in an external system, it thinks
-the resource doesn't exist. When the provider thinks a resource doesn't exist
-it creates the resource.
+The provider looks up the resource in the external system to determine if it
+exists, and if it matches the managed resource's desired state. If the provider
+can't find the resource, it creates it.
 
 Some external systems don't let a provider specify a resource's name when the
 provider creates it. Instead the external system generates an nondeterministic
@@ -688,14 +688,20 @@ A provider can't guarantee that it can save the annotation. The provider could
 restart or lose network connectivity between creating the resource and saving
 the annotation.
 
-{{<hint "important">}}
-Anytime an external system generates a resource's name there is a risk the
-provider could leak the resource.
-{{</hint>}}
-
 A provider can detect that it might have leaked a resource. If the provider
 thinks it might have leaked a resource, it stops reconciling it until you tell
 the provider it's safe to proceed.
+
+{{<hint "important">}}
+Anytime an external system generates a resource's name there is a risk the
+provider could leak the resource.
+
+The safest thing for a provider to do when it detects that it might have leaked
+a resource is to stop and wait for human intervention.
+
+This ensures the provider doesn't create duplicates of the leaked resource.
+Duplicate resources can be costly and dangerous.
+{{</hint>}}
 
 When a provider thinks it might have leaked a resource it creates a `cannot
 determine creation result` event associated with the managed resource. Use
@@ -730,14 +736,6 @@ The provider knows it might have leaked a resource because it updates all the
 resource's annotations at the same time. If the provider couldn't update the
 creation annotations after it created the resource, it also couldn't update the
 `crossplane.io/external-name` annotation.
-
-{{<hint "important">}}
-The safest thing for a provider to do when it detects that it might have leaked
-a resource is to stop and wait for human intervention.
-
-This ensures the provider doesn't create duplicates of the leaked resource.
-Duplicate resources can be costly and dangerous.
-{{</hint>}}
 
 {{<hint "tip">}}
 If a resource has a `cannot determine creation result` error, inspect the
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -653,7 +653,6 @@ Providers set three creation annotations to avoid and detect leaked resources:
 * `crossplane.io/external-create-failed` - The last time the provider failed to
   create the resource.
 
-
 Use `kubectl get` to view the annotations on a managed resource. For example, an
 AWS VPC resource:
 
@@ -681,9 +680,9 @@ Some external systems don't let a provider specify a resource's name when the
 provider creates it. Instead the external system generates an nondeterministic
 name and returns it to the provider.
 
-When the external system generates the resource's name, it's critical that the
-provider saves it to the managed resource's `crossplane.io/external-name`
-annotation. If it doesn't, it leaks the resource.
+When the external system generates the resource's name, the provider attempts to
+save it to the managed resource's `crossplane.io/external-name` annotation. If
+it doesn't, it _leaks_ the resource.
 
 A provider can't guarantee that it can save the annotation. The provider could
 restart or lose network connectivity between creating the resource and saving
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -658,7 +658,7 @@ Use `kubectl get` to view the annotations on a managed resource. For example, an
 AWS VPC resource:
 
 ```yaml {label="creation" copy-lines="2-9"}
-$ kubectl get vpc my-vpc
+$ kubectl get -o yaml vpc my-vpc
 apiVersion: ec2.aws.upbound.io/v1beta1
 kind: VPC
 metadata:
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -635,7 +635,8 @@ my-rds-instance      True    True     my-custom-name       11m
 
 ### Creation annotations
 
-In some rare situations a provider can forget that it created a resource. When
+When an external system like AWS generates nondeterministic resource names it's
+possible for a provider to create a resource but not record that it did. When
 this happens the provider can't manage the resource.
 
 {{<hint "tip">}}
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -635,13 +635,29 @@ my-rds-instance      True    True     my-custom-name       11m
 
 ### Creation annotations
 
-Providers use three annotations to avoid and detect leaking resources:
+In some rare situations a provider can forget that it created a resource. When
+this happens the provider can't manage the resource.
 
-* `crossplane.io/external-create-pending`
-* `crossplane.io/external-create-failed`
-* `crossplane.io/external-create-succeeded`
+{{<hint "tip">}}
+Crossplane calls resources that a provider creates but doesn't manage _leaked
+resources_.
+{{</hint>}}
 
-```yaml {label="creation"}
+Providers set three creation annotations to avoid and detect leaked resources:
+
+* {{<hover label="creation" line="8">}}crossplane.io/external-create-pending{{</hover>}} -
+  The last time the provider was about to create the resource.
+* {{<hover label="creation" line="9">}}crossplane.io/external-create-succeeded{{</hover>}} -
+  The last time the provider successfully created the resource.
+* `crossplane.io/external-create-failed` - The last time the provider failed to
+  create the resource.
+
+
+Use `kubectl get` to view the annotations on a managed resource. For example, an
+AWS VPC resource:
+
+```yaml {label="creation" copy-lines="2-9"}
+$ kubectl get vpc my-vpc
 apiVersion: ec2.aws.upbound.io/v1beta1
 kind: VPC
 metadata:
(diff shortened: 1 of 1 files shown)


content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,137 @@ my-rds-instance      True    True     my-custom-name       11m
...
+The safest thing for a provider to do when it detects that it might have leaked
+a resource is to stop and wait for human intervention.
+
+Waiting for human intervention ensures the provider doesn't create a duplicate
+of the leaked resource. Creating duplicate resources can be costly and
+dangerous.
+
+If you see the `cannot determine creation result` error, inspect the external
+system for resources created around the time recorded in the
 `crossplane.io/external-create-pending` annotation.
```
plumbis, line 728:
```suggestion
Waiting for human intervention prevents the provider from creating duplicate
resources. 

Inspect the external system when resources have a 
`cannot determine creation result` error.

Use the timestamp from the
`crossplane.io/external-create-pending` annotation to determine when the 
provider tried to create the resource.
```

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,137 @@ my-rds-instance      True    True     my-custom-name       11m
...
 
-The Provider applies the `crossplane.io/external-create-succeeded` or
-`crossplane.io/external-create-failed` annotation after making the external API
-call and receiving a response. 
+If you find a leaked resource, and it's safe to do so, delete it. Remove the
```
plumbis, line 730: delete it from the external system?

plumbis (changes requested):
A few questions and some edits just to make some of the writing more direct. Nothing major. Great content!

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,137 @@ my-rds-instance      True    True     my-custom-name       11m
...
 
-Providers create new managed resources with the
+Providers use three annotations to avoid and detect leaking resources:
+
+* `crossplane.io/external-create-pending`
+* `crossplane.io/external-create-failed`
+* `crossplane.io/external-create-succeeded`
+
+```yaml {label="creation"}
+apiVersion: ec2.aws.upbound.io/v1beta1
```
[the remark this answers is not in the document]
negz: Good idea, I was thinking this would look better with some text to break up the bullets and the snippet.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,137 @@ my-rds-instance      True    True     my-custom-name       11m
 
 ### Creation annotations
 
-Providers create new managed resources with the
+Providers use three annotations to avoid and detect leaking resources:
```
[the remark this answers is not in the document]
negz: I'm leaning toward keeping this as is (though `s/leaking/leaked`).

Generally I agree with introducing the term later, but I think "track the state" doesn't really say anything specific here. I feel okay with introducing the term, then building to explain what it means.

Also FWIW I think "leaked <thing>" is a fairly well understood concept so hopefully not too scary/confusing up front.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,137 @@ my-rds-instance      True    True     my-custom-name       11m
...
+
+{{<hint "tip">}}
+A provider uses a managed resource's `crossplane.io/external-name` annotation to
+know if it needs to create the resource in the external system. The annotation
+uniquely identifies the resource in the external system, such as AWS.
+
+If the provider doesn't save this annotation, it forgets that it created the
+resource. In fact, it doesn't know the resource ever existed and can't manage
+it. Crossplane calls this a leaked resource.
+{{</hint>}}
```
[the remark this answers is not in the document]
negz: I didn't go with exactly this suggestion, but did rephrase to hopefully capture the spirit.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,137 @@ my-rds-instance      True    True     my-custom-name       11m
...
+guarantees the provider doesn't leak the resource.
+
+Other external systems don't let the provider specify the name of the resource.
+Instead they generate an unpredictable name and include it in the response to
+the create API call the provider makes.
+
+{{<hint "important">}}
+For this second kind of external system, there is always a risk that a provider
+leaks a newly created resource.
+{{</hint>}}
```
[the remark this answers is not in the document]
negz: I like the suggestion for the `important` box.

For the rest, I think the key suggestion is to not spend time on the "happy path" that describes how a well-designed external system works (and where these annotations aren't needed). If so, done. 😄

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,137 @@ my-rds-instance      True    True     my-custom-name       11m
...
+
+A provider can't guarantee that it can save the generated name to the
+`crossplane.io/external-name` annotation. If the provider crashes, restarts, or
+can't reach the network after creating the resource it can't save the
+annotation. If this happens, the provider leaks the resource.
```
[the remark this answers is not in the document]
negz: Close. Perhaps it would help to mention that providers are stateless. When a provider starts, it can only go off the state it reads from the API server (i.e. the recorded state of the MR).

So the timeline is more like this:

* T0: POST from provider
* T1: External resource is created
* T2: Provider stops (e.g. crash, restart) before it can save the `crossplane.io/external-name` annotation
* T3: Provider starts
* T4: Provider loads MR from API server

At T4 the provider only knows what it can read from the API server. When it reads the MR, it has no `crossplane.io/external-name` annotation. Or rarely (if the provider was _re-creating_ the resource because someone deleted it using the cloud console) has a stale `crossplane.io/external-name` annotation, pointing to an external resource that no longer exists.

So at this point as far as the provider can tell the MR doesn't exist, and needs to be created.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,137 @@ my-rds-instance      True    True     my-custom-name       11m
...
+* If the create fails, it writes the current time to the
+  `crossplane.io/external-create-failed` annotation.
+
+Providers try hard to write these annotations to the managed resource.
+If writing the annotations fails, the provider retries several times.
+
+When a provider sees a managed resource with:
+
+* A `crossplane.io/external-create-pending` annotation.
+* No `crossplane.io/external-create-succeeded` or
```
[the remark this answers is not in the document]
negz: I rephrased this to be more accurate and hopefully clearer.

Technically the condition that must be true is the create-pending annotation is the most recent of the three. If an annotation is missing we consider it to be set at time zero (i.e. the epoch). So create-pending is always more recent than (e.g.) create-succeeded it create-succeeded isn't set.

negz:
@plumbis PTAL. I started addressing feedback and ended up with a fairly extensive refactoring.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
...
+  name: my-vpc
+  annotations:
+    crossplane.io/external-name: vpc-1234567890abcdef0
+    crossplane.io/external-create-pending: "2023-12-18T21:48:06Z"
+    crossplane.io/external-create-succeeded: "2023-12-18T21:48:40Z"
+```
+
+A provider uses the
+{{<hover label="creation" line="7">}}crossplane.io/external-name{{</hover>}}
+annotation to find a resource in an external system, for example AWS.
```
plumbis, line 673:
```suggestion
annotation to map a resource in an external system, for example AWS, to a specific managed resource.
```

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
...
+
+If the provider can't find a managed resource in an external system, it thinks
+the resource doesn't exist. When the provider thinks a resource doesn't exist
+it creates the resource.
```
plumbis, line 677:
```suggestion
If the provider finds a managed resource with a
{{<hover label="creation" line="7">}}crossplane.io/external-name{{</hover>}} 
annotation and doesn't find an external resource with a matching name, the 
provider creates a new external resource.
```
Feel free to rework, but I think this should be "dumbed down" a bit and be more explicit. 
- What is the provider looking for in the external system?
- How does that get connected to an MR?
- Avoid "think"

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
...
+
+Some external systems don't let a provider specify a resource's name when the
+provider creates it. Instead the external system generates an unpredictable name
```
plumbis, line 680:
```suggestion
provider creates it. Instead the external system generates a name
```

It's only unpredictable from the POV of the provider. This could trip someone up if they say "hey, the external system docs say this is how they generate a name, right here, so I guess this doesn't apply"

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
...
+and returns it to the provider.
+
+When the external system generates the resource's name, it's critical that the
+provider saves it to the managed resource's `crossplane.io/external-name`
+annotation. If it doesn't, it leaks the resource.
```
plumbis, line 685:
```suggestion
When the external system generates the resource's name, the
provider attempts saves it to the managed resource's `crossplane.io/external-name`
annotation. If it doesn't, it _leaks_ the resource.
```

The "critical" part here is a detail for building providers, not operating crossplane.

Putting `leaks` in italics just to draw attention and strengthen the point.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
...
+  Warning  CannotInitializeManagedResource  29m (x19 over 19h)  managed/queue.sqs.aws.crossplane.io  cannot determine creation result - remove the crossplane.io/external-create-pending annotation if it is safe to proceed
+```
+
+{{<hint "important">}}
+The safest thing for a provider to do when it detects that it might have leaked
+a resource is to stop and wait for human intervention.
+
+This ensures the provider doesn't create duplicates of the leaked resource.
+Duplicate resources can be costly and dangerous.
+{{</hint>}}
```
plumbis, line 721: I'd move this after you introduce and discuss how resource leak detection works.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
...
+
+The provider knows it might have leaked a resource because it updates all the
+resource's annotations at the same time. If the provider couldn't update the
+creation annotations after it created the resource, it also couldn't update the
+`crossplane.io/external-name` annotation.
+
+{{<hint "tip">}}
+Inspect the external system when resources have a `cannot determine creation
+result` error.
```
plumbis, line 744:
```suggestion
If a resource has a `cannot determine creation
result` error, inspect the external system.
```

Rewrite as "if <condition> then <action>". Improves skim-ability.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
...
+Use the timestamp from the `crossplane.io/external-create-pending` annotation to
+determine when the provider might have leaked a resource. Look for resources
+created around this time.
+
+If you find a leaked resource, and it's safe to do so, delete it from the
+external system.
+
+Remove the `crossplane.io/external-create-pending` annotation from the managed
+resource after you're sure no leaked resource exists. This tells the provider to
+resume reconciliation of the managed resource.
```
plumbis, line 755:
```suggestion
resume reconciliation and recreate the managed resource.
```
Explicitly state that the provider will try to create a new external resource now.

plumbis (changes requested):

content/master/concepts/managed-resources.md
negz: I don't think that's true. In many cases it's unpredictable to any outside observer.

I wanted to say non-deterministic. Happy to say that instead if we feel comfortable with that terminology.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
...
+  The last time the provider successfully created the resource.
+* `crossplane.io/external-create-failed` - The last time the provider failed to
+  create the resource.
+
+
+Use `kubectl get` to view the annotations on a managed resource. For example, an
+AWS VPC resource:
+
+```yaml {label="creation" copy-lines="2-9"}
+$ kubectl get vpc my-vpc
```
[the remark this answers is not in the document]
negz: Updated to `kubectl get -o yaml`.

content/master/concepts/managed-resources.md
negz: I don't love "map" as what's being done here. I switched to "lookup".

content/master/concepts/managed-resources.md
negz: Updated to say nondeterministic.

This detail feels important to me, as an explanation of why we can't do better here. If there was a way to predict the name, e.g. because it would be generated in a predictable way, we wouldn't need to do this potentially frustrating annotation dance.

content/master/concepts/managed-resources.md
negz: I moved this, but then moved it back up and merged it with the other important note heh.

I feel like this section is building up the information needed to understand what the problem is:

1. Provider uses external-name to lookup resource.
2. If resource doesn't exist, provider creates it.
3. Provider can't guarantee external-name will be saved.
4. Safest thing to do is stop, because per step 2, this would otherwise result in a duplicate.

content/master/concepts/managed-resources.md
negz: Reworked, PTAL.

I prefer not to mention `external-name` again explicitly here. It feels repetitive - how the provider looks up the MR - which is what the sentence immediately prior says.

"If the provider finds an MR with..." reads strangely to me too - MRs always have this annotation.

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,145 @@ my-rds-instance      True    True     my-custom-name       11m
 
 ### Creation annotations
 
-Providers create new managed resources with the
-`crossplane.io/external-create-pending` annotation.
+In some rare situations a provider can forget that it created a resource. When
```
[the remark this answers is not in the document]
negz: Point taken. Reworded to avoid "rare" and "forgot".

That said, I would prefer if folks did not read this section and think "Crossplane's going to leak my resources all the time, I shouldn't trust it".

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,143 @@ my-rds-instance      True    True     my-custom-name       11m
...
-`crossplane.io/external-create-pending` annotation.
+When an external system like AWS generates nondeterministic resource names it's
+possible for a provider to create a resource but not record that it did. When
+this happens the provider can't manage the resource.
 
-The Provider applies the `crossplane.io/external-create-succeeded` or
-`crossplane.io/external-create-failed` annotation after making the external API
-call and receiving a response. 
+{{<hint "tip">}}
+Crossplane calls resources that a provider creates but doesn't manage _leaked
```
luxas, line 643:
```suggestion
Crossplane calls resources that a provider created but lost track of and therefore doesn't manage _leaked
```

luxas: trying to word in such a way so that readers would understand instantly that this is an error case, without using the word "error", "exception" or "rare"

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,143 @@ my-rds-instance      True    True     my-custom-name       11m
...
+exists, and if it matches the managed resource's desired state. If the provider
+can't find the resource, it creates it.
+
+Some external systems don't let a provider specify a resource's name when the
+provider creates it. Instead the external system generates an nondeterministic
+name and returns it to the provider.
+
+When the external system generates the resource's name, the provider attempts to
+save it to the managed resource's `crossplane.io/external-name` annotation. If
+it doesn't, it _leaks_ the resource.
```
luxas, line 685:
```suggestion
it cannot (after multiple retries), it _leaks_ the resource.
```

content/master/concepts/managed-resources.md
```diff
@@ -635,21 +635,143 @@ my-rds-instance      True    True     my-custom-name       11m
...
+  Warning  CannotInitializeManagedResource  29m (x19 over 19h)  managed/queue.sqs.aws.crossplane.io  cannot determine creation result - remove the crossplane.io/external-create-pending annotation if it is safe to proceed
+```
+
+Providers use the creation annotations to detect that they might have leaked a
+resource.
+
+Each time a provider reconciles a managed resource it checks the resource's
+creation annotations. If the provider sees a create pending time that's more
+recent than the most recent create succeeded or create failed time, it knows
+that it might have leaked a resource.
```
luxas, line 727: Are there any other common cases where this would happen than when someone went to the AWS UI and removed a reconciled bucket manually, and then Crossplane tried to recreate it, but failed to save the new name in the annotations?

luxas (commented):
Thanks a lot for this write-up @negz 💯 

Level triggered systems are very interesting 😄 

One of my favorite presentations about this is from thockin (avoiding at-ing him here): https://speakerdeck.com/thockin/kubernetes-what-is-reconciliation

With that comes to mind, that how many Crossplane controllers are able to attach (in general, approximately) labels, annotations, or similar to their managed data in the external provider, such that they can (at least in theory), find and point out all of these leaked resources, as they can see that they are indeed owned by "me" (through that label or annotation in the external system), but there is no external-name reference to them in Kube?

This indeed, would require controllers to have some kind of "identity", however, which is probably a big challenge itself (such that two controllers in two different clusters could distinguish their managed objects in the external system).

plumbis (approved):

negz:
> how many Crossplane controllers are able to attach (in general, approximately) labels, annotations, or similar to their managed data in the external provider, such that they can (at least in theory), find and point out all of these leaked resources, as they can see that they are indeed owned by "me" (through that label or annotation in the external system), but there is no external-name reference to them in Kube?

@luxas We considered this. It's a non-zero amount of controllers. However given:

* The number of controllers
* The asymmetry of the APIs those controllers orchestrate
* That many of those APIs are eventually consistent (will take a while for reads to show results after a write)
* The fact that many APIs that support tags/labels use a separate API call to do so

I'm skeptical there's value in exploring that path. Or at least it's deep into diminishing returns. At best we (Crossplane core) could offer some interface an `ExternalClient` (MR implementation) could optionally satisfy to lookup by label rather than by name and fall back to using that. It would then be up to any MR implementation that could to satisfy that interface (presumably by hand-writing the relevant code).

content/master/concepts/managed-resources.md
negz: No, nothing comes to mind.

negz:
Thanks folks! We've spent a lot of time this week wordsmithing one this week. I like where it's at right now and I need to move on to other things, so I'm going to merge it as-is. 🙂

PR #697, opened 2024-02-04
Remove false restriction about using '.status.atProvider' into ToEnvironmentFieldPath

yogeek:
As confirmed @phisco in [Slack](https://crossplane.slack.com/archives/CEG3T90A1/p1706970216073849?thread_ts=1706970216.073849&cid=CEG3T90A1), the 'patch and transform' documentation is wrong about the impossibility to use data from `.status.atProvider` in the `fromFielPath` of the `ToEnvironmentFieldPath` patch.

This restriction is incompatible with the example in the `EnvironmentConfig` page that shows it is :
 ```
patches:
- type: ToEnvironmentFieldPath
   fromFieldPath: status.atProvider.id
   toFieldPath: vpcId
```
https://docs.crossplane.io/latest/concepts/environment-configs/#patch-an-individual-resource

### DIFF
--- a/content/master/concepts/patch-and-transform.md
+++ b/content/master/concepts/patch-and-transform.md
@@ -827,15 +827,6 @@ kind: Composition
 Because the environment is in-memory, there is no command to confirm the patch
 wrote the value to the environment.
 
-{{<hint "important">}}
-The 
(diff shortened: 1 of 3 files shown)


phisco (approved):
Thanks @yogeek!

Issue #600, opened 2023-11-01
Make "master" directory use easier


plumbis:
When we are about to publish docs the fact there is "master" and a unpublished release creates a commit and syncing problem. 

It's also confusing when navigating the docs since you can't (accidentally) install the master release. 

We should (again) consider getting rid of the master folder.

phisco:
If we removed master today, where should we put docs for new functionalities that will be in 1.15, now that 1.14 is out? I thought master was for that.

plumbis:
Today that's true. The problem is that when we prepare a new release there are a [number of things ](https://github.com/crossplane/docs/blob/master/.github/ISSUE_TEMPLATE/new_release.md) that have to be updated to make the build and publish parts all point to the new $latest version. 

There are two options:
1. We create a new version folder before the release and add new content there along with the required infra changes to make sure it all works in staging. Contributors have to be mindful to contribute to two or three places (master, $old_latest, $new_latest) and/or the right branch.
2. We turn master in to $new_latest as the last step in the release process, change the infra settings and push.

The problem with (2) is that it makes the release unnecessarily risky and stressful. Failure to get some of the things correct either delays docs readiness or causes other issues in the docs, like what redirects, dropdowns and landing pages look like. 

I don't have an answer to the problem yet, but I'm not a fan with what we have today.

phisco:
I feel like you are describing all good reasons to use release branches 😅

plumbis:
The challenge is that our build artifact is a monolith. It doesn't lend itself well to having branches with different shortcodes/layout/templates/structure. 

IMO release branches just shifts the problem around, making the standard build process more complex.

plumbis:
After the 1.15 release I'm not sure there's a great solution that isn't just more process without any benefit. Closing this as "won't fix"

PR #701, opened 2024-02-13
v1.15 Docs

plumbis:
Docs for new v1.15 release.


content/master/software/install.md
```diff
@@ -244,9 +244,9 @@ at the table below.
 {{< table caption="Feature flags" >}}
```
jbw976, line 244: i would expect this page in `master` to be updated with the same content that is in the v1.15 version.  e.g. this block in the 1.15 page about changing the default registry:

* https://deploy-preview-701--crossplane.netlify.app/v1.15/software/install/#change-the-default-package-registry 
* https://github.com/crossplane/docs/pull/701/files#diff-69a7157c116add5db8956596b933da896f47aafaaf82f6f562e6abd1a409e1f2R264-R272

Does not seem to be in this master version of the page, e.g. https://deploy-preview-701--crossplane.netlify.app/master/software/install/#change-the-default-package-registry doesn't exist.

Is that expected for you? and if not, is there possibly more missing content that is in 1.15 but not showing up in master?

content/v1.15/software/install.md
```diff
@@ -264,18 +247,29 @@ at the table below.
...
 `--set` flag, for example: `--set
 args='{"--enable-composition-functions","--enable-composition-webhook-schema-validation"}'`.
 
+#### Change the default package registry
+
+Beginning with Crossplane version 1.15.0 Crossplane downloads packages from the
+[Upbound Marketplace](https://marketplace.upbound.io) at `xpkg.upbound.io` 
+instead of DockerHub. 
+
+Change the default registry location during the Crossplane install with
```
jbw976, line 270: Did you also want to include content around how to clean-up/fix the scenario where folks miss the notification about the default registry changing and they upgrade to v1.15 without taking that into account?

e.g. the very raw content from https://gist.github.com/jbw976/793210f26f2dfbb458a41d5dc437ce7e?

content/master/software/install.md
plumbis: I'm planning on syncing the `master` folder from `1.15` once all the content is in. right now half the stuff is in master and half in 1.15. 

Because of the way some 1.14 fixes went into master but not 1.14, some in 1.14 and not master and most new content in master, but not the stuff I'm writing it's a bit of a mess. 

Consider the 1.15 directory as the source of truth right now.

[Related discussion](https://github.com/crossplane/docs/issues/600).

content/v1.15/software/install.md
plumbis: I want to bring this into a new "upgrade" chapter that I'm going to write shortly after release, since the context for this is only related to upgrades and not new installs.

jbw976: Sounds good - a fast follow on that would be very appreciated, because I know over the next few days we'll have at least some folks that need to get sorted out with their packages and registry default 😞 - so being able to efficiently support them (e.g. a consistent docs spot to point them at) will be really helpful! 🙇‍♂️

jbw976 (approved):

PR #705, opened 2024-02-14
Document extra resources for functions

phisco:


plumbis (changes requested):
Can you add `cluster-scoped` to utils/vale/styles/Crossplane/allowed-jargon.txt to pass Vale checks?

plumbis (approved):

PR #549, opened 2023-09-14
Update guidlines on envConfigs and patching to XRs

plumbis:
Currently there are no docs on patching to an XR using an environment config using `spec.environment.patches` of a Composition. 

This adds details on this and removes an incorrect statement about the EC key needing to exist before writing to it.

PR #707, opened 2024-02-14
remove incorrect restriction about envkeys in patches

plumbis:
In #549 we removed references that envkeys had to exist before a patch, but missed some references. This should clean up the remaining references.

### DIFF
--- a/content/v1.13/concepts/patch-and-transform.md
+++ b/content/v1.13/concepts/patch-and-transform.md
@@ -964,10 +964,6 @@ strings.
 The {{<hover label="combineToEnv" line="20">}}toFieldPath{{</hover>}} is the
 key in the environment to write the new string to. 
 
-{{< hint "important" >}}
-The environment's key must already exist. Patches can't create new environment
-keys. 
-{{< /hint >}}
 
 ```yaml {label="combineToEnv",copy-lines="none"}
 apiVersion: apiextensions.crossplane.io/v1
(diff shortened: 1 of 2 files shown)


bobh66 (approved):

phisco (commented):
Turns out that was actually correct, a bug, but correct 😅

plumbis:
I feel less crazy then @phisco .

The previous PR removed half the references but missed some. Should we pull them all and document expected behavior or keep the restriction notices in since it sounds like we may not fix this bug for some time? 

I'd vote to keep the notices in until we know the bug will be fixed.

phisco:
Yeah, probably reverting that PR makes more sense. Until we decide to fix it and cut a patch release at least. Maybe we could mention that the restriction doesn't apply to function-patch-and-transform.

bobh66:
#5388 talks about not patching from composed resources - how does that relate to creating new keys in the environment?  Does that mean all non-`environment.patches` patching to the environment is broken?

phisco:
🤦‍♂️ @bobh66 you are right, I thought this was about https://github.com/crossplane/docs/pull/697, that was indeed unfortunately correct and a bug, see https://github.com/crossplane/crossplane/issues/5388. So we could think about reverting that one. This one looks good.
Sorry for the confusion, it's been a long week 😅

phisco (approved):

jeanduplessis (approved):

Issue #709, opened 2024-02-14
Composition Revisions are GA


negz:
### What's Wrong?

https://docs.crossplane.io/knowledge-base/guides/composition-revisions/
https://docs.crossplane.io/knowledge-base/guides/composition-revisions-example/

Composition revisions became GA in v1.12 (https://github.com/crossplane/crossplane/issues/3715), but our KB guides still show them as beta features.

I believe the general idea is:

* Alpha features live outside the "main" (versioned) docs flow, in the KB.
* When a feature hits GA (or possibly beta?) it gets integrated into the main docs flow.

So we should probably merge the composition revisions documentation into the general compositions documentation.

PR #710, opened 2024-02-14
Don't say composition revisions are beta

negz:
Relates to https://github.com/crossplane/docs/issues/709. I don't think it's a complete 'fix'.

Revisions became GA in v1.12 per https://github.com/crossplane/crossplane/issues/3715. I think there's more to do to integrate them into the main docs flow as a GA feature, but at minimum let's stop actively saying they're beta.

When the feature became GA, we:

* Made it impossible to disable the feature using the flag.
* Hid the feature flag from the help output.
* Log a warning that the flag will be removed if it is passed.

### DIFF
--- a/content/knowledge-base/guides/composition-revisions-example.md
+++ b/content/knowledge-base/guides/composition-revisions-example.md
@@ -1,8 +1,5 @@
 ---
 title: Composition Revision Example
-state: beta
-alphaVersion: "1.4"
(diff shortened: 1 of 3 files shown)


negz:
Looks like all the Vale errors pertain to content I didn't touch. In the interest of time I'd prefer not to address them in this PR.

Issue #698, opened 2024-02-05
Document the `--enable-claim-ssa` feature flag


negz:
### What's Missing?

We should document the flag introduced in https://github.com/crossplane/crossplane/pull/5313:

* The new SSA-based claim controller fixes a bunch of bugs (especially when used w/functions), mostly because it implements https://github.com/crossplane/crossplane/issues/4581.
* There's a behavior change in that changes won't be back-propagated from XR spec -> claim spec anymore.

plumbis:
Is this planned to be in 1.15?

negz:
@plumbis It is, yes.

PR #711, opened 2024-02-15
Document the --enable-claim-ssa feature flag

negz:
Fixes https://github.com/crossplane/docs/issues/698

This feature flag changes how the claim controller works. It's a bug fix. We usually wouldn't put a bug fix behind a feature flag, but this one introduces several behavior changes that we don't want to catch people unaware.

### DIFF
--- a/content/master/concepts/server-side-apply.md
+++ b/content/master/concepts/server-side-apply.md
@@ -1,0 +1,105 @@
+---
+title: Server-Side Apply
+state: alpha
+alphaVersion: "1.15"
+weight: 300
+---
+
+Crossplane can use server-side apply to sync claims with composite resources
+(XRs), and to sync composite resources with composed resources.
+
+When Crossplane uses server-side apply, the Kubernetes API server helps sync
+resources. Using server-side apply makes syncing more predictable and less
+buggy.
+
+{{<hint "tip">}}
+Server-side apply is a Kubernetes feature. Read more about server-side apply in
+the [Kubernetes documentation](https://kubernetes.io/docs/reference/using-api/server-side-apply/).
+{{</hint>}}
+
+## Use server-side apply to sync claims with composite resources
+
+When you create a claim, Crossplane creates a corresponding composite resource.
+Crossplane keeps the claim in sync with the composite resource. When you change
+the claim, Crossplane reflects those changes on the composite resource.
+
+Read the [claims documentation]({{<ref "./claims">}}) to learn more about claims
+and how they relate to composite resources.
+
+Crossplane can use server-side apply to keep the claim in sync with the
+composite resource.
+
+Use the `--enable-ssa-claims` feature flag to enable using server-side apply.
+Read the [Install Crossplane documentation]({{<ref "../software/install#feature-flags">}})
+to learn about feature flags.
+
+If you see fields reappearing after you delete them from a claim's `spec`,
+enable server-side apply to fix the problem. Enabling server-side apply also
+fixes the problem where Crossplane doesn't delete labels and annotations from
+the composite resource when you delete them from the claim.
+
+{{<hint "important">}}
+When you enable server-side apply, Crossplane is stricter about how it syncs
+a claim with its counterpart composite resource:
+
+- The claim's `metadata` syncs to the composite resource's `metadata`.
+- The claim's `spec` syncs to the composite resource's `spec`.
+- The composite resource's `status` syncs to the claim's `status`.
+
+When you enable server-side apply Crossplane doesn't sync the composite resource's `metadata`
+and `spec` back to the claim's `metadata` and `spec`. It also doesn't sync the
+claim's `status` to the composite resource's `status`.
+{{</hint>}}
+
+## Use server-side apply to sync claims end-to-end
+
+To get the full benefit of server-side apply, use the `--enable-ssa-claims`
+feature flag together with composition functions.
+
+When you use composition functions, Crossplane uses server side apply to sync
+composite resources with composed resources. Read more about this in the
+[composition functions documentation]({{<ref "./composition-functions#how-composition-functions-work">}}).
+
+```mermaid
+graph LR
+  A(Claim) -- claim server-side apply --> B(Composite Resource)
+  B -- function server-side apply --> C(Composed Resource)
+  B -- function server-side apply --> D(Composed Resource)
+  B -- function server-side apply --> E(Composed Resource)
+```
+
+When you use server-side apply end-to-end there is a clear, predictable
+propagation of fields from claim to composed resources, and back:
+
+* `metadata` and `spec` flow forwards, from claim to XR to composed resources.
+* `status` flows backwards, from composed resources to XR to claim.
+
+{{<hint "important">}}
+When you use composition functions, Crossplane is stricter about how it syncs
+composite resources (XRs) with composed resources:
+
+- The XR's `metadata` syncs to the composed resource's `metadata`.
+- The XR's `spec` syncs to the composed resource's `spec`.
+- The composed resource's `status` syncs to the XR's `status`.
+
+When you use composition functions Crossplane doesn't sync the composed resource's `metadata`
+and `spec` back to the XR's `metadata` and `spec`.
+{{</hint>}}
+
+When Crossplane uses server-side apply end-to-end to sync claims with composed
+resources, it deletes fields from a composed resource's `spec` when you
+delete fields from the claim's `spec`.
+
+When Crossplane uses server-side apply end-to-end it's also able to merge claim
+fields into complex composed resource fields. Objects and arrays of objects are
+examples of complex composed resource fields.
+
+{{<hint "tip">}}
+Crossplane can only merge complex fields for resources that use server-side
+apply merge strategy OpenAPI extensions. Read about these extensions in the
+Kubernetes [server-side apply documentation](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy).
+
+If you find that Crossplane isn't merging managed resource fields, raise an
+issue against the relevant provider. Ask the provider maintainer to add
+server-side apply merge strategy extensions to the managed resource.
+{{</hint>}}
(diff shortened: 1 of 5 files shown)


content/master/concepts/claim-server-side-apply.md
```diff
@@ -0,0 +1,74 @@
...
+- The claim's `metadata` syncs to the composite resource's `metadata`.
+- The claim's `spec` syncs to the composite resource's `spec`.
+- The composite resource's `status` syncs to the claim's `status`.
+
+When you enable server-side apply Crossplane doesn't sync the composite resource's `metadata`
+and `spec` back to the claim's `metadata` and `spec`. It also doesn't sync the
+claim's `status` to the composite resource's `status`.
+{{</hint>}}
+
+## Use claim server-side apply with composition functions
```
pedjak, line 53: IMHO, there is no direct connection between claim server-side apply and usage of composition functions, and hence a reader might be confused. Perhaps the following paragraph should be moved somewhere under `composition-function.md`?

content/master/concepts/claim-server-side-apply.md
```diff
@@ -0,0 +1,74 @@
...
+```mermaid
+graph LR
+  A(Claim) -- claim server-side apply --> B(Composite Resource)
+  B -- function server-side apply --> C(Composed Resource)
+  B -- function server-side apply --> D(Composed Resource)
+  B -- function server-side apply --> E(Composed Resource)
+```
+
+{{<hint "tip">}}
+When Crossplane uses server-side apply to sync claims with composed resources,
```
pedjak, line 68: > When Crossplane uses server-side apply to sync claims with composed resources,

Do you mean here actually "to sync composite with composed resources"?

`--enable-claim-ssa` is not about merging complex fields - given that Crossplane owns both the claim and the composite, it ensures just the proper propagation of claim's changes down to the counterpart composite.

Merging complex fields are interesting when several different actors are owning a part of it (e.g. an item in the list or a field in an object). Such situations can appear for composite to composed syncs.

content/master/software/install.md
```diff
@@ -251,6 +251,7 @@ at the table below.
 | Alpha | `--enable-external-secret-stores` | Enable support for External Secret Stores. |
 | Alpha | `--enable-usages` | Enable support for Usages. |
 | Alpha | `--enable-realtime-compositions` | Enable support for real time compositions. |
+| Alpha | `--enable-claim-ssa` | Enable support for using server-side apply to sync claims with XRs. |
```
pedjak, line 254: We perform field manager migration for claims created by previous versions of crossplane. This migration is irreversible, i.e. once performed by turning the flag on, it cannot be reverted by turning the flag off.

plumbis: Would it require a reinstalling Crossplane to reset?

plumbis (changes requested):

content/master/software/install.md
pedjak: reinstalling Crossplane would not help - the data are affected. The best bet would be to make a backup before turning the flag on. The backup solution should be able to restore `managedFields` properly, in case when the we want to restart Crossplane with the flag off.

phisco: Tested this out and in my scenario it worked fine disabling the flag.
I did the following:
- installed with --enable-ssa-claim
- deployed an XRC with a map field with keys “foo” and “bar”
- removing “foo” from the XRC did remove it from the XR as expected
- added back “foo” to the XRC
- reinstalled without --enable-ssa-claim
- tried removing foo from the XRC, foo was NOT removed from the XR as expected
- added something to the map in the XRC, which was correctly propagated to the XR
Did I miss something?

pedjak: My comment was motivated by https://github.com/kubernetes/client-go/blob/master/util/csaupgrade/upgrade.go#L61-L80.  @phisco I have tried the same steps, and noticed that ssa manager remains after reinstalling without the flag. Not sure if this can produce new issues.

content/master/concepts/claim-server-side-apply.md
```diff
@@ -0,0 +1,74 @@
...
+change the claim, Crossplane reflects those changes on the composite resource.
+
+Read the [claims documentation]({{<ref "./claims">}}) to learn more about claims
+and how they relate to composite resources.
+
+Crossplane can use server-side apply to keep the claim in sync with the
+composite resource.
+
+{{<hint "tip">}}
+Use the `--enable-claim-ssa` feature flag to enable using server-side apply.
```
pedjak, line 19: by looking at the crossplane source code, it looks that the flag name is actually `--enable-ssa-claims`, right @negz ?

content/master/concepts/claim-server-side-apply.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Claim Server-Side Apply
```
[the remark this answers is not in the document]
negz: I prefer keeping alpha features out of the main flow - not mixed in. This makes it _really_ clear that it's separate and alpha.

@plumbis FWIW about 60% of this preference is because AFAIK our nice "this feature is alpha, it was introduced in version X" widget only works as front matter for a whole page. If I could do it on a heading I'd be more amenable to merging alpha features into main docs flows as a separate heading.

```markdown
state: alpha
alphaVersion: "1.15"
```

content/master/concepts/claim-server-side-apply.md
negz: The point is:

* There are additional benefits to end-to-end SSA. From claim -> XR -> composed.
* You only get end-to-end SSA when you use this alpha feature AND functions.

Given you have to use an alpha feature, I think the docs fit better "aligned" with the alpha feature (claim SSA) than the functions docs.

content/master/concepts/claim-server-side-apply.md
negz: Haha yes it is. Good catch. 🙃

content/master/concepts/claim-server-side-apply.md
```diff
@@ -0,0 +1,74 @@
...
+Crossplane can use server-side apply to keep the claim in sync with the
+composite resource.
+
+{{<hint "tip">}}
+Use the `--enable-claim-ssa` feature flag to enable using server-side apply.
+
+Read the
+[Install Crossplane documentation]({{<ref "../software/install#feature-flags">}})
+to learn about feature flags.
+{{</hint>}}
```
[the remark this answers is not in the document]
negz: Fair point. I was mostly going with tip to make it "pop" on the page.

content/master/software/install.md
negz: Would it make sense to address this by adding a note to `claim-server-side-apply.md` that if you disable the feature flag, Crossplane won't clean up SSA field managers? We could link to upstream docs about what field managers are.

It sounds like leaving SSA field managers might have no meaningful impact though, so not sure how valuable this is document.

content/master/concepts/claim-server-side-apply.md
negz: > Do you mean here actually "to sync composite with composed resources"?

I mean to sync claims with composed resources. I'm intentionally omitting the fact that there's actually an XR in the middle, to avoid going into too much detail. My hope is that the diagram above, which shows the XR in the middle, would make this apparent.

I understand that claim level SSA does not in and of itself help with merging fields. The point I'm trying to make is this one: https://github.com/crossplane/docs/pull/711/files#r1491652688. I think my point is not coming across. Let me try to rewrite this section.

content/master/concepts/sever-side-apply.md
```diff
@@ -0,0 +1,106 @@
+---
+title: Server-Side Apply
+state: alpha
+alphaVersion: "1.15"
```
negz, line 4: It would be ideal to clarify that only claim SSA (the first heading) is alpha.

That said, I think this is fine as-is. Better to be a bit scary and frame all of SSA as alpha, even if the functions part of it is beta, than to risk it seeming more mature than it is.

negz:
@plumbis @pedjak PTAL.

I've added a commit that:

* Frames the concept around SSA in general, not just claims.
* Explains the claims part
* Explains why SSA is best used end-to-end (claims and functions)

content/master/software/install.md
pedjak: > It sounds like leaving SSA field managers might have no meaningful impact though, so not sure how valuable this is document.

It is really hard to judge about the implications, because fields in a composite will be still owned by now not existing manager.

content/master/concepts/server-side-apply.md
```diff
@@ -0,0 +1,105 @@
...
+  B -- function server-side apply --> E(Composed Resource)
+```
+
+When you use server-side apply end-to-end there is a clear, predictable
+propagation of fields from claim to composed resources, and back:
+
+* `metadata` and `spec` flow forwards, from claim to XR to composed resources.
+* `status` flows backwards, from composed resources to XR to claim.
+
+{{<hint "important">}}
```
pedjak, line 77: It would not hurt to repeat this hint in `composition-functions.md` as well?

pedjak (commented):
a few minor things, lgtm.

content/master/software/install.md
plumbis: I don't totally follow this conversation but it sounds like if ssa is enabled there are some things that can't be undone, right? 

I just want to make sure we call this out. We don't need to make the reader nervous but something like "Enabling SSA is a permanent change to your Kubernetes cluster and can only be undone by <uninstalling crossplane?>"

plumbis (approved):

content/master/concepts/server-side-apply.md
negz: The wording is different, but this constraint is documented at https://docs.crossplane.io/v1.15/concepts/composition-functions/#desired-state.

content/master/concepts/server-side-apply.md
```diff
@@ -0,0 +1,105 @@
...
+to learn about feature flags.
+
+If you see fields reappearing after you delete them from a claim's `spec`,
+enable server-side apply to fix the problem. Enabling server-side apply also
+fixes the problem where Crossplane doesn't delete labels and annotations from
+the composite resource when you delete them from the claim.
+
+{{<hint "important">}}
+When you enable server-side apply, Crossplane is stricter about how it syncs
+claims with composed resources:
```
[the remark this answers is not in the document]
negz: I do, updated. Thanks!

negz:
@plumbis Want me to replicate this content in the v1.15 directory before I merge?

plumbis:
yes please!

Issue #665, opened 2023-12-28
Argo docs don't work correctly


teich:
the docs on argo don't seem to work with the latest/current version of argo correctly.

https://docs.crossplane.io/knowledge-base/integrations/argo-cd-crossplane/#hl-1-6

I setup a test repo and had to remove the crossplane reference, AND also add a skip for providerconfig on line 24.

https://github.com/teich/argo-xp/blob/main/argocd/configmap.yaml

PR #734, opened 2024-03-19
fix: add missing conditional for ignore Diffs

haooliveira84:
# What this PR does?

Add missing conditional from `providerConfig.aws.crossplane.io` status

Fixes: https://github.com/crossplane/docs/issues/665#issue-2057871413


plumbis:
thanks for the fix!

PR #735, opened 2024-03-21
Fix: update quickstart azure provider version

loomkoom:
I was following the Azure quickstart docs but had trouble getting it to work.
The error I got was the following:
```
Error: unable to build authorizer for Resource Manager API: could not configure AzureCli Authorizer: could not parse Azure CLI version: launching Azure CLI: exec: "az": executable file not found in $PATH
```
Which seemed like an issue with Terraform after some superficial searching but not too sure what was going on.

I was using [client certificates for authorization](https://docs.upbound.io/providers/provider-azure/authentication/#create-a-service-principal-with-client-certificate-credentials-using-the-azure-cli-tool) and after looking at the provider versioning realized that this was only introduced in [v0.40.0](https://docs.upbound.io/providers/provider-azure/#v0400) and the version in the docs was way behind.

After using the latest version everything worked fine.

### DIFF
--- a/content/master/getting-started/provider-azure-part-2.md
+++ b/content/master/getting-started/provider-azure-part-2.md
@@ -45,7 +45,7 @@ kind: Provider
 metadata:
   name: provider-azure-network
 spec:
-  package: xpkg.upbound.io/upbound/provider-azure-network:v0.34.0
+  package: xpkg.upbound.io/upbound/provider-azure-network:v0.42.1
 EOF
 ```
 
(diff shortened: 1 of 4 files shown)


plumbis:
Thanks for the version bump, can you make the same changes in `/content/master` as well?

loomkoom:
I added the change to `/content/master`

plumbis (approved):
Excellent, thanks again!

PR #744, opened 2024-04-08
Update roadmap link

plumbis:
Updates the roadmap link per https://github.com/crossplane/crossplane/pull/5546/files


jbw976 (approved):
Thanks for catching this @plumbis, I forgot about updating it here too! 🙇‍♂️

PR #748, opened 2024-04-10
Corrects usage of Trim Prefix/Suffix. 

### LINKED ISSUE #747: The descriptions for `TrimPrefix` and `TrimSuffix` transforms are not accurate
### URL

* https://docs.crossplane.io/latest/concepts/patch-and-transform/#trim-prefix
* https://docs.crossplane.io/latest/concepts/patch-and-transform/#trim-suffix

### What's Wrong?

I believe that the descriptions for these transforms do not accurately describe how these transforms are intended to work. Basically, if the string doesn't start with the prefix or end with the suffix, then these transforms do nothing.

See 
https://github.com/crossplane/crossplane/issues/4722#issuecomment-2047847698 for a more detailed explanation of how (I think) they are actually supposed to work.

plumbis:
Resolves #747

### DIFF
--- a/content/v1.13/concepts/patch-and-transform.md
+++ b/content/v1.13/concepts/patch-and-transform.md
@@ -1573,7 +1573,7 @@ patches:
 #### Trim suffix
 
 The {{<hover label="typeTrimS" line="8">}}type: TrimSuffix{{</hover>}} uses 
-Go's [TrimSuffix](https://pkg.go.dev/strings#TrimPrefix) and removes characters 
+Go's [[TrimSuffix](https://pkg.go.dev/strings#TrimSuffix)](https://pkg.go.dev/strings#TrimSuffix) and removes characters 
 from the end of a line.
 
 ```yaml {label="typeTrimS"}
--- a/content/v1.13/concepts/patch-and-transform.md
+++ b/content/v1.13/concepts/patch-and-transform.md
@@ -1573,7 +1573,7 @@ patches:
 #### Trim suffix
 
 The {{<hover label="typeTrimS" line="8">}}type: TrimSuffix{{</hover>}} uses 
-Go's [[TrimSuffix](https://pkg.go.dev/strings#TrimSuffix)](https://pkg.go.dev/strings#TrimSuffix) and removes characters 
+Go's [TrimSuffix](https://pkg.go.dev/strings#TrimSuffix) and removes characters 
 from the end of a line.
 
 ```yaml {label="typeTrimS"}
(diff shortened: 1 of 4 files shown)


jbw976 (commented):
Should this be fixed in `master` too?
https://github.com/crossplane/docs/blob/master/content/master/concepts/patch-and-transform.md

content/v1.13/concepts/patch-and-transform.md
```diff
@@ -1571,10 +1572,11 @@ patches:
 
 #### Trim suffix
 
-The {{<hover label="typeRegex" line="8">}}type: TrimSuffix{{</hover>}} removes
-the matching string and all proceeding characters. 
+The {{<hover label="typeTrimS" line="8">}}type: TrimSuffix{{</hover>}} uses 
+Go's [TrimSuffix](https://pkg.go.dev/strings#TrimPrefix) and removes characters
```
jbw976, line 1576: this should link to the TrimSuffix function - i think this problem exists for all versions

plumbis:
🤦‍♂️ I fixed it and forgot to add it to the commit. Done.

phisco (approved):

content/v1.13/concepts/patch-and-transform.md
```diff
@@ -1571,10 +1572,11 @@ patches:
 
 #### Trim suffix
 
-The {{<hover label="typeRegex" line="8">}}type: TrimSuffix{{</hover>}} removes
-the matching string and all proceeding characters. 
+The {{<hover label="typeTrimS" line="8">}}type: TrimSuffix{{</hover>}} uses 
+Go's [[TrimSuffix](https://pkg.go.dev/strings#TrimSuffix)](https://pkg.go.dev/strings#TrimSuffix) and removes characters
```
jbw976, line 1576:
```suggestion
Go's [TrimSuffix](https://pkg.go.dev/strings#TrimSuffix) and removes characters
```

this formatting is not like the others and is causing the URL to be displayed on the page. This is the only instance where this is an issue.

Screenshot 2024-04-11 at 10 28 01 AM

jbw976 (changes requested):
Everything looks good except for the link on the v1.13 page 🙏

content/v1.13/concepts/patch-and-transform.md
plumbis: fixed!

jbw976 (approved):
awesome, looks good to me now @plumbis!! 💪

PR #752, opened 2024-04-12
Expand MR Status Condition Type details

### LINKED ISSUE #750: Provide details about standard conditions used by Crossplane
### URL

https://docs.crossplane.io/latest/concepts/managed-resources/#conditions

### What's Wrong?

Discussed a bit in this PR: https://github.com/crossplane/crossplane/pull/5453#discussion_r1562242295

Our docs focus on the possible specific reasons for each condition, but don't seem to focus on the high level conditions themselves, e.g. `Ready` and `Synced`. What are their intent and semantics, what scenarios do they cover?

@negz quite write-up that is useful:

> Today we have the following status conditions:
>
> * Ready - The external resource is ready to use, (where applicable) according to its API.
> * Synced - The managed resource's desired state has been successfully applied to the external resource.

plumbis:
This PR resolves #750, adding details on the Condition types.

I also added `hover` tags for all status reasons.

### DIFF
--- a/content/master/concepts/managed-resources.md
+++ b/content/master/concepts/managed-resources.md
@@ -839,60 +839,93 @@ Providers may define their own custom `Conditions`.
 {{</hint >}}
 
 
-### Available
-`Reason: Available` indicates the Provider created the managed resource and it's
-ready for use. 
+### Type: Ready
 
-```yaml {copy-lines="none"}
+Condition `Type: Ready` indicates if the external resource is ready to use.
+
+The Condition `Type: Ready` and `Status: True` indicates Provider created
+the external resource and notified Crossplane it's ready to use.
+
+{{<hint "important" >}}
+Crossplane doesn't update `Status: True` during a resource update.
+{{< /hint >}}
+
+The Condition `Type: Ready` and `Status: False` indicates the external resource
+isn't available. 
+
+
+#### Available
+{{<hover label="available" line="4">}}Reason: Available{{</hover>}} indicates 
+the Provider created the managed resource and it's ready for use. 
+
+```yaml {copy-lines="none",label="available"}
 Conditions:
   Type:                  Ready
   Status:                True
   Reason:                Available
 ```
-### Creating
 
-`Reason: Creating` indicates the Provider is attempting to create the managed
-resource. 
+#### Creating
 
-```yaml {copy-lines="none"}
+{{<hover label="creating" line="4">}}Reason: Creating{{</hover>}} indicates the 
+Provider is attempting to create the managed resource. 
+
+```yaml {copy-lines="none",label="creating"}
 Conditions:
   Type:                  Ready
   Status:                False
   Reason:                Creating
 ```
 
-### Deleting
-`Reason: Deleting` indicates the Provider is attempting to delete the managed
-resource. 
+#### Deleting
+{{<hover label="deleting" line="4">}}Reason: Deleting{{</hover>}} indicates the 
+Provider is attempting to delete the managed resource. 
 
-```yaml {copy-lines="none"}
+```yaml {copy-lines="none",label="deleting"}
 Conditions:
   Type:                  Ready
   Status:                False
   Reason:                Deleting
 ```
 
-<!-- vale off -->
-### ReconcilePaused
-<!-- vale on -->
-`Reason: ReconcilePaused` indicates the managed resource has a [Pause](#paused)
-annotation 
+#### Unavailable
+{{<hover label="unavailable" line="4">}}Reason: Unavailable{{</hover>}}
+indicates Crossplane expects the managed resource to be available, but the 
+Provider reports the resource is unhealthy.
 
-```yaml {copy-lines="none"}
+```yaml {copy-lines="none",label="unavailable"}
 Conditions:
-  Type:                  Synced
+  Type:                  Ready
   Status:                False
-  Reason:                ReconcilePaused
+  Reason:                Unavailable
 ```
 
+### Type: Synced
+
+Condition `Type: Synced` indicates Crossplane has checked the status of the managed
+resource with the Provider.
+
+The condition `Type: Synced` and `Status: True` means Crossplane successfully
+communicated with the Provider on the status of the manged resource. 
+
+{{< hint "note" >}}
+`Type: Synced` and `Status: True` only shows success between Crossplane and 
+the Provider. This status doesn't mean the Provider successfully changed the
+external resource.
+{{< /hint >}}
+
+The condition `Type: Synced` and `Status: False` means Crossplane failed to
+update or determine the current state of the managed resource.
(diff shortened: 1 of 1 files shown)


content/master/concepts/managed-resources.md
```diff
@@ -839,96 +839,139 @@ Providers may define their own custom `Conditions`.
...
+resource with the Provider.
+
+The condition `Type: Synced` and `Status: True` means Crossplane successfully
+communicated with the Provider on the status of the manged resource. 
+
+{{< hint "note" >}}
+`Type: Synced` and `Status: True` only shows success between Crossplane and 
+the Provider. This status doesn't mean the Provider successfully changed the
+external resource.
+{{< /hint >}}
```
negz, line 915: This isn't what the synced condition represents. 🙂 

The synced condition represents whether a resource's controller made it through the most recent reconcile without hitting an error.

This this context, this means the synced condition represents whether the provider successfully synced the MR's desired state (spec) with the external system, and successfully synced the observed state of the external system back to the MR (status).

negz: Synced means the same for claims and XRs, at their level of abstraction.

So a claim is synced when its controller successfully syncs the claim's desired state with an XR. An XR is synced when its controller successfully syncs its desired state with its composed resources.

negz (commented):
Thanks @plumbis!

My comments might be too much detail for the docs, so take from them whatever you think is appropriate.

Another note of potential interest: readiness is cumulative but synced-ness is not.

A claim only becomes ready when its XR and its entire tree of composed resources (including nested XRs and their composed resources) become ready.

On the other hand, a claim becomes synced as soon as it applies it desired state to the XR, even if that state is nonsensical and results in a broken XR. (The claim controller still did what it was asked to without hitting an error.) Ditto with XRs and MRs - as long as they successfully do what you told them they'll be synced.

In general (across claims, XRs, and MRs) you can think of "synced" as meaning "Crossplane was able to do what you told it to do". "Ready" means "what you told Crossplane to do resulted in (apparently) working infrastructure".

plumbis:
Thanks @negz ! I updated those two sections, let me know if you're good with this or have any other feedback. When we're good I'll copy this to the v1.15 folder as well.

content/master/concepts/managed-resources.md
```diff
@@ -839,96 +839,143 @@ Providers may define their own custom `Conditions`.
...
+  Reason:                Unavailable
 ```
 
+### Type: Synced
+
+Condition `Type: Synced` indicates Crossplane has checked the status of the managed
+resource with the Provider.
+
+The condition `Type: Synced` and `Status: True` means Crossplane successfully
+communicated with the Provider on the status of the manged resource and synced
```
negz, line 911:
```suggestion
communicated with the Provider on the status of the managed resource and synced
```

Typo in managed.

content/master/concepts/managed-resources.md
```diff
@@ -839,96 +839,143 @@ Providers may define their own custom `Conditions`.
...
+the managed resource's desired state with the observed state of the external
+resource.
```
negz, line 913: I'd suggest leaving out "Crossplane communicated with the provider". I think it's misleading.

There's really no communication between Crossplane and the provider here, everything is done entirely by the provider.

You could either just talk about what the provider did, without mentioning Crossplane, or simplify and talk about what Crossplane did. I'd prefer the former, but the latter also makes sense if I think of "Crossplane" as a whole system, rather than a specific component.

Edit: Or by "Provider" do you mean "external system"? The capital P in Provider makes me think you're talking about the component of Crossplane, not the external system,

content/master/concepts/managed-resources.md
```diff
@@ -839,96 +839,143 @@ Providers may define their own custom `Conditions`.
...
+  Type:                  Ready
   Status:                False
-  Reason:                ReconcilePaused
+  Reason:                Unavailable
 ```
 
+### Type: Synced
+
+Condition `Type: Synced` indicates Crossplane has checked the status of the managed
+resource with the Provider.
```
negz, line 908: Is there a verb other than "checked" that could work here?

Checked reads as if "syncing" a resource is a read-only operation, but it isn't.

I like "update or determine" below if that's not too unwieldy here. So "Crossplane has updated and determined the status of the managed resource".

plumbis: How about "validated the status of the managed resource..."?

I'm trying to break the Condition and Status into two parts, where the Condition is "we tried to inspect at the MR" and the status is "inspection worked and now update if needed". Because of that I want to avoid "updated" in the first paragraph.

negz: > I'm trying to break the Condition and Status into two parts, where the Condition is "we tried to inspect at the MR" and the status is "inspection worked and now update if needed".

Breaking it into two parts doesn't make sense to me. The status condition represents the entire reconcile operation, not only the inspection of the MR.

Maybe writing out what happens in pseudocode would help explain this better:

```python
def sync_resource():
  desired = read_desired_state()
  actual = read_actual_state()
  if actual != desired:
    update_actual_state(desired)

def main():
  while True:
    try:
      sync_resource()
    except SyncException:
     set_condition_synced_false()
    else:
     set_condition_synced_true()
    sleep(60)
```

Most reconcile loops boil down to read desired state, read actual state, update actual state to match desired state. If __all__ of that works we set `Synced: true`. If __any__ of that fails we set `Synced: false`.

plumbis: Got it.

When does a resource move from Ready/True to Synced/True? On the first reconcile after it's ready?

negz: > When does a resource move from Ready/True to Synced/True?

The two condition types (Synced and Ready) are independent of each other, so I wouldn't think a resource moving from one to the other.

That said, Synced: True is usually going to happen first, because for Synced to be True the controller just needs to get through one reconcile loop without hitting an error.

Let's say it's the very first reconcile of an MR, and the code managed to observe the external resource, notice it doesn't exist, and create it. Synced would be True because the controller just "synced" the desired state. In this case, by creating the external resource.

Often Ready: True would happen a bit later, because many MR controllers observe the external resource to see whether it appears to be ready (according to the API). So at that point the MR would be Synced: True and Ready: True.

negz: Expanding on the pseudocode to include the ready condition:

```python
def sync_resource():
  desired = read_desired_state()
  actual = read_actual_state()
  if actual != desired:
    update_actual_state(desired)
  return actual

def main():
  while True:
    try:
      actual_state = sync_resource()
      if is_ready(actual_state):
        set_condition_ready_true()
      else:
        set_condition_ready_false()
    except SyncException:
     set_condition_synced_false()
    else:
     set_condition_synced_true()
    sleep(60)
```

PR #757, opened 2024-04-17
API Docs Backend Refactor

plumbis:
This PR is a modularizes the API docs generation code that is in place today. This will make maintenance way easier for anyone that isn't me. 

This also moves the CRD directory from `/content/<version>/api/yaml` to the more obvious `/content/<version>/api/crds`.

This also fixes existing bugs found during development around processing and displaying enum data types.

(this is a revamp of #737 without the description files changes)


jbw976 (approved):
Tested this one through fairly thoroughly on the preview site: https://deploy-preview-757--crossplane.netlify.app/v1.15/api/

all functionality appears to be intact and behaving well. Even noticed a couple improvements to the way we render things too.  I don't see any major regressions from the refactor, so looks good to ship to me!

jbw976:
@plumbis consider squashing your commits as there are definitely some of the 40 commits here that aren't necessary to tell a complete story in the commit history 😉

PR #764, opened 2024-05-07
fix: broken links

cr7258:
The URLs are not enclosed by bracket "()".

image
-->
image


jeanduplessis (approved):
Thank you for these fixes.

Issue #765, opened 2024-05-07
version selector is missing


jbw976:
### URL

Basically any page in the docs that has multiple versions of the content, e.g. https://docs.crossplane.io/v1.15/concepts/compositions/ 

### What's Wrong?

There used to be a version selector dropdown that would enable you to select which version of the content you want to see, e.g. v1.15, v1.14, or v1.13.

This dropdown is completely missing (as pointed out by @jeanduplessis), so users cannot select their desired version in the UI. Perhaps this was accidentally removed in some recent refactors of the site structure.

#### Workaround

Manually edit the URL, e.g. to  https://docs.crossplane.io/v1.14/concepts/compositions/

PR #767, opened 2024-05-07
Update version dropdown

tr0njavolta:
Fixes https://github.com/crossplane/docs/issues/765

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -91,6 +91,7 @@ security:
 params:
   # The current "latest" version. Used in the version dropdown
   latest: "1.15"
+  docs: true 
   anchors:
     # Generate heading anchors for any heading between min and max
     min: 2


jbw976 (approved):
awesome, this looks like it works on the preview site! thanks for quickly addressing this @tr0njavolta and @plumbis for the hints! 💪 

one thought: going forward, it'll be important for everyone to easily understand how the functionality on this docs site works. Is this `docs` bool value documented somewhere? if not, is it worth adding a comment to this file that says what this value does? apparently it was easy to accidentally delete it, perhaps we can make it more obvious what it does 😅

jbw976:
Thanks @plumbis for the explanation!! Let's go ahead and get this merged to fix this issue in production 🙏

PR #773, opened 2024-05-15
v1.16 Docs

tr0njavolta:
https://github.com/crossplane/docs/issues/772

Updates master with version 1.16

- [x] Update the $LATEST_VER parameter in [netlify_build.sh](https://github.com/crossplane/docs/blob/master/netlify_build.sh#L3)
- [x] Update params.latest in [config.yaml](https://github.com/crossplane/docs/blob/master/config.yaml#L93)
- [x] Update version in the _index.md file of /content/<new latest> from master to the correct version.
- [x] Copy Crossplane [cluster/crds](https://github.com/crossplane/crossplane/tree/master/cluster/crds) contents to /content/<new latest>/api/crds.
- [x] Create a [new release/tag](https://github.com/crossplane/docs/releases/new) named "v-archive" to snapshot EOL'd docs.
- [x] Remove EOL'd docs version from "/content" directory and run hugo locally to check for broken links.
- [ ] Trigger [Algolia Crawler](https://crawler.algolia.com/) after publishing to reindex results.

PR #774, opened 2024-05-15
Update composition-functions.md

joebowbeer:
Fix several nits in the composition functions doc.

Also fixed a couple typos in the vale config file comments.

### DIFF
--- a/content/master/concepts/composition-functions.md
+++ b/content/master/concepts/composition-functions.md
@@ -249,7 +249,7 @@ spec:
 
 {{<expand "The xr.yaml, composition.yaml and function.yaml files used in the example">}}
 
-You can recreate the output below using by running `crossplane beta render` with
+You can recreate the output below by running `crossplane beta render` with
 these files.
 
 The `xr.yaml` file contains the composite resource to render:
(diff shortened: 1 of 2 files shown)


plumbis (approved):

plumbis:
Let's wait for the 1.16 docs to publish in #773 before merging

joebowbeer:
@plumbis PTAL - I improved wording to address a passive voice lint check

joebowbeer:
Please overrule the lint check. It is not helping in this case.

jbw976:
> Please overrule the lint check. It is not helping in this case.

This can be done within your changes as described in the contributing guide: https://docs.crossplane.io/contribute/vale/#ignore-vale-rules

want to give that a try with another commit @joebowbeer?

joebowbeer:
@jbw976 PTAL - lint rule disabled

By the way, as an author and technical reviewer, I don't think I could function without passive voice.

I think Robert Delwood makes some good points in this article:

[There is No You in API: A Case for Passive Voice in API Documentation](https://robertdelwood.medium.com/there-is-no-you-in-api-856a63e8084e)

> Use active voice where it makes sense. But understand that passive voice isn’t a grammatical error; it’s a matter of style. It’s your friend when the thing receiving an action is the important part of the sentence — especially in scientific and legal contexts — when the performer of an action is unknown or in cases where the performer distracts or is irrelevant.

jbw976:
Thanks @joebowbeer for both the contribution and the good article about passive voice! that was an interesting read. it's good to have the option to use passive voice when it makes more sense and it looks like there are a good number of those exceptions in this docs repo already.  the more writer/author inclined maintainers on this repo can consider if more relaxed rules would be better for a new default, but I have a feeling that keeping the default as is and allowing exceptions when needed is a reasonable compromise :)

PR #790, opened 2024-06-30
add unknown field check to validate command

enesonus:
Adds the improvements made at PR https://github.com/crossplane/crossplane/pull/5791 to `beta validate` command's reference docs for informational purposes.

### DIFF
--- a/content/master/cli/command-reference.md
+++ b/content/master/cli/command-reference.md
@@ -828,7 +828,8 @@ Configuration/platform-ref-aws                             v0.9.0    True       
 
 The `crossplane beta validate` command validates 
 [compositions]({{<ref "../concepts/compositions">}}) against provider or XRD 
-schemas using the Kubernetes API server's validation library.
+schemas using the Kubernetes API server's validation library
+with additional validation checks such as Unknown Field Check.
 
 The `crossplane beta validate` command supports validating the following 
 scenarios:
--- a/content/master/cli/command-reference.md
+++ b/content/master/cli/command-reference.md
@@ -829,7 +829,8 @@ Configuration/platform-ref-aws                             v0.9.0    True       
 The `crossplane beta validate` command validates 
 [compositions]({{<ref "../concepts/compositions">}}) against provider or XRD 
 schemas using the Kubernetes API server's validation library
-with additional validation checks such as Unknown Field Check.
+with additional validation such as checking for unknown fields,
+a common source of difficult to debug issues in Crossplane.
 
 The `crossplane beta validate` command supports validating the following 
 scenarios:
--- a/content/master/cli/command-reference.md
+++ b/content/master/cli/command-reference.md
@@ -829,7 +829,7 @@ Configuration/platform-ref-aws                             v0.9.0    True       
 The `crossplane beta validate` command validates 
 [compositions]({{<ref "../concepts/compositions">}}) against provider or XRD 
 schemas using the Kubernetes API server's validation library
-with additional validation such as checking for unknown fields,
+with extra validation such as checking for unknown fields,
 a common source of difficult to debug issues in Crossplane.
 
 The `crossplane beta validate` command supports validating the following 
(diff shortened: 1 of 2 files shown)


content/master/cli/command-reference.md
```diff
@@ -828,7 +828,8 @@ Configuration/platform-ref-aws                             v0.9.0    True
 
 The `crossplane beta validate` command validates 
 [compositions]({{<ref "../concepts/compositions">}}) against provider or XRD 
-schemas using the Kubernetes API server's validation library.
+schemas using the Kubernetes API server's validation library
+with additional validation checks such as Unknown Field Check.
```
jbw976, line 832: I don't think Unknown Field Check is a proper noun, so no need to capitalize it. How about this suggestion that may flow more naturally?
```suggestion
with additional validation such as checking for unknown fields, a common source of difficult to debug issues in Crossplane.
```

content/v1.16/cli/command-reference.md
```diff
@@ -828,7 +828,8 @@ Configuration/platform-ref-aws                             v0.9.0    True
 
 The `crossplane beta validate` command validates 
 [compositions]({{<ref "../concepts/compositions">}}) against provider or XRD 
-schemas using the Kubernetes API server's validation library.
+schemas using the Kubernetes API server's validation library
```
jbw976, line 831: Something I realized since we synced on Monday is that we should only make this docs change if the code change itself is going to be back ported to previous releases in a patch release (e.g. a v1.16.1).

We'll wait to ship this functionality in v1.17, so there's no need to make any docs changes for anywhere but in the `master` directory. You can remove this change 😇

jbw976 (changes requested):
Thanks for taking the initiative to make a docs update too @enesonus! a couple of suggestions for you to consider 🙇‍♂️

enesonus:
> Thanks for taking the initiative to make a docs update too @enesonus! a couple of suggestions for you to consider 🙇‍♂️

Thanks for the suggestions, I updated the changes accordingly.

content/master/cli/command-reference.md
```diff
@@ -828,7 +828,9 @@ Configuration/platform-ref-aws                             v0.9.0    True
 
 The `crossplane beta validate` command validates 
 [compositions]({{<ref "../concepts/compositions">}}) against provider or XRD 
-schemas using the Kubernetes API server's validation library.
+schemas using the Kubernetes API server's validation library
+with additional validation such as checking for unknown fields,
```
jbw976, line 832: ha, looks like my suggestion introduced a Vale error in https://github.com/crossplane/docs/actions/runs/9758986765/job/26934740522?pr=790.  
```
content/master/cli/command-reference.md
 832:[6](https://github.com/crossplane/docs/actions/runs/9758986765/job/26934740522?pr=790#step:5:7)  warning  'additional' is too wordy.  write-good.TooWordy
```

Vale is intended to help us write in a high quality consistent format, but its rules can be difficult to comply with sometimes 🤓 

I think this is the rule that is failing:
https://github.com/crossplane/docs/blob/master/utils/vale/styles/write-good/TooWordy.yml#L17

`additional` is also mentioned in https://github.com/crossplane/docs/blob/master/utils/vale/styles/Microsoft/ComplexWords.yml#L21, which suggests using `more` or `extra` instead.  So perhaps to make Vale happy we could just use `extra` instead of `additional` here? 🤔

enesonus: Yeah I was actually aware of the warning, thought it is Vale's own warning and couldn't find the location to look for alternatives. Thanks for pointing out the guides @jbw976. I changed `additional` to `extra`. One question is should I leave it as `with extra validation` or change it to `with extra validations` what do you think? I think the latter one sounds better but we dont have more than one extra validation currently so I couldnt decide :sweat_smile:

jbw976: `with extra validation` sounds good enough to my ears! 👂 ✅

jbw976 (approved):
awesome, this looks great to me now! Thanks @enesonus!!

PR #792, opened 2024-07-04
Fix compositionrevision docs

ezgidemirel:
Description and context provided in https://github.com/crossplane/docs/pull/792#pullrequestreview-2266770668

### DIFF
--- a/content/master/concepts/composition-revisions.md
+++ b/content/master/concepts/composition-revisions.md
@@ -4,7 +4,7 @@ weight: 35
 ---
 
 This guide discusses the use of "Composition Revisions" to safely make and roll
-back changes to a Crossplane [`Composition`][composition-type]. It assumes
+back changes to a Crossplane [`Composition`][composition type]. It assumes
 familiarity with Crossplane, and particularly with
 [Compositions].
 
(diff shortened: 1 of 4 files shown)


negz:
@ezgidemirel Can you add some context around what this is fixing in the description please?

Some of the new changes are failing the Vale grammar linter (see the Checks tab). I suspect not all the warnings were actually introduced by this PR, but it may be worth trying to address some.

jbw976 (commented):
I believe the context for this PR was to restore lost content about `CompositionRevisions`, as first noticed in https://github.com/crossplane/crossplane/issues/2882#issuecomment-2208222232. After some investigation, it looks like [previous edits](https://github.com/crossplane/crossplane/pull/3481/files#diff-be7bd746bfeae28b06d16ba11d7e31c64157c24299b7fb615696ff3d7c929c02) to add this same content in the past had been lost awhile back due to challenges associated with the branching/releasing strategies used in this repo 🪦

From that investigation, we also determined that this was the only affected and still current content that would need to be restored.

Does this need to be updated for the functions first docs approach in https://github.com/crossplane/docs/pull/799/files#diff-2008ef6855bc5d20d1ee7f4cba4f5e7934943c65bf56f3cdb8de4717bd02eb67? I think it does not need to be. 

@ezgidemirel if you could address any vale errors for new content here, that would be great! I wouldn't bother addressing any errors for stuff you're not changing.

ezgidemirel:
> @ezgidemirel Can you add some context around what this is fixing in the description please?
> 
> Some of the new changes are failing the Vale grammar linter (see the Checks tab). I suspect not all the warnings were actually introduced by this PR, but it may be worth trying to address some.

As Jared mentioned, these changes were reverted accidentally some time ago. 

I've fixed some of the vale errors. The remaining ones are spelling checks for words like  `CompositionRevisions`, `revision:1` etc..

jbw976 (commented):
d'oh, now that we've released v1.17 @ezgidemirel, would you mind updating this PR to also update v1.17 content and no longer update v1.14 content? 😇

jbw976 (approved):
Thanks for seeing this through @ezgidemirel! 🙇‍♂️

PR #793, opened 2024-07-08
Fix indent in composition-functions.md

Moulick:
Minor fixing indent in all versions of `composition-functions.md`

PR #794, opened 2024-07-08
Cleanup spaces in `composition-functions.md`

Moulick:


plumbis:
I'm hesitant to approve a whitespace cleanup PR. End of line whitespace can have meaning in markdown, creating a line return. It makes this kind of PR hard to review.

I know most of these are single space, but there is at least one instance where it's removing a line return. 

Unless there's a compelling reason to do this work I'd rather close out this PR.

Moulick:
> I'm hesitant to approve a whitespace cleanup PR. End of line whitespace can have meaning in markdown, creating a line return. It makes this kind of PR hard to review.
> 
> I know most of these are single space, but there is at least one instance where it's removing a line return.
> 
> Unless there's a compelling reason to do this work I'd rather close out this PR.

Thanks for the review. As far as I found, a _single_ space at the end of line does not have any meaning but double space does. The one where I see it's broken a link return is 
> Function Patch and Transform implements Crossplane resource
templates.  
The input kind is 

I can go over and compare the current rendered docs vs the whitespace removed version and add the proper breaks using <br> wherever the intended line break currently are.

The reason I open this PR is while making the previous PR #793, I had to do through three different markdown editors to make them not format the text on save. Finally, only the macOS built-in TextEdit was used to make the fix as it does not do any formatting at all 🤣. So, the main intention is to prevent this problem for any contributors to the docs and make them easier to contribute ❤️

Issue #791, opened 2024-07-03
Docs should take a "functions first" approach


jbw976:
### What's Missing?

Functions are the future of Crossplane's experience and how platform engineers should be composing resources. As we move towards making Functions [GA](https://github.com/crossplane/crossplane/issues/4951), and even plan on deprecating classic patch and transform (https://github.com/crossplane/crossplane/issues/4746), we should be making them more of the "default" path when using Crossplane. New users of Crossplane should be using Functions from the very start of their Crossplane journey.

The most obvious place to update with a "functions first" approach would be the quick start guides that are still currently using classic patch and transform, e.g.:

* https://docs.crossplane.io/v1.16/getting-started/provider-aws-part-2/#create-a-custom-api

Below will track a list of broad areas that we should be updating to use and evangelize Functions more aggresively:

```[tasklist]
### Topics to update
- [ ] https://docs.crossplane.io/v1.16/getting-started/provider-aws-part-2/
- [ ] https://docs.crossplane.io/master/getting-started/provider-azure-part-2/ 
- [ ] https://docs.crossplane.io/master/getting-started/provider-gcp-part-2/
- [ ] https://docs.crossplane.io/master/concepts/compositions/ 
- [ ] https://docs.crossplane.io/master/concepts/patch-and-transform/
```

bobh66:
The existing Composition Functions documentation needs to include more detailed information regarding the importance of the `crossplane.io/composition-resource-name` annotation and how it affects reconciliation.  Specifically:
- The `composition-resource-name` MUST be unique within the composition/composite instance
- The `composition-resource-name` MUST be static for a given resource.  If the name changes it will cause the existing resource (with the old name) to be deleted and a new resource will be created, even if nothing else in the resource changes.
- The `composition-resource-name` SHOULD NOT use a loop index counter as part of the name (bucket-0, bucket-1) unless it is guaranteed that the loop will not reshuffle the indices and cause the names to change

PR #799, opened 2024-08-17
Promote composition functions to v1

negz:
This is related to https://github.com/crossplane/docs/issues/791 and https://github.com/crossplane/crossplane/pull/5885.

I don't intend to truly make our docs "functions first" with this PR. Instead the plan is to take the first baby steps. Those will be:

- [x] Update all `Composition` examples to use function-patch-and-transform
- [x] Ensure all the getting started guides mention installing function-patch-and-transform
- [x] Remove the notices that functions are beta
- [x] Move everything patch-and-transform related to a guide on function-patch-and-transform

I'd also like to add a brief guide on migrating a function from v1beta1 to v1 (it's basically `s/v1beta1/v1` on the imports), but I think I'll do that in a separate PR in the interest of getting this merged and unblocking the Crossplane release.

### Reviewer notes

This also removes the release notes from the docs. They're still available at https://github.com/crossplane/crossplane/releases. See the relevant commit message for details.

Vale is failing, but only for content I didn't touch in this PR. I'm guessing the problem is that I touched pages with pre-existing content that doesn't pass the linter.


content/master/cli/command-reference.md
```diff
@@ -32,6 +32,135 @@ Client Version: v1.16.0
...
+
+### Include function results
+
+If a function produces Kubernetes events with statuses use the 
+`--include-function-results` to print them along with the managed resource 
+outputs. 
+
+### Include the composite resource 
+
+Composition functions can only change the `status` field of a composite
```
bobh66, line 126: I'll do another full pass through but I think this limitation needs to be called out more specifically at a higher level.  It was common to use metadata.annotations and labels to store information, and there were examples in the docs showing this pattern, so breaking that pattern needs to be more clearly documented.   This also has impacts on the migration from legacy P&T to functions, since the existing compositions that patch into metadata need to be rewritten to use status instead.

content/master/cli/command-reference.md
```diff
@@ -341,9 +516,8 @@ converts a Crossplane resource to a new version or kind.
 Use the `crossplane beta convert` command to convert an existing
 [ControllerConfig]({{<ref "../concepts/providers#controller-configuration">}})
 to a [DeploymentRuntimeConfig]({{<ref "../concepts/providers#runtime-configuration">}}) 
-or a Composition using [patch and transforms]({{<ref "../concepts/patch-and-transform">}}) 
-to a 
-[Composition pipeline function]({{< ref "../concepts/compositions#use-composition-functions" >}}).
+or a legacy Composition using `mode: Resources` to a 
+[Composition pipeline function]({{< ref "../concepts/compositions" >}}).
```
bobh66, line 520: This should probably make clear that legacy Compositions that patch information into the XR `metadata` will not work properly after conversion and will need to be updated to use `status` instead.

content/master/cli/command-reference.md
negz: I definitely agree, but I'd rather raise a tracking issue for that and follow up ASAP rather than tackle it in this PR.

To get this PR finished as quickly as possible I'd like to keep it focused on the promotion to GA. I'm only adding new content where it pertains to the fact that functions are now GA, and native P&T is deprecated.

bobh66:
Tagging crossplane/crossplane#5434 here as a bread crumb

content/master/cli/command-reference.md
```diff
@@ -32,6 +32,135 @@ Client Version: v1.16.0
...
 ```
 
+## render 
+
+The `crossplane render` command previews the output of a 
+[composite resource]({{<ref "../concepts/composite-resources">}}) after applying 
+any [composition functions]({{<ref "../concepts/compositions">}}).
+
+{{< hint "important" >}}
+The `crossplane render` requires you to use composition functions.
```
jeanduplessis, line 42:
```suggestion
The `crossplane render` command requires you to use composition functions.
```

content/master/cli/command-reference.md
```diff
@@ -32,6 +32,135 @@ Client Version: v1.16.0
...
+| Short flag   | Long flag                             | Description                                           |
+| ------------ | -------------                         | ------------------------------                        |
+|              | `--context-files=<key>=<file>,<key>=<file>`    | A comma separated list of files to load for function "contexts." |
+|              | `--context-values=<key>=<value>,<key>=<value>` | A comma separated list of key-value pairs to load for function "contexts."                                                    |
+| `-r`         | `--include-function-results`          | Include the "results" or events from the function.   |
+| `-o`         | `--observed-resources=<directory or file>`               |
+Provide artificial managed resource data to the function.
+|
+| `-x`         | `--include-full-xr`          | Include a copy of the input Composite Resource spec and metadata fields in the rendered output.   |
+|              | `--timeout=`                          | Amount of time to wait for a function to finish.                    |
```
jeanduplessis, line 104: Is there a default value here that's worth listing?

content/master/cli/command-reference.md
```diff
@@ -32,6 +32,135 @@ Client Version: v1.16.0
...
+{{< /table >}}
+
+The `crossplane render` command relies on standard 
+[Docker environmental variables](https://docs.docker.com/engine/reference/commandline/cli/#environment-variables) 
+to connect to the local Docker engine and run composition functions.
```
jeanduplessis, line 109:
```suggestion
to connect to the local Docker Engine and run composition functions. 
```

Capitalizing Docker `E`ngine for consistency.

content/master/concepts/compositions.md
```diff
@@ -29,385 +65,210 @@ Crossplane has four core components that users commonly mix up:
...
-            key: creds
-```
+<!-- vale write-good.Passive = NO -->
+The `Resources` mode is deprecated, and you shouldn't use it. Crossplane
+supports Compositions that use the `Resources` mode for backward compatibility,
+but the feature is no longer maintained. Crossplane doesn't accept new
+`Resources` features, and only accepts security bug fixes.
+<!-- vale write-good.Passive = YES -->
 
+See the [CLI documentation]({{<ref "../cli">}}) to learn how to use the
```
jeanduplessis, line 110:
```suggestion
See the [CLI documentation]({{<ref "../cli/command-reference/#beta-convert">}}) to learn how to use the
```

jeanduplessis (approved):
Just a few nits, nothing blocking. LGTM 👍

content/master/concepts/compositions.md
```diff
@@ -29,385 +65,210 @@ Crossplane has four core components that users commonly mix up:
...
-Compositions allow applying a `metadata.name` to a resource's 
-{{<hover label="resources" line="16">}}spec{{</hover>}} but ignores it. The
-`metadata.name` field doesn't influence the name
-of the managed resource in Crossplane or the external resource inside the 
-Provider.
-
-Use the `crossplane.io/external-name` annotation on the resource to influence
-the external resource name.
+{{<hint "tip" >}}
+The Crossplane community has built lots of functions that let you template
```
turkenh, line 84: Should we call out https://github.com/crossplane-contrib/function-kcl as well considering its popularity.

content/master/concepts/compositions.md
```diff
@@ -29,385 +65,210 @@ Crossplane has four core components that users commonly mix up:
...
-the label
-{{<hover label="controller2" line="34">}}type: internet{{</hover>}} and only
-match objects in the same composite resource with 
-{{<hover label="controller2" line="32">}}matchControllerRef{{</hover>}}.
+Crossplane passes each Function in the pipeline the result of the previous
+Function. This enables powerful combinations of Functions. In this example,
+Crossplane calls {{<hover label="double" line="10">}}function-cue{{</hover>}} to
+create an S3 bucket. Crossplane then passes the bucket to 
+{{<hover label="double" line="23">}}function-auto-ready{{</hover>}}, which marks the
+composite resource as ready when the bucket becomes ready.
```
turkenh, line 242: nit: function-auto-ready could be confusing for someone learning Crossplane / compositions, e.g. why do I need to mark it to ready again with a separate function, if it is already ready :) It would demonstrate the idea better if the second function adds a common label/annotation etc.

turkenh (approved):

content/master/cli/command-reference.md
negz: These command reference docs are all just relocated existing content, so these were all pre-existing issues. They're easy to fix up though so I've pushed them in a separate commit.

content/master/concepts/compositions.md
negz: This content is pre-existing so I'm going to leave it as is.

content/master/concepts/compositions.md
negz: Done.

Issue #795, opened 2024-07-15
Document how functions can surface events/conditions on XRs/Claims


jbw976:
### What's Missing?

The ability for functions to bubble up events and status conditions to XR and/or claim was merged for v1.17 in https://github.com/crossplane/crossplane/pull/5450.

As @dalton-hill-0 describes in https://github.com/crossplane/crossplane/issues/4947#issuecomment-2192405501, we now have:

> * The ability for results to optionally emit events to the claim as well as the XR.
> * The ability to set status conditions on the XR and optionally the claim.

This feature should be documented so the community can understand what it does and how they can start using it within their own functions.

dalton-hill-0:
I identified a few spots to add this documentation. Please let me know what you think.
- https://github.com/crossplane/function-sdk-go/pull/160
- https://github.com/crossplane/function-template-go/pull/78
- https://github.com/crossplane/docs/pull/809

cc: @jbw976 @negz

jbw976:
https://github.com/crossplane-contrib/function-status-transformer is also related here ✅

jbw976:
All the locations in @dalton-hill-0's [list](https://github.com/crossplane/docs/issues/795#issuecomment-2327301214) have been merged, so I'll go ahead and close this issue now. Thanks @dalton-hill-0! 🙇‍♂️

PR #809, opened 2024-09-03
Claim Communication Examples

dalton-hill-0:
Adds examples of communicating to the user by setting status conditions and emitting events.

Related to https://github.com/crossplane/docs/issues/795.


negz (approved):

negz:
@dalton-hill-0 This looks great! Unfortunately this repo is using directories rather than git branches, so you need to make the same changes to `content/master`. Once that's done we can merge.

jbw976 (approved):
awesome @dalton-hill-0, thanks for making the updates for both v1.17 and master too!

PR #819, opened 2024-10-04
A few grammatical updates

olearycrew:

### DIFF
--- a/content/master/concepts/packages.md
+++ b/content/master/concepts/packages.md
@@ -6,7 +6,7 @@ weight: 200
 ---
 
 A _Configuration_ package is an 
-[OCI container images](https://opencontainers.org/) containing a collection of
+[OCI container image](https://opencontainers.org/) containing a collection of
 [Compositions]({{<ref "./compositions" >}}), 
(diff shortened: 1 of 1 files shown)


content/master/concepts/packages.md
```diff
@@ -437,7 +437,7 @@ Including YAML files that aren't Compositions or CompositeResourceDefinitions,
 including Claims isn't supported.
 {{</hint >}}
 
-By default, Crossplane creates an `.xpkg` file of the Configuration name and 
+By default, Crossplane creates a `.xpkg` file of the Configuration name and
```
jbw976, line 440: i didn't think this was right at first, but I guess it does actually work because it should be read as `creates a dot xpkg file`

jbw976 (approved):
nice @olearycrew, thanks for making these updates! They all seem reasonable enough to me.

I think we just need a DCO sign-off for your commit, here's info on that from the contributing guide: https://github.com/crossplane/crossplane/tree/main/contributing#certificate-of-origin

olearycrew:
@jbw976 sorry I missed the DCO - signed now!

PR #827, opened 2024-11-05
Simplify provider versions in examples

jeanduplessis:
Provider versions change constantly, and updating the examples in the docs with every release is not feasible.
Users of the docs sometimes end up installing very old versions of the providers, which results in a sub-optimal experience and the need for help.

This PR introduces two main changes:
1. The provider version to be installed is set to the major version only, e.g., v1 instead of v1.2.3. This has the effect of installing the latest provider version of the v1 major release. Note this capability is limited to the Upbound Marketplace for the providers used in the Quickstart guide.
2. Updates the example output where provider versions are listed all to v1.0.0 for consistency.

### DIFF
--- a/content/master/concepts/providers.md
+++ b/content/master/concepts/providers.md
@@ -339,7 +339,7 @@ For example, this installation of the Upbound AWS reference platform is
 ```shell {copy-lines="1"}
 kubectl get providers
 NAME              INSTALLED   HEALTHY   PACKAGE                                           AGE
-provider-aws-s3   True        False     xpkg.upbound.io/upbound/provider-aws-s3:v1.17.0   12s
+provider-aws-s3   True        False     xpkg.upbound.io/upbound/provider-aws-s3:v1.0.0   12s
 ```
 
 To see more information on why the Provider isn't `HEALTHY` use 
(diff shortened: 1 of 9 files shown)


content/master/concepts/providers.md
```diff
@@ -640,7 +640,7 @@ kind: Provider
 metadata:
   name: provider-gcp-iam
 spec:
-  package: xpkg.upbound.io/upbound/provider-gcp-iam:v0.37.0
+  package: xpkg.upbound.io/upbound/provider-gcp-iam:v1.0.0
```
jbw976, line 643: could the `v1` tag just be used here, since this is a manifest that the user would apply?

content/master/getting-started/provider-azure.md
```diff
@@ -39,7 +39,7 @@ kind: Provider
 metadata:
   name: provider-azure-network
 spec:
-  package: xpkg.upbound.io/upbound/provider-azure-network:v0.42.1
+  package: xpkg.upbound.io/upbound/provider-azure-network:v1.0.0
```
jbw976, line 42: same, could this be `v1`?

jbw976 (approved):
This looks like a nice experience update @jeanduplessis. A couple places to check if we could just use `v1` instead. If you look at those and confirm with testing that `v1` tag pulling works, then this looks good to me! 🙇‍♂️

content/master/concepts/providers.md
jeanduplessis: Yup, thanks for catching it.

PR #834, opened 2024-11-06
Fix case of deletionPolicy Delete

sankalp-khare:


sankalp-khare:
The preview linked by netlify loads version 1.18 docs. Is that where I should make this change?

(and not under master/ ?)

jeanduplessis:
@sankalp-khare thanks for picking up this issue. Would you mind also making the update to the 1.16, 1.17 and 1.18 docs. Also would you mind signing off your commit so that the DCO check can pass please.

PR #857, opened 2025-01-21
Update provider versions in examples (apply to older versions)

jeanduplessis:
This is a follow-up to https://github.com/crossplane/docs/pull/827 where I forgot to make the changes to all the versions.


jeanduplessis:
@jbw976 the v1 tags are currently active in the registry.

PR #860, opened 2025-01-22
Updated delete providers command

YashxG:
The command `kubectl delete provider <provider_name>` gives errors, the correct command to delete provider was `kubectl delete providers <provider_name>` replaced provider with providers.
I have attached the relevant screenshots.

`kubectl delete provider`
Screenshot 2025-01-22 at 11 50 33 AM

`kubectl delete providers`
Screenshot 2025-01-22 at 11 50 08 AM

<!--
Thanks for contributing to the docs! 

For information about local previews with Hugo read the contributing getting started guide:
https://docs.crossplane.io/contribute/contribute/

For information about Hugo's features like tabs and hint boxes read the styling features guide:
https://docs.crossplane.io/contribute/features/

For information on using Vale and testing locally read our Using Vale guide:
https://docs.crossplane.io/contribute/vale/

Need more help? Join the Crossplane Slack and ask in the #documentation channel.
https://slack.crossplane.io/

### DIFF
--- a/content/v1.18/software/uninstall.md
+++ b/content/v1.18/software/uninstall.md
@@ -138,10 +138,10 @@ NAME                   INSTALLED   HEALTHY   PACKAGE                            
 upbound-provider-aws   True        True      xpkg.upbound.io/upbound/provider-aws:v1.0.0   8h
 ```
 
-Remove the installed _providers_ with `kubectl delete provider`.
+Remove the installed _providers_ with `kubectl delete providers`.
 
 ```shell
-kubectl delete provider upbound-provider-aws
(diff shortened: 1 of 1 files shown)


jbw976 (commented):
Thanks for taking the time to make this update @YashxG!

I wasn't able to reproduce the behavior, but I know that depending on what CRDs are installed in your cluster, `kubectl` can have different behavior for using kinds without their groups or without plurals, etc. so I think this change is reasonable 😂 

Could you also please make this same edit to the other support version folders under the [`content` dir](https://github.com/crossplane/docs/tree/master/content)? i.e. v1.17, v1.19, and master?

PR #863, opened 2025-01-29
Align metadata.generateName in Create Managed Resource Script

timyip3:


content/master/getting-started/provider-aws.md
```diff
@@ -201,7 +201,7 @@ The {{< hover label="xr" line="3">}}apiVersion{{< /hover >}} and
 {{< hover label="xr" line="4">}}kind{{</hover >}} are from the provider's CRDs.
 
 
-The {{< hover label="xr" line="6">}}metadata.name{{< /hover >}} value is the 
+The {{< hover label="xr" line="6">}}metadata.generateName{{< /hover >}} value is the
```
jbw976, line 204: i think this `line="6"` can also be updated to `5` so it's highlighting the right line number on hover

content/master/getting-started/provider-aws.md
```diff
@@ -201,7 +201,7 @@ The {{< hover label="xr" line="3">}}apiVersion{{< /hover >}} and
 {{< hover label="xr" line="4">}}kind{{</hover >}} are from the provider's CRDs.
```
jbw976, line 201: i think most of these line numbers are 1 off, maybe something got removed in the past and they weren't updated. Are you interested in updating them so that hovering over them highlights the correct line? 😇

jbw976 (commented):
Thank you @timyip3 for providing this update! There's a bit more that could be done for completeness with the line numbers, let me know if you don't want to take those on also in this PR. If you're not up for it then we can merge this fix as is because it does make progress 😉

content/master/getting-started/provider-aws.md
timyip3: Thanks for the review! Let me update the line number together with this PR as well!

jbw976 (approved):
Awesome, this looks good on the [preview site](https://deploy-preview-863--crossplane.netlify.app/master/getting-started/provider-aws/#create-a-managed-resource) now, so let's go ahead and merge! thanks for also taking care of the line numbers @timyip3! 🙇‍♂️

PR #868, opened 2025-02-03
Update validate flags for new error on missing schemas flag

alwalker:
See https://github.com/crossplane/crossplane/issues/6003


jbw976 (approved):
This looks to be in agreement with what was merged recently in the following PR:

* https://github.com/crossplane/crossplane/pull/6264 

Thank you @alwalker! 🙇‍♂️

PR #881, opened 2025-02-23
Sync helm chart config values from crossplane/crossplane repo

jbw976:
This PR updates all versions of the docs to match their respective Crossplane version's Helm chart config values table. I don't think we have any automation around this and I noticed it was pretty out of date when looking into how to document https://github.com/crossplane/crossplane/pull/5540.

I basically just copied over the table from the helm chart README from each release branch, e.g. https://raw.githubusercontent.com/crossplane/crossplane/refs/heads/release-1.19/cluster/charts/crossplane/README.md for v1.19.

I've also tested locally with `hugo server`, the tables looked OK to me.


jbw976:
Looks like there are some new Vale errors to deal with 😇 https://github.com/crossplane/docs/actions/runs/13479464629/job/37662867954?pr=881

haarchri (approved):

PR #889, opened 2025-03-09
Update dead link in providers.md

matmilbury:

### DIFF
--- a/content/master/concepts/providers.md
+++ b/content/master/concepts/providers.md
@@ -21,10 +21,6 @@ Examples of providers include:
 * [Provider GCP](https://github.com/upbound/provider-gcp)
 * [Provider Kubernetes](https://github.com/crossplane-contrib/provider-kubernetes)
 
-{{< hint "tip" >}}
-Find more providers in Crossplane's [public package registries](https://www.crossplane.io/registries).
-{{< /hint >}}
-
 <!-- vale write-good.Passive = NO -->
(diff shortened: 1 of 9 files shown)


content/master/concepts/providers.md
```diff
@@ -22,7 +22,7 @@ Examples of providers include:
 * [Provider Kubernetes](https://github.com/crossplane-contrib/provider-kubernetes)
 
 {{< hint "tip" >}}
-Find more providers in Crossplane's [public package registries](https://www.crossplane.io/registries).
+Find more providers in Crossplane's [public package registries](https://marketplace.upbound.io/).
```
jbw976, line 25: great catch @matmilbury! We should actually remove this link entirely. For vendor neutrality, we don't want to link directly to the upbound marketplace at all. 

There's a similar hit that should be also entirely removed at https://github.com/crossplane/docs/blob/master/content/master/getting-started/introduction.md#providers

Would you be able to make these changes across all the version folders in the `content` dir? i.e. v1.17, v1.18, v1.19, and master? That would be really helpful! 🙇‍♂️

jbw976: @matmilbury someone brought this up in Crossplane slack again today, so I went ahead and pushed a commit to your branch that fixes all the locations, and kept you as the commit author 😉

jbw976 (approved):
Thanks again for the initiative on this @matmilbury! 🙇‍♂️

Issue #888, opened 2025-03-07
[Web Bug] - Quickstarts do not account for required RBAC permission and fail when managed resource is created


mattwelke:
The quickstarts as written now do not account for the fact that the provider pod (e.g. the S3 provider pod in the AWS quickstart) throws an RBAC error when the managed resource is created. This can be assumed to affect each quickstart (AWS and GCP tested manually so far).

The error is `k8s.io/client-go@v0.30.0/tools/cache/reflector.go:232: failed to list *v1beta1.ProviderConfig: providerconfigs.aws.upbound.io is forbidden: User "system:serviceaccount:crossplane-system:provider-aws-s3-f388c2201fc3" cannot list resource "providerconfigs" in API group "aws.upbound.io" at the cluster scope`.

See https://github.com/crossplane/docs/issues/887 for more context.

PR #890, opened 2025-03-10
Update function-patch-and-transform version to 0.8.2

mattwelke:
Updates the master folder of the docs to use the latest version of the transform and patch function so that a user following the quickstart will be successful.

There are some whitespace changes that made their way into this PR too. I'm not sure why. I tried to exclude them, but the files are changed when I run `git add`. I checked for Git hooks and the like, but didn't see any. Let me know if this is an issue.

**Background:**

I found when completing the GCP quickstart that during part 2, it fails when you get to the step where you apply a `Function`. The function never reaches the installed state because it cannot fetch the package:

```
cannot unpack package: failed to fetch package digest from remote: failed to fetch package descriptor with a GET request after a previous HEAD request failure: HEAD https://xpkg.crossplane.io/v2/crossplane-contrib/function-patch-and-transform/manifests/v0.1.4: unexpected status code 404 Not Found (HEAD responses have no body, use GET for details): GET https://xpkg.crossplane.io/v2/crossplane-contrib/function-patch-and-transform/manifests/v0.1.4: MANIFEST_UNKNOWN: manifest unknown
```

I resolved this issue by finding out what the latest version of the function package was (v0.8.2) and trying that instead. It worked, and I was able to complete the quickstart.

Note that there is another known [issue](https://github.com/crossplane/docs/issues/888) with the quickstart in the master folder of the docs that I had to work around too (in this case, I did so by adding the `cluster-admin` role to the Pub/Sub provider pod's service account).

### DIFF
--- a/content/master/concepts/compositions.md
+++ b/content/master/concepts/compositions.md
@@ -1,7 +1,7 @@
 ---
 title: Compositions
 weight: 30
-aliases: 
+aliases:
   - composition
   - composition-functions
   - /knowledge-base/guides/composition-functions
@@ -9,14 +9,14 @@ description: "Compositions are a template for creating Crossplane resources"
 ---
 
 Compositions are a template for creating multiple managed resources as a single
-object. 
+object.
 
 A Composition _composes_ individual managed resources together into a larger,
 reusable, solution.
 
 An example Composition may combine a virtual machine, storage resources and
 networking policies. A Composition template links all these individual
-resources together. 
+resources together.
 
 Here's an example Composition. When you create an
 {{<hover label="intro" line="8">}}AcmeBucket{{</hover >}} composite resource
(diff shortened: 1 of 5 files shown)


jbw976 (commented):
This is very helpful @mattwelke, thank you for taking care of this!! could you also make these same changes to the `content/v1.19` folder too? that's the latest docs version and the default that everyone sees, so it will be good to have the changes there so folks see it by default 😇

jbw976 (approved):
Actually, let's go ahead and merge this as is @mattwelke - i'm doing a big pass through all the docs for https://github.com/crossplane/crossplane/issues/6320 / https://github.com/crossplane/docs/issues/888, so I'll hit all these as I'm going through that. If we both do some work we'll likely just run into conflicts together 😂

PR #893, opened 2025-03-21
Bootstrap Crossplane v2.0-preview documentation

negz:
This'll be a somewhat unique version in that while it's the latest semver, it's a preview. So we want it to appear in the dropdown but we still want `/latest/` to link to v1.x.


jbw976 (approved):
This is a fine starting point for us to start collaborating and remodeling this for our v2 vision! 🧹 🚧  🏗️ 

preview site doesn't appear to have major regressions: https://deploy-preview-893--crossplane.netlify.app/v2.0-preview/

PR #896, opened 2025-03-26
Update install guide to install the preview

negz:
* Use the preview Helm repo
* Don't discuss the master Helm repo
* Defer to upstream and Helm docs to document Helm and our chart's options


content/v2.0-preview/get-started/install-crossplane-include.md
jbw976, line 1: this include file was only getting used in the provider getting started guides, right? the plan going forward is to just link to the install page as a pre-req on those now?

jbw976 (approved):
Looks OK on the preview site! https://deploy-preview-896--crossplane.netlify.app/v2.0-preview/get-started/install/

content/v2.0-preview/get-started/install-crossplane-include.md
negz: I'm not actually sure where it was used. 🤔 But yeah the plan is to just link to the install page.

PR #897, opened 2025-03-26
[v2] Add a "What's Crossplane?" page

negz:
This PR:

 * Moves all content that explains what Crossplane is to a new "What is Crossplane page"
 * Removes the content explaining what Crossplane is from the Get Started landing page
 * Removes the content explaining what Crossplane is from the docs landing (now welcome) page


content/v2.0-preview/_index.md
```diff
@@ -1,51 +1,53 @@
...
-one roof.
+Control planes expose an API. You use the API to tell the control plane what
+software it should configure and how - this is your _desired state_.
 
-Custom APIs, created by platform teams, allow security and compliance
-enforcement across resources or clouds, without exposing any complexity to the
-developers. A single API call can create multiple resources, in multiple clouds
-and use Kubernetes as the control plane for everything.
+A control plane can configure any cloud native software. It could deploy an app,
+create a load balancer, or create a GitHub repository.
```
jbw976, line 26: i think it'd be good to still mention an example of cloud infrastructure so people don't lose the fact that we still very much do infrastructure in addition to now being better suited for application level components :)

content/v2.0-preview/_index.md
```diff
@@ -1,51 +1,53 @@
...
+Crossplane has a rich ecosystem of extensions that make building a control plane
+faster and easier. It's built on Kubernetes, so it works with all the Kubernetes
+tools you already use.
 
-Crossplane lets you manage anything, anywhere, all through standard Kubernetes
-APIs. Crossplane can even let you 
-[order a pizza](https://blog.crossplane.io/providers-101-ordering-pizza-with-kubernetes-and-crossplane/) 
-directly from Kubernetes. If it has an API, Crossplane can connect to it. 
+{{< hint "tip" >}}
+**A control plane is software that controls other software.**
```
jbw976, line 20: i like ana's idea of immediately giving the practical example of `The major cloud providers are all built using control planes` right here near the top instead of the very end.

content/v2.0-preview/_index.md
```diff
@@ -1,51 +1,53 @@
...
 ---
 
-{{< img src="/media/banner.png" alt="Crossplane Popsicle Truck" size="large" >}}
+Welcome to the Crossplane documentation. Crossplane is a control plane framework
+for platform engineering. 
 
-<br />
+Crossplane lets you build control planes to manage your cloud native software.
+It lets you design the APIs and abstractions that your users use to interact
+with your control planes.
```
jbw976, line 13: should we add just another snippet here to help people understand why users will interact with control planes, to accomplish what? (provision and manage their resources?)

content/v2.0-preview/_index.md
```diff
@@ -1,51 +1,53 @@
 ---
-title: "Overview"
+title: "Welcome"
```
jbw976, line 2: i feel like this page is missing an end to end practical example to tie everything together - e.g. a user comes to a control plane built by the platform team with crossplane and uses it to deploy their application along with everything it needs. should that big picture practical value be conveyed on this page somehow?

jbw976 (commented):
i definitely like the direction here! a couple comments to consider, but i know we'll iterate on this more too over the coming days 😁

content/v2.0-preview/_index.md
negz: Okay, I'll do that.

content/v2.0-preview/_index.md
negz: I tried to do that by including the example of a load balancer. Is there a better example I could use? A more recognizable brand like RDS or ELB?

content/v2.0-preview/_index.md
negz: The tip box that explains what a control plane is already elaborates on that. Is that sufficient?

content/v2.0-preview/_index.md
negz: I'm not sure I agree.

I agree we need an example like that, but I'm not sure this landing page is the right place for it. I prefer to keep it relatively short and sweet - but hopefully enough that folks'll click through to other pages to learn more.

If you look at https://helm.sh/docs/ or https://kubernetes.io/docs/home/ for example, they only _very_ briefly mention what the actual project is.

negz: I could imagine us having a distinct "Why Crossplane" page similar to https://kubernetes.io/docs/concepts/overview/#why-you-need-kubernetes-and-what-can-it-do

negz: I ended up refactoring this into a new "What is Crossplane?" page. It now has practical examples and diagrams.

content/v2.0-preview/_index.md
```diff
@@ -1,51 +1,30 @@
...
 weight: -1
 cascade:
     version: "2.0-preview"
 ---
 
-{{< img src="/media/banner.png" alt="Crossplane Popsicle Truck" size="large" >}}
```
negz, line 8: I'm on the fence about whether we should keep this banner or not. LMK what you think.


Screenshot 2025-03-27 at 12 09 44 PM

vs

Screenshot 2025-03-27 at 12 10 03 PM

jbw976 (approved):
Overall, I think this is great @negz!! I like the what is crossplane messaging and sequence, but have a couple suggestions that may help the message land with more understandable impact. Feel free to push back on any of them if you think they don't fit well.

content/v2.0-preview/_index.md
negz: I like having the visual element, but I feel it pushes the actual content too far down the page.

content/v2.0-preview/_index.md
jbw976: i keep reading "load balancer" and thinking it's an in-cluster k8s component, even though it's not. I wonder if other folks make the same mental mistake. Regardless, I do think a more recognizable clear infrastructure piece could work better, such as a database.

jbw976:
btw, you weren't kidding that the mermaid graphs look WAY better in light mode - they even have animated lines! i didn't see any change when i switched to light mode, but then after refreshing the page the diagrams look way better 🥳

content/v2.0-preview/_index.md
jbw976: i like having a crossplane banner, i love the artwork - can it maybe be made smaller? 🤔

jbw976: i like having a crossplane banner, i love the artwork - can it maybe be made smaller? 🤔

negz: Not sure, and I don't feel like there's time to investigate. I'm going to leave it out for now. I'm open to re-adding it in a future PR if we find some way to make it take up less of the page.

content/v2.0-preview/what-is-crossplane/_index.md
```diff
@@ -0,0 +1,230 @@
...
+lifecycle. If your software ever _drifts_ from your desired state, the control
+plane automatically corrects the drift.
+{{< /hint >}}
+
+Crossplane has a rich ecosystem of extensions that make building a control plane
+faster and easier. It's built on Kubernetes, so it works with all the Kubernetes
+tools you already use.
+
+**Crossplane's key value is that it unlocks the benefits of building your own
+Kubernetes custom resources without having to write controllers for them.**
```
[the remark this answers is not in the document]
negz: I'm inclined to keep it short and sweet here.

There's a risk that someone reads this bolded text, doesn't get it, and closes the browser tab without scrolling down. However as long as they're curious enough to scroll down I think they'll get some examples of what they might use controllers for - e.g. by looking at the composition and MR sections.

content/v2.0-preview/what-is-crossplane/_index.md
```diff
@@ -0,0 +1,230 @@
...
+controller watch@<-- watch --> api
+controller -- create --> deployment
+controller -- create --> service
+
+watch@{animate: true}
+```
+
+**With Crossplane you don't have to write a controller**. Instead you configure
+a pipeline of functions. The functions return declarative configuration that
+Crossplane should apply.
```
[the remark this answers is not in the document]
negz: I like the idea - without this it doesn't really explain why functions are preferable to controllers.

I went with a different reason though: less boilerplate code, more focus on your business logic, and use a language you already know.

PR #898, opened 2025-03-27
[v2] Link to Crossplane install guide

negz:
Don't repeat it.

This is a follow on from https://github.com/crossplane/docs/pull/896. That PR actually deleted the install include, but the document is actually including the file from the `content/master` directory, not the new `content/2.0-preview` directory. 🙃

### DIFF
--- a/content/v2.0-preview/get-started/get-started-with-managed-resources.md
+++ b/content/v2.0-preview/get-started/get-started-with-managed-resources.md
@@ -10,13 +10,11 @@ with
 
 ## Prerequisites
 This quickstart requires:
-* a Kubernetes cluster with at least 2 GB of RAM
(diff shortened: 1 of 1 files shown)


jbw976 (approved):

jbw976 (approved):
This certainly does streamline the flow of a guide, and it helps get to the point/focus of the guide more easily, so I definitely like it from that perspective.

I worry maybe 10% about the case where folks don't read anything at all and just go to the first copy/paste box they find and start running commands from there. We'll always have folks like that and they will hit an error, but hopefully it will be easy enough for them to sort out. From that inevitable, but not super common scenario, I can understand why the full install instructions were included into each page 😇. I don't love them, but I do understand the motivation.

PR #899, opened 2025-03-27
Ignore Mermaid styling when purging CSS

negz:
We notice Mermaid styling looks good in Hugo (locally) but not when pushed to Netlify. Maybe this is why?

https://docs.crossplane.io/contribute/infrastructure/#how-optimization-works

### DIFF
--- a/themes/geekboot/assets/scss/docs.scss
+++ b/themes/geekboot/assets/scss/docs.scss
@@ -50,4 +50,7 @@
 /* purgecss end ignore */
 @import "skippy";
 @import "fonts";
+// TODO(negz): Does import order matter? Can this go in the existing ignore block above?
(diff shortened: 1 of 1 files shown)


themes/geekboot/assets/scss/docs.scss
```diff
@@ -50,4 +50,7 @@
 /* purgecss end ignore */
 @import "skippy";
 @import "fonts";
-@import "mermaid";
\ No newline at end of file
+// TODO(negz): Does import order matter? Can this go in the existing ignore block above?
```
RedbackThomson, line 53: Mermaid might eventually depend on fonts so it would need to come after that. I don't know what skippy is.

RedbackThomson (approved):

PR #900, opened 2025-03-27
full walkthrough for get started with managed resources guide

jbw976:
This PR updates the "Get started with Managed Resources" guide to work with the v2 preview compatible AWS provider.

* the provider version was updated to the latest build for v2
* namespaced managed resources are used
* a new section introduces the big change of namespaced managed resources and what they look like

### DIFF
--- a/content/v2.0-preview/get-started/get-started-with-managed-resources.md
+++ b/content/v2.0-preview/get-started/get-started-with-managed-resources.md
@@ -3,9 +3,8 @@ title: Get Started With Managed Resources
 weight: 200
 ---
 
-Connect Crossplane to AWS to create and manage cloud resources from Kubernetes 
-with  
-[provider-upjet-aws](https://github.com/crossplane-contrib/provider-upjet-aws).
+Connect Crossplane to AWS to create and manage cloud resources from Kubernetes
+with [provider-upjet-aws](https://github.com/crossplane-contrib/provider-upjet-aws).
 
 
 ## Prerequisites
@@ -16,10 +15,37 @@ This quickstart requires:
 * An AWS account with permissions to create an S3 storage bucket
 * AWS [access keys](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-creds)
 
+## About Managed Resources in Crossplane v2
+A _managed resource_ is anything Crossplane creates and manages outside of the
+Kubernetes cluster.
+
+This guide creates an AWS S3 bucket with Crossplane.
+
+The S3 bucket is a _managed resource_.
+
+Crossplane v2 allows you to compose namespaced resources. To better support this
+new ability, managed resources (MRs) are now namespaced in Providers that have
+been updated for Crossplane v2.
+
+To support backwards compatibility while users are adopting Crossplane v2, each
+provider will offer the legacy cluster scoped MRs in addition to the new
+namespaced MRs.
+
+For example, when the AWS provider that has been upated to support Crossplane v2
+is installed during this guide, you will see two CRDs for each type of managed
+resource:
+
+1. A legacy cluster scoped MR in the `*.aws.upbound.io` API group
+1. A namespaced MR in the `*.aws.m.upbound.io` API group
+
+{{< hint type="tip" >}}
+More about namespaced managed resources can be read in the [Crossplane v2 proposal](https://github.com/crossplane/crossplane/pull/6255).
+{{< /hint >}}
+
 ## Install the AWS provider
 
-Install the AWS S3 provider into the Kubernetes cluster with a Kubernetes 
-configuration file. 
+Install the AWS S3 provider into the Kubernetes cluster with a Kubernetes
+configuration file.
 
 ```yaml {label="provider",copy-lines="all"}
 cat <<EOF | kubectl apply -f -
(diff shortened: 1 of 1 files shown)


jbw976:
Just chatted with @negz about removing some additional details that aren't really necessary to be quickly successful in a getting started guide 🤓 - will take a pass through to streamline now

jbw976:
@negz pushed https://github.com/crossplane/docs/pull/900/commits/f78a611aaf57f4b6a0e2d53fe3298bab66e90e3f to remove the fine grained details and streamline the guide 🤓

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -16,10 +15,37 @@ This quickstart requires:
 * An AWS account with permissions to create an S3 storage bucket
 * AWS [access keys](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-creds)
 
+## About Managed Resources in Crossplane v2
+A _managed resource_ is anything Crossplane creates and manages outside of the
+Kubernetes cluster.
```
negz, line 20: Should we say control plane instead of cluster?

I go back and forth. "Kubernetes cluster" is more common and likely to make more sense to most people.

On the other hand "control plane" fits more closely with our framing of what Crossplane is.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -16,10 +15,37 @@ This quickstart requires:
...
+For example, when the AWS provider that has been upated to support Crossplane v2
+is installed during this guide, you will see two CRDs for each type of managed
+resource:
+
+1. A legacy cluster scoped MR in the `*.aws.upbound.io` API group
+1. A namespaced MR in the `*.aws.m.upbound.io` API group
+
+{{< hint type="tip" >}}
+More about namespaced managed resources can be read in the [Crossplane v2 proposal](https://github.com/crossplane/crossplane/pull/6255).
+{{< /hint >}}
```
negz, line 43: I'd actually prefer to leave this detail out (in this guide at least).

Instead I'd like to completely ignore legacy topics, and just show a v2 style namespaced MR in this guide. That way folks who aren't coming from the legacy Crossplane v1 world won't need to worry about any of this.

Or another way to put it: write the v2 preview documentation as if Crossplane v1 never existed.

For folks who _are_ coming from the legacy Crossplane v1 world I'd like to separately add a "What's new in Crossplane v2" page that explains things like this. I'm tentatively thinking we could add that under Guides, but (if possible) update the "This isn't the latest version" header to link to that guide if they're looking at a v2 page.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -109,32 +134,12 @@ generic aws-secret \
...
-customizes the settings of the AWS Provider.  
+customizes the settings of the AWS Provider.
 
-Apply the 
-{{< hover label="providerconfig" line="3">}}ProviderConfig{{</ hover >}} 
+Apply the
+{{< hover label="providerconfig" line="3">}}ProviderConfig{{</ hover >}}
 with the this Kubernetes configuration file:
 ```yaml {label="providerconfig",copy-lines="all"}
 cat <<EOF | kubectl apply -f -
```
negz, line 145: I know this is subjective but per my rant about `cat <<EOF` in examples yesterday I've been omitting these here docs from the getting started with composition examples.

Instead I'm just saying (e.g.) "Save this as `xrd.yaml`" and showing the kubectl command to apply it.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -152,33 +157,29 @@ spec:
...
+This attaches the AWS credentials, saved as a Kubernetes secret, as a
 {{< hover label="providerconfig" line="9">}}secretRef{{</ hover>}}.
 
-The 
-{{< hover label="providerconfig" line="11">}}spec.credentials.secretRef.name{{< /hover >}} 
-value is the name of the Kubernetes secret containing the AWS credentials in the 
-{{< hover label="providerconfig" line="10">}}spec.credentials.secretRef.namespace{{< /hover >}}.
+## Create a namespace
+Before we can create our namespaced S3 bucket managed resource, we must create a
+namespace for it.
```
negz, line 165: Can we omit this and just use the `default` namespace? That's what I'm doing in the Composition docs.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -188,44 +189,42 @@ spec:
...
-{{< hover label="xr" line="3">}}kind{{</hover >}} are from the provider's CRDs.
-
-
-The {{< hover label="xr" line="5">}}metadata.generateName{{< /hover >}} value is the 
-name of the created S3 bucket in AWS.  
-This example uses the generated name `crossplane-bucket-<hash>` in the 
-{{< hover label="xr" line="5">}}$bucket{{</hover >}} variable.
+The {{< hover label="xr" line="6">}}metadata.generateName{{< /hover >}} gives a
+pattern that the provider will use to create a unique name for the bucket in S3.
+The generated name will look like `crossplane-bucket-<hash>`.
```
negz, line 194:
```suggestion
The {{< hover label="xr" line="6">}}metadata.generateName{{< /hover >}} gives a
pattern that Kubernetes will use to create a unique name for the bucket in S3.
The generated name will look like `crossplane-bucket-<hash>`.
```

Nit: this is more technically accurate. That said, I don't feel strongly if it feels less confusing to pretend it's the provider doing it in this context.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -188,44 +189,42 @@ spec:
...
-NAME                      READY   SYNCED   EXTERNAL-NAME             AGE
-crossplane-bucket-hhdzh   True    True     crossplane-bucket-hhdzh   5s
+kubectl -n crossplane-aws-app get buckets.s3.aws.m.upbound.io
+NAME                      SYNCED   READY   EXTERNAL-NAME             AGE
+crossplane-bucket-7tfcj   True     True    crossplane-bucket-7tfcj   3m4s
 ```
 
 ## Delete the managed resource
-Before shutting down your Kubernetes cluster, delete the S3 bucket just created.
+Before shutting down your Kubernetes cluster, delete the S3 bucket that was just created.
```
negz, line 210:
```suggestion
Delete the S3 bucket that was just created.
```

Nit: Leading with "Before shutting down your Kubernetes cluster" feels weird to me for some reason here. What if they're using an existing cluster and not e.g. `kind`?

I'd suggest leaving it out. Maybe we could add a `{{< hint >}}` mentioning that it's important to delete the bucket before uninstalling Crossplane/deleting cluster/etc.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -188,44 +189,42 @@ spec:
...
-The {{< hover label="xr" line="8">}}spec.forProvider.region{{< /hover >}} tells 
-AWS which AWS region to use when deploying resources.  
-
-The region can be any 
-[AWS Regional endpoint](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints) code.
-
-Use `kubectl get buckets` to verify Crossplane created the bucket.
+Use `kubectl -n crossplane-aws-app get buckets.s3.aws.m.upbound.io` to verify Crossplane created the bucket.
 
 {{< hint type="tip" >}}
```
negz, line 198:
```suggestion
{{< hint "tip" >}}
```

This is an existing issue, but I noticed these aren't actually supposed to have a `type=`. It just happens that when you do `type="tip"` it doesn't know what it should be and defaults to `tip`. So if you do e.g. `type="note"` you'll get a `tip` instead. 🤔

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -188,44 +189,42 @@ spec:
...
-bucket.s3.aws.upbound.io "crossplane-bucket-hhdzh" deleted
+kubectl -n crossplane-aws-app delete buckets.s3.aws.m.upbound.io crossplane-bucket-7tfcj
+bucket.s3.aws.m.upbound.io "crossplane-bucket-7tfcj" deleted
 ```
 
+## Composing managed resources
+Crossplane v2 allows you to compose **any type of resource** into custom APIs
+for your users, which includes managed resources. Enjoy the freedom that
+Crossplane v2 gives you to compose the diverse set of resources your
+applications need for their unique environments, scenarios, and requirements.
```
negz, line 223: I'd drop v2 here and just say "Crossplane".

negz (changes requested):
Thanks @jbw976! This looks great - only blocker is my first and last comments WRT not specifically talking about v2 and how it's different in these getting started guides.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
jbw976: Got it, thanks for seeing that tone to follow here, will incorporate that.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
jbw976: I think we should be consistent, so I'll follow what you're doing, but I have realized when running through this guide multiple times that I do actually like the single step copy/paste that heredocs enable. As opposed to copying the content, saving it to a file, then having to get a terminal to that location too, which is more steps 😇

content/v2.0-preview/get-started/get-started-with-managed-resources.md
jbw976: yes that's definitely a better approach, no need to create a new namespace 🤓

content/v2.0-preview/get-started/get-started-with-managed-resources.md
jbw976: ah good call out, there were 2 other `"note"` boxes on this page that were being displayed as tips. I'll update them all.

content/v2.0-preview/get-started/get-started-with-managed-resources.md
jbw976: I'll remove this language from the mainline path and include it in a `<hint>` box.

jbw976:
@negz your feedback has been incorporated in https://github.com/crossplane/docs/pull/900/commits/3462331316920c0bf8f7f03311d5938ddee0108c

content/v2.0-preview/get-started/get-started-with-managed-resources.md
jbw976: I made this update and when I ran through the guide to test, it definitely felt slower and more clumsy to have to save things to files and then apply them. I do prefer the zippy copy pasty heredocs! 😂

negz: Long term I'd like to go back to a variant of what we used to do:

* Add the YAML manifests as distinct files
* Embed the YAML file in the examples (to avoid duplicates)
* Provide a command that applies the file by URL - e.g. `kubectl apply -f https://docs.crossplane.io/...`

This is what Kubernetes does - e.g. https://kubernetes.io/docs/tutorials/configuration/updating-configuration-via-a-configmap/#rollout-configmap-volume

I find it to be a sweet spot. If you want to just run the example in your shell you can easily copy the command. If you want to copy the YAML and edit it in vim or YAML you don't have a weird here doc getting in the way.

negz (approved):

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -149,83 +129,83 @@ spec:
...
 ```
 
+{{< hint "important" >}}
+Make sure to delete the S3 bucket before uninstalling the provider or shutting
+down your control plane. If those are no longer running, they can't clean up any
+managed resources and you would need to do so manually.
+{{< /hint >}}
+
+## Composing managed resources
+Crossplane allows you to compose **any type of resource** into custom APIs for
```
negz, line 201: Vale hack: it thinks "type of" is too wordy but "kind of" is fine. 🙃

content/v2.0-preview/get-started/get-started-with-managed-resources.md
```diff
@@ -17,66 +20,67 @@ This quickstart requires:
...
+```shell {label="kube-apply-provider",copy-lines="all"}
+kubectl apply -f provider.yaml
 ```
 
-The Crossplane {{< hover label="provider" line="3" >}}Provider{{</hover>}}
+The Crossplane {{< hover label="provider" line="2" >}}Provider{{</hover>}}
 installs the Kubernetes _Custom Resource Definitions_ (CRDs) representing AWS S3
-services. These CRDs allow you to create AWS resources directly inside 
+services. These CRDs allow you to create AWS resources directly inside
 Kubernetes.
```
negz, line 43: I'd like to drop all mentions of CRDs here - too much detail for this context IMO.

PR #903, opened 2025-03-28
update CRDs for v2.0-preview

jbw976:
This PR updates the CRDs for the API reference docs for the v2 preview. I simply copied the latest CRDs from the [v2 branch](https://github.com/crossplane/crossplane/tree/v2/cluster/crds) of upstream crossplane to the content directory here in this repo. This is no different than what we normally do for a docs release as per the [release template](https://github.com/crossplane/docs/blob/master/.github/ISSUE_TEMPLATE/new_release.md).

I will verify how the preview site displays these CRD updates and make sure they look OK.


jbw976:
The [preview site](https://deploy-preview-903--crossplane.netlify.app/v2.0-preview/api/) looks OK. A couple small observations:

* i was expecting to see at least one `v2*` API version, but I do not, probably because none of the v2 versions are marked as the storage version yet
* there's a small bug where you click on the `+` to expand either `Usage` type (there are 2 now, but in different API groups) and they both get expanded. Maybe some weird CSS conflict because of the same name.

negz:
Not showing the v2alpha1 XRD API is really unfortunate - that's the version we want folks to use.

https://github.com/crossplane/docs/blob/6a4cf06ee758d85f5260a31565057163255619a2/themes/geekboot/layouts/partials/crds.html#L26

Could we update this to either show only the newest version (not the storage version), or show all versions?

Or maybe support an annotation on the CRD manifests that overrides storage version as the selected version? We could just add that annotation manually after generation.

negz:
It'd also be really nice it did something to downplay deprecated versions - e.g. so folks can tell which Usage to use.

jbw976:
@negz pushed 2 commits to:

* manually set XRD's v2alpha1 version as the storage version so it will be displayed on the API reference page. I played around with sorting and displaying the highest version, but hugo doesn't understand how to sort k8s versions, so that did not work well. It thinks `v1beta1` is newer than `v1`.
* show `(deprecated)` after a CRDs name if it has been marked as deprecated. I am now only appending this text to the kind name for display, hopefully that's sufficient and we don't have to try any CSS business 🙏

negz (approved):

PR #904, opened 2025-03-28
Use a different warning for preview versions

negz:
This'll match any version ending with "-preview" and show a warning that it's a preview, instead of warning that it's an old version.


jbw976 (approved):

jbw976:
nice improvement to distinguish the warning on preview release pages!

PR #911, opened 2025-04-05
Wrote about crossplanes default limited access to third party custom …

QuadmanSWE:
…resources, and how to remedy.

As we talked about @negz [on slack](https://crossplane.slack.com/archives/CEG3T90A1/p1743812912331719?thread_ts=1743792119.574379&cid=CEG3T90A1)


negz (approved):
Thanks! I want to move this to another page but I can do that in a follow-up PR.

Issue #883, opened 2025-02-27
[Web Bug] - AWS Quickstart Part 2 - Missing Configuration in Documentation for Crossplane Composition


wnqueiroz:
**Description:**

I found an issue in the documentation related to Crossplane compositions. The provided example lacks a necessary configuration, which can cause misconfigurations or deployment failures.  

I was following the example that defines a composition with an S3 bucket and a DynamoDB table using a function-based patch-and-transform approach. However, the required configuration for provider reference is not included in the documentation.

**Steps to Reproduce:**

1. Follow the documentation example to create a `Composition` similar to:

   ```yaml
   apiVersion: apiextensions.crossplane.io/v1
   kind: Composition
   metadata:
     name: dynamo-with-bucket
   spec:
     mode: Pipeline
     pipeline:
       - step: patch-and-transform
         functionRef:
           name: function-patch-and-transform
         input:
           apiVersion: pt.fn.crossplane.io/v1beta1
           kind: Resources
           resources:
             - name: s3Bucket
               base:
                 apiVersion: s3.aws.upbound.io/v1beta1
                 kind: Bucket
                 metadata:
                   name: crossplane-quickstart-bucket
                 spec:
                   forProvider:
                     region: us-east-2
                   providerConfigRef:
                     name: default
               patches:
                 - type: FromCompositeFieldPath
                   fromFieldPath: "spec.location"
                   toFieldPath: "spec.forProvider.region"
                   transforms:
                     - type: map
                       map:
                         EU: "eu-north-1"
                         US: "us-east-2"
             - name: dynamoDB
               base:
                 apiVersion: dynamodb.aws.upbound.io/v1beta1
                 kind: Table
                 metadata:
                   name: crossplane-quickstart-database
                 spec:
                   forProvider:
                     region: "us-east-2"
                     writeCapacity: 1
                     readCapacity: 1
                     attribute:
                       - name: S3ID
                         type: S
                     hashKey: S3ID
               patches:
                 - type: FromCompositeFieldPath
                   fromFieldPath: "spec.location"
                   toFieldPath: "spec.forProvider.region"
                   transforms:
                     - type: map
                       map:
                         EU: "eu-north-1"
                         US: "us-east-2"
     compositeTypeRef:
       apiVersion: database.example.com/v1alpha1
       kind: NoSQL
   ```

2. Deploy the composition and observe the error/missing behavior.

**Expected Behavior:**  

[...]

PR #916, opened 2025-04-11
update dynamodb spec

iheartNathan:
Description

Issue: #883 

Fixes: #883

### DIFF
--- a/content/master/getting-started/provider-aws-part-2.md
+++ b/content/master/getting-started/provider-aws-part-2.md
@@ -385,6 +385,8 @@ spec:
                   - name: S3ID
                     type: S
                 hashKey: S3ID
+              providerConfigRef:
+                name: default
           patches:
             - type: FromCompositeFieldPath
               fromFieldPath: "spec.location"
--- a/content/v1.17/getting-started/provider-aws-part-2.md
+++ b/content/v1.17/getting-started/provider-aws-part-2.md
@@ -385,6 +385,8 @@ spec:
                   - name: S3ID
                     type: S
                 hashKey: S3ID
+              providerConfigRef:
+                name: default
           patches:
             - type: FromCompositeFieldPath
               fromFieldPath: "spec.location"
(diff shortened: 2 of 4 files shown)


negz:
I don't think this is actually necessary. When you omit `providerConfigRef`, it defaults to using the `ProviderConfig` named `default`. So this change is just documenting the default behavior.

https://github.com/crossplane/docs/issues/883 may be misleading. It's unclear whether anything actually broke, or whether @wnqueiroz just expected to see a `providerConfigRef` but didn't.

I could see a case for it being a good practice to show the `providerConfigRef` in the docs so that folks know what's happening. However I can also see a case for omitting it, to keep the getting started example as simple as possible.

iheartNathan:
There is no where in the Part 1 and Part 2 of the AWS Quickstart I can see the below mentioned.
> When you omit `providerConfigRef`, it defaults to using the `ProviderConfig` named default. 

In the part 1 of the AWS quickstart the  `providerConfigRef` was provided as part of the `bucket` yaml file, it's not surprising user might be expecting the same thing in the part 2, atleast for users who are new to crossplane.

For consistency in the docs I think it will be OK to have the `providerConfigRef` field present in both the `s3bucket` and `dynamodb`  spec or excluded on both.

wnqueiroz:
Hey folks!

What I was expecting from a quickstart experience was more of a copy-paste approach — just enough to get the changes applied in my environment and quickly "feel" how the tool works.

I understand the behavior @negz mentioned. Omitting certain fields can definitely make it harder for newcomers to understand how things are wired together under the hood.

If this behavior is something users are expected to be aware of, I believe it would be helpful to make that more explicit in the documentation.

I agree with @iheartNathan — my initial impression wasn’t that the field was optional or that omitting it would default to a specific `ProviderConfig`. It felt more like something was missing entirely.

negz:
I'd rather address this by updating the docs to explain the defaulting than by including the default value.

PR #919, opened 2025-04-25
change logs guide

### LINKED ISSUE #806: Document the change logs feature
### What's Missing?

runtime/core support for [cloud change logs](https://github.com/crossplane/crossplane/issues/5477) was added in [v1.17.0](https://github.com/crossplane/crossplane-runtime/releases/tag/v1.17.0). We still have some work to enable it across the ecosystem of Providers, but we'll need some good user facing documentation for how to enable and use this new feature.

Resources:

* runtime PR: https://github.com/crossplane/crossplane-runtime/pull/754
* design PR: https://github.com/crossplane/crossplane/pull/5822
* **early** example walkthrough of using the feature: https://github.com/jbw976/change-log-sidecar/blob/main/README.md 

Some specific content we'll want to include:

* how to enable the change logs feature in a provider that supports it
* example [`DeploymentRuntimeConfig`](https://github.com/jbw976/change-log-sidecar/blob/main/drc.yaml) to setup sidecar container and change logs output
* how to parse the change logs effectively

jbw976:
This PR adds a guide for enabling and using the change logs feature. It walks through a simple scenario with `provider-kubernetes` to enable change logs on the provider, create some resources and perform lifecycle operations on them, and how to check the change logs to learn more about the operations that were recorded.

I encourage reviewers to walk through the guide on the preview site and verify it works for them too. I have done so multiple times on my machine, but it would be good to get further validation.

Fixed #806

### DIFF
--- a/content/master/guides/change-logs.md
+++ b/content/master/guides/change-logs.md
@@ -1,0 +1,271 @@
+---
+title: Change Logs
+weight: 210
+description: "Change logs help you audit all changes made to your resources"
+state: alpha
+alphaVersion: "1.17"
+---
+
+The "change logs" feature is designed to help users of Crossplane Providers to
+understand what changes a provider is making to the resources it's managing.
+Whenever a provider creates, updates, or deletes a managed resource, an entry
+explaining the details of the change is recorded in the provider's change log.
+
+Change logs are important for awareness of the changes that a provider is
+making to its managed resources. Due to the nature of Crossplane's active
+reconciliation, it's possible for a provider to make changes to managed
+resources without any user interaction. Consider the scenario when someone
+updates a resource outside of Crossplane, for example via the AWS console or
+`gcloud` CLI. When Crossplane detects this configuration drift it will
+enforce its source of truth to eventually correct this unexpected change
+without any user interaction.
+
+With Crossplane acting continuously and autonomously to update critical
+infrastructure, it's vital for users to have insight into the operations being
+performed, so they can build and maintain a strong sense of confidence and trust
+in their control planes. Change logs provide details about all changes the
+provider makes, so users can remain aware of any changes, even when they aren't
+explicitly expecting any.
+
+{{<hint "tip">}} Change logs help you understand all the changes a provider is
+making to your resources, even when changes weren't explicitly requested, for
+example as a result of Crossplane's automatic correction of configuration drift.
+{{</hint>}}
+
+## Enabling Change Logs
+
+{{<hint "important" >}} Change logs are an alpha feature and must be explicitly
+enabled for each provider through the use of a `DeploymentRuntimeConfig`.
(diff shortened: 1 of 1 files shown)


content/master/guides/change-logs.md
```diff
@@ -0,0 +1,271 @@
...
+   enables communication of change events between the two containers.
+
+### Prerequisites
+
+This guide assumes you have a control plane with [Crossplane installed]({{<ref "../software/install">}}).
+
+It also assumes you have the [`jq` tool installed](https://jqlang.org/download/),
+to perform lightweight querying and filtering of the content in the change logs.
+
+The only other prerequisite for enabling change logs is that the provider must
```
haarchri, line 60: i see folks asking how to detect if a provider includes the changelog feature. In the meantime, could we add a  link tomain.go example from the provider-kubernetes project to help illustrate this? https://github.com/crossplane-contrib/provider-kubernetes/blob/3e694c4f2b644664167e67c58f3de8d2f9ffba0f/cmd/provider/main.go#L76

jbw976: yeah i was thinking that was going to be a likely question also 😂 - but i don't love the idea of having people go search through source code of their providers

another idea could be to look at the help output of the provider image, e.g.:
```
❯ docker run --rm xpkg.crossplane.io/crossplane-contrib/provider-kubernetes:v0.17.1 --help 2>&1 | grep enable-changelogs
❯ docker run --rm xpkg.crossplane.io/crossplane-contrib/provider-kubernetes:v0.18.0 --help 2>&1 | grep enable-changelogs
      --[no-]enable-changelogs  Enable support for capturing change logs during
```

jbw976: There was some offline discussion on a better long term approach that would allow packages to specify the features they support. That is not happening any time soon (it's just an idea right now), so can we punt on this for the first version of this guide and still get this otherwise very helpful material released on the docs site? 🙏 

@jeanduplessis @haarchri

content/master/guides/change-logs.md
```diff
@@ -0,0 +1,271 @@
...
+spec:
+  credentials:
+    source: InjectedIdentity
+EOF
+```
+
+### Create a resource
+
+Now that the provider is installed and configured with change logs enabled,
+create a resource that will generates change logs entries reflecting the actions
```
jeanduplessis, line 192:
```suggestion
create a resource that will generate change logs entries reflecting the actions
```

jeanduplessis (approved):
LGTM. Just one minor grammar fix proposal.

content/master/guides/change-logs.md
jbw976: thanks @jeanduplessis for the review! i fixed this grammar issue 🙇‍♂️

PR #928, opened 2025-05-06
Add autocompletion for CLI docs

twobiers:
Related to https://github.com/crossplane/crossplane/pull/6119

### DIFF
--- a/content/master/cli/_index.md
+++ b/content/master/cli/_index.md
@@ -62,3 +62,20 @@ For example, to install CLI version `v1.14.0` add `XP_VERSION=v1.14.0` to the
 download script curl command:  
 
 `curl -sL "https://raw.githubusercontent.com/crossplane/crossplane/main/install.sh" | XP_VERSION=v1.14.0 sh`
+
+## Installing shell autocompletions
+
+The Crossplane CLI supports shell autocompletions for `bash`, `zsh` and `fish`.
+You can install the autocompletions with the `completions` command by adding it to
+your shell's configuration file.
+
+```shell
+source <(crossplane completions)
+```
+
+{{<hint "note" >}}
+The `completions` command generates the autocompletions for your default shell.
+It's not possible to generate autocompletions for a different shell, if you want to
+install the autocompletions for a different shell, you have to configure the Crossplane
+CLI as the completer manually.
+{{< /hint >}}


content/master/cli/_index.md
```diff
@@ -61,4 +61,21 @@ By default the CLI installs from the `XP_CHANNEL` named `stable` and the
...
+`curl -sL "https://raw.githubusercontent.com/crossplane/crossplane/main/install.sh" | XP_VERSION=v1.14.0 sh`
+
+## Installing Shell Autocompletions
+
+The Crossplane CLI supports shell autocompletions for `bash`, `zsh` and `fish`.
+You can install the autocompletions with the `completions` command by adding it to
+your shell's configuration file.
+
+```shell
+source <(crank completions)
```
twobiers, line 73:
```suggestion
source <(crossplane completions)
```

content/master/cli/_index.md
```diff
@@ -61,4 +61,21 @@ By default the CLI installs from the `XP_CHANNEL` named `stable` and the
...
+your shell's configuration file.
+
+```shell
+source <(crossplane completions)
+```
+
+{{<hint "note" >}}
+The `completions` command generates the autocompletions for your default shell.
+It's not possible to generate autocompletions for a different shell, if you want to
+install the autocompletions for a different shell, you have to configure the crossplane
```
jbw976, line 79: I think this crossplane mention needs to be capitalized, since it's not in a code block. I think that's what's causing the vale error in https://github.com/crossplane/docs/actions/runs/14866616811/job/41744888843?pr=928, even though the line number seems off. I made this change locally and it made the vale error go away 😂 

More about working with Vale: https://docs.crossplane.io/contribute/vale/

Note, only errors must be fixed, warnings are OK if we don't think it's worth fixing them.

jbw976 (commented):
This looks pretty good on the preview site https://deploy-preview-928--crossplane.netlify.app/master/cli/#installing-shell-autocompletions.

What do you think about also adding a section to https://github.com/crossplane/docs/blob/master/content/master/cli/command-reference.md since this is a new command. The section on that page could just reference this new section on the top level index page.

content/master/cli/_index.md
twobiers: Actually it's because of the capitalization. If `Crossplane` is capitalized vale seems to treat it is a name and doesn't complain

twobiers:
> What do you think about also adding a section to https://github.com/crossplane/docs/blob/master/content/master/cli/command-reference.md since this is a new command. The section on that page could just reference this new section on the top level index page.

I don't know if that makes sense. Yes, it is technically a new command, but most CLI tools I know of list completions as part of the setup instructions and not in the command reference as it doesn't bring semantics and is only used a single time while setting the tool up and for that purpose there is already a reference:
https://github.com/crossplane/docs/blob/7b445dcff8eb6fe195e3be52f7f84405177b5380/content/master/cli/command-reference.md?plain=1#L11-L12

jbw976:
> most CLI tools I know of list completions as part of the setup instructions and not in the command reference

cool, I can buy that logic - sounds reasonable enough to me, thanks for explaining that perspective! 😉

jbw976 (approved):
Thank you @twobiers for making this docs contribution to go with your great new CLI functionality of autocompletion!

PR #936, opened 2025-05-21
Document the ImageConfig path rewriting feature in Crossplane 1.20

adamwg:
While we're here, move the common content on matching and debugging to their own top-level headings, since they apply to all the different ImageConfig features.

### DIFF
--- a/content/v1.20/concepts/image-configs.md
+++ b/content/v1.20/concepts/image-configs.md
@@ -258,9 +258,11 @@ Status:
 
 ### Interaction with other operations
 
+{{<hint "tip" >}}
 Image rewriting is always done before other `ImageConfig` operations. If you
 wish to configure pull secrets or signature verification as well as rewriting,
-other `ImageConfig` resources must match the rewritten image path.
+additional `ImageConfig` resources must match the rewritten image path.
+{{< /hint >}}
 
 For example, if you are mirroring packages from `xpkg.crossplane.io` to
 `registry1.com` and need to configure pull secrets for `registry1.com`, two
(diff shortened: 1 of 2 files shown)


content/v1.20/concepts/image-configs.md
```diff
@@ -211,4 +203,127 @@ If you can't see this condition on the package revision resource, namely
...
+  ...
+  Package:                        xpkg.crossplane.io/crossplane-contrib/provider-family-aws:v1.22.0
+Status:
+  ...
+  Resolved Package:        registry1.com/crossplane-contrib/provider-family-aws:v1.22.0
+```
+
+### Interaction with other operations
+
+Image rewriting is always done before other `ImageConfig` operations. If you
```
jbw976, line 261: this seems pretty important for folks to know, consider putting it into a "hint - tip" box to help it stand out a bit more. details on how to use them are in https://docs.crossplane.io/contribute/features/#hints-and-alert-boxes.

content/v1.20/concepts/image-configs.md
```diff
@@ -10,6 +10,35 @@ description: "Image Configs is an API for centralized control of the configurati
 Crossplane package images. It allows you to configure package manager behavior
 for images globally, without needing to be referenced by other objects.
 
+## Matching image references
```
jbw976, line 13: This feature is also in `master` builds, so the same changes should be copied to `/content/master` as well. This docs repo does not use git branches for the different versions.

jbw976 (commented):
This is great @adamwg, thanks for writing up such helpful information about the image rewriting feature you built! 🙌 

This also should be copied to `/content/master` so it's applicable there too, and then just one small suggestion for you to consider beyond that. Thank you!

content/v1.20/concepts/image-configs.md
adamwg: Done!

jbw976 (approved):
Great work @adamwg, thanks for this contribution! 🎉

PR #950, opened 2025-07-17
Fix --embed-runtime-image build doc

joebowbeer:
Fixes `--embed-runtime-image` flag in docs

https://docs.crossplane.io/master/cli/command-reference/#xpkg-build

```bash
sed s/runtime-image-name/runtime-image/g
```


jeanduplessis:
Thank you @joebowbeer

PR #955, opened 2025-08-05
Prepare docs for v2.0 release

negz:
This PR:

* Removes the old content/master and replaces it with a copy of content/v2.0-preview
* Fixes all the links that broke - mostly because versions were linking to content in master when they shouldn't
* Addresses all vale errors (in all versions)
* Addresses all vale warnings in content/master

I took a shot at using Claude Code to address all vale issues but it was estimating more than 24 hours to do so. 🙃 

There's tons of Vale issues because this PR moves and touches a lot of files. None of them are introduced by this PR.


jeanduplessis (changes requested):
LGTM in general. I think we just need to update the provider versions to the latest v2 compatible ones.

jeanduplessis (approved):

PR #958, opened 2025-08-07
Document MRD/MRAP/SafeStart

n3wscott:
first draft of MRD documentation based on the spec and PRs related to MRD/MRAP/SafeStart


content/master/get-started/get-started-with-mrds.md
```diff
@@ -0,0 +1,529 @@
...
+```shell
+error validating data: ValidationError(Cluster): unknown field "apiVersion" in io.k8s.api.core.v1.Cluster
+```
+
+## Expand activation with wildcards
+
+Add more resources using wildcard patterns. Update your activation policy:
+
+```shell
+kubectl patch mrap aws-demo-resources --type merge -p '{
```
phisco, line 379: nit: I prefer `kubectl apply -f - <<EOF` with full manifests, so that people can copy paste from any point in the guide, rather than incremental patches. Definitely not a showstopper though.

content/master/get-started/get-started-with-mrds.md
```diff
@@ -0,0 +1,529 @@
+---
```
negz, line 1: I slept on it, and still feel this get started guide is out of place here.

Big +1 to having a guide like this in general. I think what I'm getting hung up on is that the "Getting started" section of the docs is today the place you come to get introduced to Crossplane's more core, valuable concepts. Until this release it only covered Composition and MRDs - now Operations too. The stuff you use to build your control plane.

MRDs and MRAPs, while important and impactful, feel more like an implementation detail. Something that supports managed resources working well, rather than one of the first things you should learn when you pick up Crossplane. Closer to the `*Config` types (`ProviderConfig`, `EnvironmentConfig` etc) which don't have getting started guides today.

So I could look at this two ways. One is "well we want a getting started guide and this is where getting started guides go". Under this approach we might eventually add getting started guides for other "supporting" features like EnvironmentConfig one day". The downside of this approach in my mind is that for new users it's not immediately clear which guides are for super important stuff and which are for supporting concepts.

The best alternative I can think of is to keep this guide but move it under guides. Which is a bit weird given it's a getting started guide. The logic there becomes "getting started is for guides for people getting started _with Crossplane_, other advanced getting started guides just go under guides". If we did this I'd also add a pointer note from the MRs getting started guides.

content/master/get-started/get-started-with-mrds.md
negz: I believe the other getting started guides _don't_ currently use the EOF style. This guide should match them closely stylistically.

FWIW preferences like this in our docs have been a big challenge over the years. We've gone back and forth on here docs. Last time we talked about it I believe the consensus was to do what the Kubernetes docs do, which is:

1. Example manifests are distinct files in the docs repo
2. Our docs site tool (currently Hugo) embeds those files inline
3. We include `kubectl apply -f https://docs.crossplane.io/...` links to quickly apply them

IMO the Kube docs approach gives the best of both worlds.

* You can quickly apply a manifest by copying a CLI command
* The example manifest shown inline is plain old YAML, not YAML with arcane shell stuff around it

content/master/get-started/get-started-with-mrds.md
jbw976: I think it's reasonable to include it under guides, instead of getting started, similar to what we do for change logs: https://docs.crossplane.io/v1.20/guides/change-logs/

I'd like to reserve the getting started guide to just the top level critical concepts to introduce folks to as they are seeing crossplane for the first time. MRDs are super important for operating Crossplane, but I'd like to keep it very scoped for what we introduce folks to right away.

content/master/get-started/get-started-with-mrds.md
negz: I don't think we had a tracking issue for this - raised https://github.com/crossplane/docs/issues/960.

negz (commented):
I've just read the getting started guide so far.

I see right as I'm hitting comment that you just moved it though. Some of my comments about sections being too detailed are less important in that context.

content/master/guides/mrd-activation-policies.md
```diff
@@ -0,0 +1,689 @@
...
+  # Storage
+  - buckets.s3.aws.crossplane.io
+  - "*.storage.azure.crossplane.io"
+  
+  # Databases
+  - "*.rds.aws.crossplane.io"
+  - instances.sql.gcp.crossplane.io
+```
+
+### Staging environment
```
negz, line 264: Covering all of dev, staging, and prod feels a bit overkill IMO. I'd keep it at dev and prod if we give advice on this at all.

content/master/guides/mrd-activation-policies.md
```diff
@@ -0,0 +1,689 @@
...
+  # Databases (read-only access through compositions)
+  - "*.rds.aws.crossplane.io"
+  - instances.sql.gcp.crossplane.io
+```
+
+## Dynamic activation patterns
+
+### Conditional activation
+
+Use multiple policies to create conditional resource activation:
```
negz, line 425: Is this actually conditional?

content/master/guides/mrd-activation-policies.md
```diff
@@ -0,0 +1,689 @@
...
+  - clusters.eks.aws.crossplane.io
+  - "*.lambda.aws.crossplane.io"
+```
+
+You can delete the `advanced-resources` policy to deactivate optional 
+resources while keeping base capability.
+
+### Feature flag patterns
+
+Use labels and naming to create feature flag-like behavior:
```
negz, line 455: How? Don't they just all roll-up to a singleton MRAP policy regardless of labels?

content/master/guides/mrd-activation-policies.md
```diff
@@ -0,0 +1,689 @@
...
+{{< /tab >}}
+{{< /tabs >}}
+
+## Policy lifecycle management
+
+<!-- vale Google.Headings = NO -->
+### GitOps workflow
+<!-- vale Google.Headings = YES -->
+
+Store activation policies in Git for proper change management:
```
negz, line 602: This feels low value. "Use GitOps" could be true for any component of Crossplane. I don't think we need a specific callout for MRAPs.

content/master/guides/mrd-activation-policies.md
```diff
@@ -0,0 +1,689 @@
+---
+title: Writing MRD activation policies
+weight: 150
+description: Learn how to create effective activation policies for managed resources
+---
```
negz, line 5: I'd be comfortable dropping this guide entirely. If there's anything important that's only mentioned in this guide right now, it could go in the adjacent "using MRDs and activation policies" guide.

n3wscott:
@negz why don't you just edit and push the changes you want? or fork the PR

content/master/get-started/get-started-with-mrds.md
```diff
@@ -0,0 +1,529 @@
...
+  - description: The public IP address assigned to the instance
+    name: public_ip
+    type: string
+  - description: The private IP address assigned to the instance  
+    name: private_ip
+    type: string
+  - description: The public DNS name assigned to the instance
+    name: public_dns
+    type: string
+```
```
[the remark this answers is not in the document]
n3wscott: no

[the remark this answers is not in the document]
negz: I'd suggest dropping it then (no-one will be able to 'follow' the guide yet).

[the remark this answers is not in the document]
n3wscott: how are the providers going to implement them if they don't get documented?

PR #972, opened 2025-08-13
remove obsolete analytics

jbw976:
This PR removes obsolete analytics from the docs site infrastructure. There is no need for these analytics anymore, so they can be safely removed.

Looks like some white space changes (remove trailing whitespace) made it in automatically from my editor too.

I have tested these changes as described in https://github.com/crossplane/docs/pull/972#issuecomment-3186279921


jbw976:
I was able to test this well locally with `hugo server` and the following approach:

* removed the `-{{ if eq hugo.Environment "production" -}}` check in https://github.com/crossplane/docs/blob/master/themes/geekboot/layouts/partials/scripts.html#L6-L8
* ran `hugo server` and navigated to http://localhost:1313/ 
* watched network traffic with chrome dev tools
* validated pages are rendering OK, no errors visible on page or in network traffic

I think this is tested enough to review now, taking out of draft 💪

PR #973, opened 2025-08-14
Update sidebar behavior for v2 docs

tr0njavolta:
This PR attempts to fix empty index page behavior in the sidebar for v2 docs. Will need to make similar adjustments to nested pages in other versions.


negz:
I much prefer this sidebar behavior. I do notice though that if you browse directly to https://deploy-preview-973--crossplane.netlify.app/v2.0/get-started/ you still get the blank page. That could be an issue for folks with outdated links.

PR #975, opened 2025-08-14
Document using required (fka extra) resources in compositions

negz:
Right now AFAIK only function-python supports this.

### DIFF
--- a/content/master/composition/compositions.md
+++ b/content/master/composition/compositions.md
@@ -606,15 +606,15 @@ Most composition functions read the observed state of the composite resource,
 and use it to add composed resources to the desired state. This tells Crossplane
 which composed resources it should create or update.
 
-If the function needs __extra resources__ to determine the desired state it can
-request any cluster-scoped resource Crossplane already has access to, either by
+If the function needs __required resources__ to determine the desired state it can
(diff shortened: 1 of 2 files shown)


content/master/composition/compositions.md
```diff
@@ -767,6 +767,185 @@ Crossplane doesn't validate function input. It's a good idea for a function to
 validate its own input.
 {{</hint>}}
 
+### Required resources
+
+{{<hint "note">}}
+Crossplane v1 called this feature "extra resources." The v2 API
+uses the name "required resources" and adds support for bootstrap requirements.
+Functions now support the new API - function-python
```
jbw976, line 775: is function-python meant as an example? it's sort of reading like it's the one authoritative implementation 

perhaps a "for example" here could help?

jbw976 (approved):
LGTM! just one nit - I assume you'll sync to v2 content also?

content/master/composition/compositions.md
negz: Whoops, yeah I noticed this too and thought I fixed it. Will fix.

negz: I just deleted it.

It was originally trying to say that only function-python really supports this right now (AFAIK). In retrospect I don't wanna say that though - it'll quickly become outdated.

PR #978, opened 2025-08-18
docs: remove X from kind

tenstad:
It's unclear how one should name XRD kinds.
Given the following: https://github.com/crossplane/docs/blob/3b8bad4228132f97aaf6b51a316fa19c1e76f76c/content/v2.0/get-started/get-started-with-composition.md#L128-L134
should the `kind`s within `composite-resource-definitions.md` also be rewritten to remove X?

If not, feel free to just close this PR.


haarchri:
thanks for taking care about the old Claim convention, we have an other contribution in this area - superseeded by https://github.com/crossplane/docs/pull/1134

PR #988, opened 2025-09-04
Fix unintended paragraph breaks

branden:
This moves HTML comments out of paragraphs because they're creating breaks. I couldn't find a way to inline these comments without adding a paragraph break, let me know if there's a way to fix the doc while keeping linter exceptions fine grained.


jbw976 (approved):
nice, thanks for taking the time to fix this @branden! 🙇‍♂️

PR #992, opened 2025-09-11
Update _index.md nit

jaymiracola:
by to be =)

### DIFF
--- a/content/master/whats-new/_index.md
+++ b/content/master/whats-new/_index.md
@@ -128,7 +128,7 @@ Read more about Crossplane v2's [backward compatibility](#backward-compatibility
 
 Crossplane v2 makes all managed resources (MRs) namespaced.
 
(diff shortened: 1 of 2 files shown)


jbw976 (commented):
haha, this update woke up Vale 👻 
https://github.com/crossplane/docs/actions/runs/17648923949/job/50154397618?pr=992

```
 131:33  warning  'be composed' may be passive    write-good.Passive voice. Use active voice if you can.
```

You can reword it as suggested, but also consider adding a Vale exception if that feels better. you can find lots of examples in the code base, but it'd be something like:

```
<!-- vale write-good.Passive = NO -->
This enables a namespaced XR....
<!-- vale write-good.Passive = YES -->
```

Also, could you copy the final changes over to the `content/v2.0` folder so they show up on that latest version right away too? Thanks!

PR #1019, opened 2025-11-05
Release v2.1 process

jbw976:
This PR is the the result of the release process for the Crossplane docs for v2.1. The checklist of actions can be seen in:

* https://github.com/crossplane/docs/issues/1018.

I've tried to make most every step as a separate commit.

I've also run `hugo server` locally and poked around for any obvious issues and we'll also check the preview site once it deploys.

Fixes https://github.com/crossplane/docs/issues/1018


negz:
https://deploy-preview-1019--crossplane.netlify.app/v2.1/api/

Not a blocker, but I notice the generated API docs show XRD v1 as deprecated, but doesn't show XRD v2. Makes it look like XRDs are deprecated.

PR #1025, opened 2025-11-26
Split metrics documentation table by source

kaessert:
The metrics documentation previously mixed metrics from different sources in a single table, making it unclear where each metric originates.

Split into four sections:
- Crossplane core metrics: function runner, response cache, circuit breaker, and engine metrics emitted by the Crossplane pod
- Provider metrics: crossplane_managed_resource_* metrics from crossplane-runtime, emitted by all providers
- Upjet provider metrics: upjet_resource_* metrics only from Upjet-based providers (provider-aws, provider-azure, provider-gcp)
- Controller-runtime and Kubernetes client metrics: external dependency metrics emitted by both Crossplane and providers

Additional changes:
- Fixed metric name composition_run_function_seconds to function_run_function_seconds (matching actual code)
- Added missing metrics: cache metrics, engine metrics, crossplane_managed_resource_drift_seconds
- Added missing upjet metrics: cli_duration, active_cli_invocations, running_processes
- Removed _bucket suffix from histogram metric names (added by Prometheus)

Applied to all doc versions: v1.20, v2.0-preview, v2.0, v2.1, master

### DIFF
--- a/content/master/guides/metrics.md
+++ b/content/master/guides/metrics.md
@@ -1,7 +1,7 @@
 ---
 title: Metrics
 weight: 60
-description: "Monitor Crossplane operations with metrics"
+description: "Track Crossplane operations with metrics"
 ---
 
 Crossplane produces [Prometheus style metrics](https://prometheus.io/docs/introduction/overview/#what-are-metrics) for effective monitoring and alerting in your environment.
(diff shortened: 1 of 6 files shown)


content/master/guides/metrics.md
```diff
@@ -23,39 +23,84 @@ prometheus.io/port: "8080"
...
+{{</table >}}
+
+## Provider metrics
+
+These metrics are emitted by Crossplane providers. All providers built with crossplane-runtime emit the `crossplane_managed_resource_*` metrics.
+
+Providers expose metrics on the `metrics` port (default `8080`). To scrape these metrics, configure a `PodMonitor` or add Prometheus annotations to the provider's `DeploymentRuntimeConfig`.
+
+{{< table "table table-hover table-striped table-sm">}}
+| Metric Name | Description | Further Explanation |
```
jbw976: same here, the "further explanation" column is empty and can be removed...

content/master/guides/metrics.md
```diff
@@ -23,39 +23,82 @@ prometheus.io/port: "8080"
 prometheus.io/scrape: "true"
 ```    
 
+## Crossplane core metrics
```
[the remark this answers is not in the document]
jbw976: that does sound useful - looks like we do have a bit of that already in the provider section, e.g.:

> Providers expose metrics on the `metrics` port (default `8080`). To scrape these metrics, configure a `PodMonitor` or add Prometheus annotations to the provider's `DeploymentRuntimeConfig`.

content/master/guides/metrics.md
kaessert: same

content/master/guides/metrics.md
```diff
@@ -23,39 +23,84 @@ prometheus.io/port: "8080"
...
+| {{<hover label="upjet_resource_active_cli_invocations" line="6">}}upjet_resource_active_cli_invocations{{</hover>}} | The number of active (running) Terraform CLI invocations |  |
+| {{<hover label="upjet_resource_running_processes" line="7">}}upjet_resource_running_processes{{</hover>}} | The number of running Terraform CLI and Terraform provider processes |  |
+{{</table >}}
+
+## Controller-runtime and Kubernetes client metrics
+
+These metrics come from the controller-runtime framework and Kubernetes client libraries. They are emitted by both Crossplane and providers.
+
 {{< table "table table-hover table-striped table-sm">}}
 | Metric Name | Description | Further Explanation |
```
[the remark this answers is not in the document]
kaessert: Merged into description

content/v1.20/guides/metrics.md
```diff
@@ -23,36 +23,91 @@ prometheus.io/port: "8080"
 prometheus.io/scrape: "true"
 ```    
 
+## Crossplane core metrics
```
jbw976: come to think of it, i don't think we should update old versions like `v1.20` and `v2.0-preview`, because a lot of these values had different names or didn't even exist in those previous versions.  but let's also try not to overcomplicate it either, so what do you think of the following?

* don't bother updating `v2.0-preview`, that release is a bit of a one-off and those docs aren't even exposed in the version picker anymore
* don't bother updating `v1.20`, unless you really want to get every old name correct for `crossplane_composition_watches_*` etc. as shown in https://github.com/crossplane/crossplane/releases/tag/v2.0.0
* remove `circuit_breaker_*` metrics from `v2.0`, those were added in `v2.1`
* keep `v2.1` and `master` as is 🎉

jbw976 (commented):
Awesome @kaessert, this is looking really good now! just another comment though on the older versions and how best we can approach those. What do you think?

content/v1.20/guides/metrics.md
kaessert: Great points!

- `v2.0-preview` changes i threw away
- put an effort to get every `v1.20` old name correct -> we already went too far for shortcuts 😅 
- `circuit_breaker_*` was removed from everywhere except `v2.1`
- `v2.1` master kept as is ;)

Issue #1001, opened 2025-10-06
Document how XR connection details work in v2


jbw976:
### What's Missing?

There's a bit of confusion, and out of date documentation, for how connection details work for XRs in v2.

Some of the problems are captured in https://github.com/crossplane/crossplane/issues/6837#issuecomment-3372365347.

TL;DR, we removed connection details from XRs in v2, but we don't have good documentation and guidance on how to handle that.  Nic's suggestion in https://github.com/crossplane/crossplane/issues/6440 is the only "doc" i know of:

> I feel comfortable dropping this functionality (from core) for v2 XRs because in v2 you can recreate it using functions - just have your XR compose a secret with the XR's connection details in it.

PR #1033, opened 2025-12-12
fix: fix connection details page so XRD and Composition keys match

jbw976:
While getting deeper into making sense of connection secrets in v2 for #1001, I wanted to walk through how connection details worked in v1 scenarios. The docs page for that content, https://docs.crossplane.io/v1.20/concepts/connection-details, has a couple issues preventing it from working correctly.

This PR updates that content to ensure that the connection details key names in the XRD match what is being used in the Composition. When they are not matching, the desired keys do not make it to the final XR and Claim connection secrets because they get filtered out by the XRD's `.spec.connectionSecretKeys` list.


negz (approved):
Thanks!

PR #1045, opened 2026-01-17
Update guide on how to write a composition function in Go after v2 updates

### LINKED ISSUE #1003: Guide for writing function in go has issues with v2 updates
### URL
https://docs.crossplane.io/v2.0/guides/write-a-composition-function-in-go/ 

### What's Wrong?

The example function in https://docs.crossplane.io/v2.0/guides/write-a-composition-function-in-go/ imports the types from `provider-upjet-aws:v2`:

```
go get github.com/crossplane-contrib/provider-upjet-aws@v2.0.0
```

Which results in:
```
❯ go get github.com/crossplane-contrib/provider-upjet-aws@v2.0.0
go: github.com/crossplane-contrib/provider-upjet-aws@v2.0.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/crossplane-contrib/provider-upjet-aws/v2")
```

This may possibly work after:
* https://github.com/crossplane/function-sdk-go gets a v2 compatible update, tracked in https://github.com/crossplane/function-sdk-go/issues/216
* provider-upjet-aws gets a `/v2` module path, tracked in https://github.com/crossplane-contrib/provider-upjet-aws/issues/1846 

But there may be further issues blocking the entire guide working smoothly. We should revisit this guide now that v2 has shipped, make sure everything works, and make any updates that are needed.

jonasz-lasut:
Fixes #1003 

Updated go get instruction for AWS Provider Go module and fixed import paths that were leading to function not compiling and not passing tests.

github.com/upbound/provider-aws/v2 import has to use at least version `v2.3.0` as it's the first version with the correct module path

### DIFF
--- a/content/master/guides/write-a-composition-function-in-go.md
+++ b/content/master/guides/write-a-composition-function-in-go.md
@@ -275,7 +275,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 
-	"github.com/crossplane-contrib/provider-upjet-aws/apis/s3/v1beta1"
+	"github.com/upbound/provider-aws/v2/apis/namespaced/s3/v1beta1"
 
 	"github.com/crossplane/function-sdk-go/errors"
 	"github.com/crossplane/function-sdk-go/logging"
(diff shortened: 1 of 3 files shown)


jbw976 (approved):
Thanks for tackling this @jonasz-lasut!! it will be great to have this guide working again since moving to v2 😇

Issue #1022, opened 2025-11-20
Document configuring DeploymentRuntimeConfig for dependencies with ImageConfig


adamwg:
Once crossplane/crossplane#6940 is merged, we should document this new `ImageConfig` capability.

PR #1050, opened 2026-01-28
docs(packages): add docs about imageconfig runtime configref

haarchri:
add docs about imageConfig `spec.runtime.configRef` 
Fixes: #1022


haarchri:
@adamwg thanks you need to merge it - have no access

PR #1062, opened 2026-02-23
docs: fix broken Open Policy Agent link in multi-tenant guide

kruthiwusirika5:
Remove space between ] and ( in markdown link so reference-style link [Open Policy Agent] resolves correctly to the URL defined at the bottom of the file.

### DIFF
--- a/content/v1.20/guides/multi-tenant.md
+++ b/content/v1.20/guides/multi-tenant.md
@@ -253,7 +253,7 @@ policy won't be flexible enough. However, because Crossplane brings
 management of external infrastructure to the Kubernetes API, it's well suited
 to integrate with other projects in the cloud native ecosystem. Organizations
 and individuals that need a more robust policy engine, or just prefer a more
-general language for defining policy, often turn to [Open Policy Agent] (OPA).
+general language for defining policy, often turn to [Open Policy Agent].
 OPA allows platform builders to write custom logic in [Rego], a domain specific
 language. Writing policy in this manner allows for not only incorporating the
 information available in the specific resource being evaluated, but also using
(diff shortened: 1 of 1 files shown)


haarchri (approved):

kruthiwusirika5:
@haarchri — gentle nudge on this one-line fix (already approved 2026-03-13). Just removes a stray space in `[Open Policy Agent] (OPA)` so the reference-style link resolves to its URL definition further down the file.

The only red CI is the Vale `lint-content` job, which reports **0 errors, 82 warnings** — all pre-existing style issues in `content/v1.20/guides/multi-tenant.md` (passive voice, word substitutions, heading capitalization, etc.) that this PR does not touch. No new prose was added that could trigger Vale.

If the warnings are a hard merge gate, happy to open a separate PR addressing them. If they're not, this should be safe to land. Thanks!

jbw976 (commented):
Thanks for following up on this! I was just checking the live page now and the link seems to resolve and work for me OK. are you still seeing this issue on the live page?

is this the right paragraph? https://docs.crossplane.io/v1.20/guides/multi-tenant/#policy-enforcement-with-open-policy-agent

kruthiwusirika5:
@jbw976 — thanks for double-checking. You're right: just verified that
the live page renders correctly. I'd misdiagnosed this — Hugo's CommonMark
renderer treats `[Open Policy Agent] (OPA)` as a valid shortcut reference
link followed by literal ` (OPA)` text, not as an inline link with a
malformed URL.

So the change here is cosmetic, not a fix: it just removes the redundant
` (OPA)` after the resolved link. Keeping the acronym is also defensible
(spells out OPA on first mention). Happy to close this PR — sorry for
the mis-framing.

PR #1069, opened 2026-03-05
docs: clarify writeConnectionSecretToRef removal is XR-only, not MR-level

ytsarev:
A common source of confusion when upgrading to v2 is assuming that the removal of native XR connection details also affects managed resources. This is not the case — writeConnectionSecretToRef on MRs is unchanged.

Add a tip callout to the upgrade guide's "Composite resource connection details" section and expand the note in the connection details composition guide to explicitly state that only the XR-level aggregation was removed, while MR-level usage continues to work as before. Applied across all v2.x versions and master.

### DIFF
--- a/content/master/guides/connection-details-composition.md
+++ b/content/master/guides/connection-details-composition.md
@@ -17,7 +17,9 @@ resources and exposes them for the XR.
 
 {{<hint "note">}}
 Crossplane v1 included a feature that automatically created connection details
-for XRs.
+for XRs. Crossplane v2 removes this feature **for XRs only**. Managed Resources (MRs)
+aren't affected by this change and still support connection details via their
+`writeConnectionSecretToRef` field.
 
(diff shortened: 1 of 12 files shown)


jbw976 (commented):
Nice, thank you @ytsarev for taking the initiative to clarify this for Crossplane users @ytsarev! I have a couple wording suggestions here to keep the message very focused - let me know what you think 🙇 

another place we should probably add a brief mention to as well is https://docs.crossplane.io/v2.2/whats-new/#backward-compatibility. e.g.:

> Composite resources no longer have native connection details support. You can recreate this feature by composing your own connection details Secret as described in the [connection details composition guide](https://docs.crossplane.io/v2.2/guides/connection-details-composition/). Connection details for managed resources (MRs) aren't affected.

I'm hoping Vale is okay with "aren't affected" since I'm suggesting we use that in a few places 🙏

haarchri (approved):

PR #1071, opened 2026-03-06
docs: fix Upjet metric names across all versions

### LINKED ISSUE #1070: Incorrect upjet metric names in monitoring documentation
### URL

https://docs.crossplane.io/v2.2/guides/metrics/

### What's Wrong?

The Crossplane docs reference upjet metrics using the upjet_resource_* prefix, but the upstream upjet repository uses different metric names.

#### Expected (from upjet source):

  Per `docs/monitoring.md` and `pkg/metrics/metrics.go#L86-L100:`

  - upjet_terraform_cli_duration
  - upjet_terraform_active_cli_invocations
  - upjet_terraform_running_processes

 #### Actual (in Crossplane docs):

  Docs currently list these as `upjet_resource_*` variants.

### Impact:

  Users configuring Prometheus scraping or alerting rules based on the Crossplane docs will use incorrect metric names and get no data.

aswnrana:
Correct CLI-related metrics from `upjet_resource_*` to `upjet_terraform_*` prefix and add the missing `upjet_resource_deletion_seconds` metric to the metrics reference table.

Fixes #1070


haarchri:
thanks we have the Same issue in the other active versions too can you add the fix too ?

haarchri (approved):

PR #1073, opened 2026-03-16
docs: add configuration instructions for provider authentication with Workload Identity

fahedouch:
We had trouble setting up WIF with the current documentation, so I am updating the documentation accordingly.

### DIFF
--- a/content/master/guides/crossplane-with-workload-identity.md
+++ b/content/master/guides/crossplane-with-workload-identity.md
@@ -697,9 +697,30 @@ metadata:
   namespace: crossplane-system
 ```
 
+### Configure provider authentication with Workload Identity
+
+By default, each provider creates its own ServiceAccount with a random name. To use Workload Identity for providers, create a `DeploymentRuntimeConfig` to assign a specific ServiceAccount.
+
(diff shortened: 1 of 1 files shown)


tr0njavolta:
Thank you for the contribution @fahedouch! 

I tested this locally and couldn't proceed with the [DeploymentRuntimeConfig](https://docs.crossplane.io/v2.3/api/#DeploymentRuntimeConfig-spec-deploymentTemplate-spec) as written. I get : 
```
The DeploymentRuntimeConfig "gcp-runtime-config" is invalid:
* spec.deploymentTemplate.spec.template.spec.containers: Required value
* spec.deploymentTemplate.spec.selector: Required value
```

Had to update with  required values of `selector` and `template.spec. containers`:

```
kubectl apply -f - <<EOF
apiVersion: pkg.crossplane.io/v1beta1
kind: DeploymentRuntimeConfig
metadata:
  name: gcp-runtime-config
spec:
  deploymentTemplate:
    spec:
      selector:
        matchLabels: {}
      template:
        spec:
          serviceAccountName: crossplane
          containers:
            - name: package-runtime
EOF
```

It'd be great to update or add a note on these required fields. I can propogate to the older versions and will open another PR to add AWS & Azure specific instructions.

Issue #1075, opened 2026-03-18
consider changing the default function for composition getting started guide


jbw976:
The composition getting started guide currently features `function-patch-and-transform` as the default tab. We should update this to a more modern function that represents a better UX offered by more capable and powerful functions.

The getting started guide is the first composition experience most new Crossplane users will see, so the default tab choice matters. We currently show multiple composition function options:

  - Patch and Transform (current default)
  - Python
  - Go Templating
  - KCL
  - Pythonic

After some maintainer discussion, the community download numbers make a compelling case for `function-go-templating` as the default:

  - function-go-templating is among the most downloaded composition functions in crossplane-contrib
  - Go templating is a familiar experience for users coming from Helm and other Kubernetes tooling
  - It provides a simple, approachable first experience without requiring knowledge of a full programming language or exposing too many function protocol details

Long term we may want to consider Python if we invest further in the SDK and DevEx, but `function-go-templating` would make a good choice for the time being.

Rishab49:
We need to make the tab element "Templated YAML" as the first element of tab list or just make it the active element by default at its (3rd) position?

jbw976:
good question! I think it would make the most sense to make "templated yaml" the first element in the tab list as a means to it becoming the default path.

jbw976:
Fixed in https://github.com/crossplane/docs/pull/1080 by @Rishab49, thank you for this contribution! 🙇‍♂️

PR #1080, opened 2026-04-18
making function-go-templating default tab in composition getting star…

Rishab49:
I have made following changes in this PR
- Updates the tab order in both `Install the function` and `Configure the composition` section making `Templated YAML` default tab.

### DIFF
--- a/content/master/get-started/get-started-with-composition.md
+++ b/content/master/get-started/get-started-with-composition.md
@@ -192,6 +192,36 @@ Pick what language to use to configure how Crossplane turns an `App` XR into a
 
 {{< tabs >}}
 
+{{< tab "Templated YAML" >}}
+Templated YAML is a good choice if you're used to writing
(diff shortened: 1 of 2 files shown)


Rishab49:
Fixes https://github.com/crossplane/docs/issues/1075

jbw976 (commented):
This looks good to me, thanks for making the change to v2.2!

Can you also make the same change to `content/master`, since that essentially becomes the next version (e.g. v2.3) during the release process?

(and please @ mention me on this PR when you've done this, so I don't miss it 😇)

Rishab49:
Hi @jbw976,

I have made the following changes

- Updated the tab order in both `Install the function` and `Configure the composition` section making Templated YAML default tab in content/master

jbw976 (approved):
Great, thanks for making this change in master too! This looks good to me now, let's get it merged! 🙇‍♂️

PR #1083, opened 2026-04-20
Document per-resource poll interval and reconcile-request annotations

### LINKED ISSUE #1082: Document per-resource poll interval and reconcile-now annotations
Document the new functionality introduced in https://github.com/crossplane/crossplane/pull/7239.

yordis:
- Crossplane now supports two new annotations (`crossplane.io/poll-interval` and `crossplane.io/reconcile-requested-at`) that give operators fine-grained control over reconciliation behavior on a per-resource basis, introduced in crossplane/crossplane#7239
- Without per-resource control, operators had to choose a single global poll interval that was either too aggressive for stable resources or too slow for dynamic ones
- The reconcile-request annotation follows the Flux CD pattern, providing a way to trigger immediate re-sync without modifying a resource's spec

Fixes #1082

### DIFF
--- a/content/master/composition/composite-resources.md
+++ b/content/master/composition/composite-resources.md
@@ -267,6 +267,47 @@ spec:
   # Removed for brevity
 ```
 
+### Per-resource poll interval
+
+The {{<hover label="xr-poll-interval" line="7">}}crossplane.io/poll-interval{{</hover>}}
+annotation overrides the controller-level `--poll-interval` for a specific
+composite resource. The annotation accepts any valid Go duration string.
+
+```yaml {label="xr-poll-interval",copy-lines="none"}
+apiVersion: example.org/v1alpha1
+kind: MyDatabase
+metadata:
+  namespace: default
+  name: my-composite-resource
+  annotations:
+    crossplane.io/poll-interval: "24h"
+spec:
+  # Removed for brevity
+```
+
+Read the [managed resource poll interval annotation]({{<ref "../managed-resources/managed-resources#poll-interval">}}) for more details on behavior and validation.
+
+### Triggering immediate reconciliation
+
+The {{<hover label="xr-reconcile-request" line="7">}}crossplane.io/reconcile-requested-at{{</hover>}}
+annotation triggers an immediate reconciliation when its value changes. Set the
+annotation to any value, such as a timestamp, to trigger a reconciliation.
+
+```yaml {label="xr-reconcile-request",copy-lines="none"}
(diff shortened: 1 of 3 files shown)


content/master/composition/composite-resources.md
```diff
@@ -267,6 +267,47 @@ spec:
   # Removed for brevity
 ```
 
+### Per-resource poll interval
+
+The {{<hover label="xr-poll-interval" line="5">}}crossplane.io/poll-interval{{</hover>}}
```
adamwg: Looks like the line numbers are a little off across this PR. This should be line 7, the `crossplane.io/poll-interval` annotation in the snippet below.

adamwg (commented):
lgtm once the line numbers are corrected. Thanks for the writeup!

adamwg (approved):

haarchri (changes requested):
@yordis thanks for the PR can you check Vale issues? 
https://github.com/crossplane/docs/actions/runs/24734156625/job/72368026690?pr=1083 
after that we can merge it in for the coming v2.3 release - thanks man

haarchri (approved):

content/master/managed-resources/managed-resources.md
```diff
@@ -746,6 +748,81 @@ Read
...
+or `24h`.
+
+```yaml {label="poll-interval",copy-lines="none"}
+apiVersion: rds.aws.m.upbound.io/v1beta1
+kind: Instance
+metadata:
+  namespace: default
+  name: my-rds-instance
+  annotations:
+    crossplane.io/poll-interval: "24h"
```
jbw976, line 768: out of curiousity, for this to work on MRs, is it correct that the provider will need to have updated their crossplane-runtime dependency to v2.3.0? anything else that is required on the provider side after that, or will it work automatically after bumping crossplane-runtime? 

is it worth adding a note in this section on this so users can understand this requirement?

joraff: Good point, I'll second this question. It's called out in the `rc.1` release notes, but if someone upgrades without catching this, or starts using crossplane before the providers update and release new versions (which could be a long tail, hopefully not), it could cause some confusion. I think this will be a popular feature.

I don't recall seeing callouts in the crossplane docs on which release introduced a specific feature, as some other project docs do. If this was added, it could provide the opportunity to note this additional requirement. Bonus points for figuring out how to clearly and concisely communicate _how_ to tell when a provider supports it.

PR #1085, opened 2026-04-27
Rewrite install-from-source guide for Nix build system

jbw976:
Crossplane migrated its build system from Earthly to Nix in crossplane/crossplane#6999. The `install-from-source` guide still referenced the old `earthly +...` targets and artifact paths, so following it no longer worked.

Rewrites the guide around `nix build` and `nix run .#push-images`, with updated prerequisites, `result/` artifact paths, and Helm/CLI install steps. The `./nix.sh` Docker wrapper is mentioned as a fallback for readers who don't want to install Nix natively.

### DIFF
--- a/content/master/guides/install-from-source.md
+++ b/content/master/guides/install-from-source.md
@@ -14,7 +14,7 @@ cluster.
 {{< hint "important" >}}
 Installing Crossplane from source is an advanced installation path for users who
 require complete control over the build and deployment process. Most users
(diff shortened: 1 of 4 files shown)


jbw976:
oops, looks like i have some vale errors to fix 😇

jbw976:
OK this PR should be ready for review! 🙇‍♂️

adamwg (approved):
lgtm

PR #1086, opened 2026-04-28
Embed example manifests inline via shortcode (#960)

### LINKED ISSUE #960: Load example manifests and embed them inline
Our docs have plenty of example manifests that users should apply. Especially the getting started guides. We've used a few approaches to this over the years and have rarely been consistent:

1. Plain YAML, with instructions saying to save it to a file and apply it
2. YAML wrapped in a `kubectl apply` command with a bash HERE/EOF doc
3. Plain YAML, with duplicated plain YAML in a manifests dir that you can `kubectl apply -f https://do...`

These all have downsides. The first makes copying less convenient for the user. The second makes the YAML manifest look uglier - wrapped in somewhat arcane shell syntax. The third means maintaining duplicate manifests.

Last time we brainstormed this with @tr0njavolta they had a good suggestion, which is a variant of 3:

* Manifests defined _only_ in a file in the docs repo, under e.g. `content/manifests/`
* Use Hugo to embed manifests inline in the docs
* Show `kubectl apply -f https://docs.crossplane.io/v2.0/manifests/...` style examples

I believe this is what the Kubernetes docs do. I'd like us to adopt this approach.

jonasz-lasut:
Replaces the inconsistent mix of save-and-apply YAML, kubectl heredocs, and duplicated manifest files with a single mechanism: example manifests live as real .yaml files under
content/<ver>/manifests/<path>.yaml, embedded into prose by a new `{{< manifest >}}` shortcode and simultaneously published as fetchable URLs at `https://docs.crossplane.io/<ver>/manifests/...` so readers can run `kubectl apply -f <url>` directly.

I see room for improvement in the future by moving `Function` manifests to a separate directory and use the shared functions in all docs across all versions as this way we'll be able to update `Function` versions in one place only.

Fixes #960 

Changes:
  - Adds a Hugo module mount in `config.yaml` that re-exposes `content/<ver>/manifests/**.yaml` under `static/`, so each manifest publishes verbatim at its natural URL.

  - Adds `themes/geekboot/layouts/shortcodes/manifest.html`. Required arg path= names the manifest relative to the calling page's version. The shortcode auto-derives the version segment from .Page.File.Path so authors never type version strings. Optional args: `apply="false"` suppresses the kubectl block; command= overrides the default verb (`kubectl apply -f`); `label=` wraps the rendered YAML in `<div label="...">` so existing `{{< hover label="..." >}}` annotations keep working. Hostname is based on base URL so preview deploys still emit usable kubectl URLs.

### DIFF
--- a/content/contribute/code-style-guide.md
+++ b/content/contribute/code-style-guide.md
@@ -264,3 +264,21 @@ Apply the {{< hover label="provider" line="2" >}}Provider{{< /hover >}} resource
 
 {{< manifest path="code-style-guide/provider.yaml" label="provider" >}}
 
+### Interact with manifests
+
+Use the `manifest-url` shortcode to interact with manifests applied to the cluster.
+
+```shell
+kubectl <command> -f {{</* manifest-url path="code-style-guide/provider.yaml" */>}}
+```
+You can use `get`, `describe`, `edit`, `delete` and other supported commands to perform the action on the live resource matching the manifest:
+
+```shell
+kubectl <command> -f {{< manifest-url path="code-style-guide/provider.yaml" >}}
+```
+
+{{< hint "note" >}}
+Refer to the
+[`kubectl`](https://kubernetes.io/docs/reference/kubectl/generated/) command
+reference for more information for supported commands and flags.
+{{< /hint >}}
--- a/content/contribute/code-style-guide.md
+++ b/content/contribute/code-style-guide.md
@@ -211,3 +211,56 @@ aws_secret_access_key = $@<aws_secret_key>$@
 ```
 {{</* /editCode */>}}
 ````
+
+## Manifest shortcode
+
+The `manifest` shortcode embeds a YAML manifest inline and generates a `kubectl apply -f` command pointing to the published URL.
+
+Store manifests under `content/<version>/manifests/<page-path>/<file>.yaml`
+based on the referencing page location.
+
+For example, a page at `content/<version>/guides/change-logs.md` references
+`content/<version>/manifests/guides/change-logs/<file>.yaml` with
+`path="guides/change-logs/<file>.yaml"`.
+
+```markdown
+{{</* manifest path="code-style-guide/provider.yaml" */>}}
+```
+This renders as:
+
+{{< manifest path="code-style-guide/provider.yaml" >}}
+
+### Remove `apply` command
+
+Use `apply="false"` to render the YAML without the `kubectl apply -f` block.
+
+```markdown
+{{</* manifest path="code-style-guide/provider.yaml" apply="false" */>}}
+```
+
+{{< manifest path="code-style-guide/provider.yaml" apply="false" >}}
+
+### `generateName`
+
+Use `command="kubectl create -f"` for manifests that use `generateName`. `kubectl apply` doesn't support `generateName`.
+
+```markdown
+{{</* manifest path="code-style-guide/bucket.yaml" command="kubectl create -f" */>}}
+```
+
+{{< manifest path="code-style-guide/bucket.yaml" command="kubectl create -f" >}}
+
+### Line highlighting
+
+Use `label=` to enable dynamic line highlighting with the `hover` shortcode.
+
+```markdown
+Apply the {{</* hover label="provider" line="2" */>}}Provider{{</* /hover */>}} resource.
+
+{{</* manifest path="code-style-guide/provider.yaml" label="provider" */>}}
+```
+
+Apply the {{< hover label="provider" line="2" >}}Provider{{< /hover >}} resource.
+
+{{< manifest path="code-style-guide/provider.yaml" label="provider" >}}
+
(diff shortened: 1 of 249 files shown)


tr0njavolta:
Thanks for the contribution @jonasz-lasut! I was able to review and test:


* Core functionality
    * kubectl apply -f works
    * kubectl edit -f works 
    * kubectl apply --dry-run=client passes for all manifests across all versions
    * Missing manifest file returns build error.
* Shortcode combinations
    * {{< manifest >}} inside {{< tabs >}} works
    * {{< manifest >}} inside {{< expand >}} works
    * apply="false" inside {{< hint >}} works
    * {{< hover >}} on manifests works

Couple things I noticed:
* Hover targeting a manifest inside a tab doesn't render the line highlighting but isn't a blocking error.
* Local hugo build absURL produces `//localhost/1313/...`. Doesn't affect production but breaks copy-paste testing during doc authoring. Manually updating to `http://localhost/1313/...` works.

[I created a PR to your forked branch to add usage information to the docs code styling guide.](https://github.com/jonasz-lasut/docs/pull/1). Once that's accepted or edited, we'll be good to approve and merge 🚀

tr0njavolta (changes requested):
Requesting a usage guide for the shortcode in https://docs.crossplane.io/contribute/code-style-guide/. Created a [PR](https://github.com/jonasz-lasut/docs/pull/1/changes) here

jonasz-lasut:
Really appreciate the review @tr0njavolta 
I've merged the PR that you've opened against my fork without changes, thank you for finding time to test all functionalities.

I'll take a look at fixing hover inside a tab and localhost baseUrl in the follow-up PR sometime next week :)

jonasz-lasut:
Looks like your commit was missing a signoff, I've invited you as a collaborator on my fork so you should be able to push directly to the branch

tr0njavolta (approved):

themes/geekboot/layouts/shortcodes/manifest.html
```diff
@@ -0,0 +1,58 @@
...
+{{- $label := .Get "label" -}}
+{{- $apply := .Get "apply" | default "true" -}}
+{{- $command := .Get "command" | default "kubectl apply -f" -}}
+
+{{- $url := printf "%s/manifests/%s" $version $rel | absURL -}}
+{{- $applyCmd := printf "%s %s" $command $url -}}
+
+{{- $highlightOpts := "linenos=table,linenostart=1" -}}
+
+<div class="gdoc-manifest">
```
jbw976, line 47: one funny thing i noticed is that the manifest code blocks now seem to use a blue/green color for their line numbers, when before they were a grey color. Is something in the logic here affecting that? Compare the line number colors on these two:

Current: https://docs.crossplane.io/v2.2/guides/change-logs/#configure-permissions

Image


NEW: https://deploy-preview-1086--crossplane.netlify.app/v2.2/guides/change-logs/#configure-permissions

Image

content/contribute/code-style-guide.md
```diff
@@ -211,3 +211,56 @@ aws_secret_access_key = $@<aws_secret_key>$@
...
+
+Use `apply="false"` to render the YAML without the `kubectl apply -f` block.
+
+```markdown
+{{</* manifest path="code-style-guide/provider.yaml" apply="false" */>}}
+```
+
+{{< manifest path="code-style-guide/provider.yaml" apply="false" >}}
+
+### `generateName`
```
jbw976, line 243: is it also worth mentioning the `manifest-url` function in this contributing guide page? i see it used quite a lot on some pages like https://deploy-preview-1086--crossplane.netlify.app/v2.2/get-started/get-started-with-composition/

jbw976 (approved):
thank you so much for taking this on @jonasz-lasut, i think it's a lot easier/quicker to walk through the interactive guides now than the previous approach of copying the content to a file each time. This is very appreciated!! 🙇 

I added two small comments of things I noticed, none of them are explicitly blocking - I'll add my approval on as well and we can merge tomorrow either way 😁

content/contribute/code-style-guide.md
tr0njavolta: Thanks for calling this out - added a mention in the contributing guide.

themes/geekboot/layouts/shortcodes/manifest.html
jonasz-lasut: The horizontal grey bar is also missing. Part of the development loop was diffing between the newly generated content and the source page and the agent missed that.
I'll have a look today what may be the root cause

jbw976: ah good eye, yes the gray border bar is missing too! hopefully we can find the root cause without too much trouble 🙏

jonasz-lasut: Line numbers were using `table` instead of `inline` styling and because of that the CSS was not applied. Fortunately that was a oneliner

jbw976:
awesome, it looks like all the feedback is incorporated now, let's merge!! great work @jonasz-lasut and @tr0njavolta!!! 🎉

PR #1090, opened 2026-05-11
feat(protection): document new provider deletion protection with clusterusages

haarchri:
documents the following new crossplane feature we added for v2.3 https://github.com/crossplane/crossplane/pull/7362

### DIFF
--- a/content/master/packages/providers.md
+++ b/content/master/packages/providers.md
@@ -467,7 +467,7 @@ Enable Provider deletion protection with the
 `--enable-provider-deletion-protection`
 [feature flag]({{<ref "../get-started/install#feature-flags">}}).
 This feature also requires
-[Usages]({{<ref "../managed-resources/usages">}}) to be enabled (on by default).
+[Usages]({{<ref "../managed-resources/usages">}}) (on by default).
 
 ```shell {label="protection-helm"}
 helm upgrade --install crossplane crossplane-stable/crossplane \
@@ -476,9 +476,9 @@ helm upgrade --install crossplane crossplane-stable/crossplane \
 ```
 {{</hint >}}
 
-When Provider deletion protection is enabled, Crossplane automatically prevents
+With Provider deletion protection, Crossplane prevents
 the deletion of Providers that still have active managed resources. This
-protects against accidentally removing a Provider and orphaning its managed
+protects against removing a Provider and orphaning its managed
 resources.
 
 Crossplane watches for managed resources associated with each Provider. When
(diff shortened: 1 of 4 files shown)


content/master/packages/providers.md
```diff
@@ -457,6 +457,99 @@ removing their finalizers.
...
+VPC.ec2.aws.upbound.io"`.
+
+When all managed resources of a given type are deleted, Crossplane automatically
+removes the corresponding `ClusterUsage`, allowing the Provider to be deleted.
+
+List all provider protection `ClusterUsages` using the
+`crossplane.io/provider-protection=true` label:
+
+```shell {copy-lines="1"}
+kubectl get clusterusages -l crossplane.io/provider-protection=true
```
jbw976, line 499: when i ran this command, i got more useful printer columns like `DETAILS` - do you not see the same? i think it looks better to have the more details by default in the docs
```
NAME                                                                       DETAILS                                                                            READY   AGE
provider-protection-0d9403ef4e758ec5ee89d2773d0b356af7635adc37548b0eaabb   Provider has active managed resources of type VPC.ec2.aws.m.upbound.io             True    22s
provider-protection-3da1709122b36532e8cbdedbcabde45947f99dd33d931eb682ee   Provider has active managed resources of type Subnet.ec2.aws.m.upbound.io          True    7s
provider-protection-774712aa693d57a5e7ce09a7754d53f3ca5cc59f417083756b2a   Provider has active managed resources of type SecurityGroup.ec2.aws.m.upbound.io   True    7s
```

content/master/packages/providers.md
```diff
@@ -457,6 +457,99 @@ removing their finalizers.
...
+    resourceRef:
+      name: crossplane-contrib-provider-aws-s3
+  reason: "Provider has active managed resources of type Bucket.s3.aws.upbound.io"
+```
+
+Attempting to delete a Provider with active managed resources returns an error
+from the Usage admission webhook:
+
+```shell
+$ kubectl delete provider.pkg.crossplane.io/provider-aws-ec2
```
jbw976: the terminal prompt char should not be included here, the docs add that automatically

currently, the copy to clipboard button is displayed and clicking it copies the entire command along with error message. If we don't mean for folks to copy it, then we should disable it with `{copy-lines="none"}` or if we do want folks to copy it then set it to just line 1 with `{copy-lines="1"}`

jbw976 (commented):
awesome @haarchri, thanks for following up to document this awesome feature! Added a couple comments for your consideration, but otherwise this is looking good! 🙇‍♂️

jbw976:
also there are some Vale errors to fix 😉 
https://github.com/crossplane/docs/actions/runs/25668772512/job/75348163891?pr=1090

haarchri:
@jbw976 done - can you check again ?

content/master/packages/providers.md
jbw976: i see the extra terminal prompt char has been removed, but still checking in with you that you intend for the entire block to be copy-able? 🤔

jbw976 (approved):
looks good @haarchri, thank you!! just the final question about if you intend to make the whole delete output copy-able, otherwise approved! ✅

PR #1091, opened 2026-05-11
docs: argocd healthcheck - mark packages Degraded when Installed=False

ytsarev:
Crossplane packages (Function, Provider, Configuration) expose two status conditions: `Installed` (the package spec resolved and was unpacked) and `Healthy` (the active PackageRevision is healthy). When `spec.package` is changed to a tag that cannot be resolved or unpacked, the new generation reports `Installed=False` (reason: `UnpackingPackage`, `InvalidPackageRevisionSpec`, ...), but the previously active revision continues to report `Healthy=True / HealthyPackageRevision` with a stale `observedGeneration`. The Lua healthcheck only matched positive condition types and therefore stamped the package as Healthy in ArgoCD even though the GitOps change had failed to roll out.

Add an explicit branch that early-returns Degraded when an `Installed` condition is False, mirroring the existing handling for `Synced` and `LastAsyncOperation`. The new branch is a no-op for non-package kinds and for the steady-state `Installed=True / ActivePackageRevision` case.

**Before - the error is unclear and hidden by Progressing:**

image

**After - the error is surface:**

image
image

### DIFF
--- a/content/master/guides/crossplane-with-argo-cd.md
+++ b/content/master/guides/crossplane-with-argo-cd.md
@@ -174,6 +174,14 @@ data:
             end
           end
 
+          if condition.type == "Installed" then
+            if condition.status == "False" then
+              health_status.status = "Degraded"
+              health_status.message = condition.message
+              return health_status
+            end
+          end
+
           if contains({"Ready", "Healthy", "Offered", "Established", "ValidPipeline", "RevisionHealthy"}, condition.type) then
             if condition.status == "True" then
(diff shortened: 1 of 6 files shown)


haarchri:
@ytsarev can you check for vale Issues ?

ytsarev:
@haarchri none of them are related to the change. Do you want me to fix it inline in this PR?

haarchri (approved):

PR #1093, opened 2026-05-14
Update render documentation

### LINKED ISSUE #1084: Document recent changes to render
### What's Missing?

https://github.com/crossplane/crossplane/pull/7339 makes a fairly significant change to how `crossplane render` works, which brings a few new limitations and a few new features. We should document:

- How to override the Crossplane release, image, or binary to use for rendering.
- Limitations around Docker:
    - By default, `render` runs the Crossplane controller image under Docker, but this can be avoided with the `--crossplane-binary` flag.
    - Injecting or showing function pipeline context requires mounting a socket into the Crossplane docker container when using Docker mode. This does not work on Windows.

adamwg:
Render has changed in three important ways that require docs updates:

- It now uses `crossplane internal render` as its backend, the invocation of which can be customized.
- `crossplane render` has moved to `crossplane composition render` (though the old name is a hidden alias for backward compatibility).
- `crossplane alpha render op` has moved to `crossplane operation render` and, being an alpha command, is hidden by default.

Fixes #1084

### DIFF
--- a/content/master/composition/compositions.md
+++ b/content/master/composition/compositions.md
@@ -344,19 +344,20 @@ learn how to install and use the Crossplane CLI.
 {{< /hint >}}
 
 {{<hint "important">}}
-Running `crossplane render` requires [Docker](https://www.docker.com).
+Running `crossplane composition render` requires
+[Docker](https://www.docker.com).
 {{< /hint >}}
 
 Provide a composite resource, composition and composition functions to render
 the output locally.
 
 ```shell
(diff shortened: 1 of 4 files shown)


lsviben (approved):
LGTM. 

One question, do we have in docs somewhere a note/hint that the command changed? As in "it was like this, and since version xxx its like this?"

haarchri:
@lsviben we have the older folders of the documentation - so folks can switch between the versions

haarchri (approved):

Issue #1092, opened 2026-05-12
Document the developer experience tooling


adamwg:
### What's Missing?

v2.3.0 of the Crossplane CLI will include the first version of our new DevEx tooling - see crossplane/crossplane#6840 and crossplane/cli#10.

adamwg:
Done in #1110

PR #1097, opened 2026-05-21
Update CLI reference docs for v2.3.0

adamwg:
Use the new `crossplane generate-docs` command to generate CLI reference docs from the built-in CLI help. Update both `master` and `v2.3`, since they're the same for now.

Toward #1092


jbw976 (commented):
there are a couple funny headings that are rendering odd, e.g. https://deploy-preview-1097--crossplane.netlify.app/v2.3/cli/command-reference/#read-the-xrd-from-stdin

are those at the right level? are they supposed to be headings? 🤔

adamwg:
> there are a couple funny headings that are rendering odd, e.g. https://deploy-preview-1097--crossplane.netlify.app/v2.3/cli/command-reference/#read-the-xrd-from-stdin
> 
> are those at the right level? are they supposed to be headings? 🤔

Good catch - we missed properly converting the help for `crossplane xrd convert` to markdown, so it's producing some weird rendering. For now, I think I'll manually edit the generated docs to fix this and the vale warnings, then propagate the changes back to the CLI's built-in help that powers the generation.

adamwg:
I've fixed as many vale issues as possible while still auto-generating docs. See crossplane/cli#30 for details.

jbw976:
amazing @adamwg, great work at tackling most all of the vale issues!

do you think it would be possible (or reasonable) to include exclusions in the CLI docs templates where needed for the specifically failing Vale styles, e.g. `gitlab.SubstitutionWarning`, `gitlab.Units`, `Google.WordList`, etc.? or is that just not a path we think is worth it? if it's not worth it (or not possible), do we live with the remaining failures? 🤔

adamwg:
> do you think it would be possible (or reasonable) to include exclusions in the CLI docs templates where needed for the specifically failing Vale styles, e.g. `gitlab.SubstitutionWarning`, `gitlab.Units`, `Google.WordList`, etc.? or is that just not a path we think is worth it? if it's not worth it (or not possible), do we live with the remaining failures? 🤔

I thought about doing this, but since the remaining issues are in text that comes from the kong command/argument tags rather than markdown-formatted text, I think we would have to add these exclusions for the entire document, rather than scoped around the specific instances we want to ignore. I also looked at whether we could ignore [only specific matches](https://vale.sh/docs/formats/markdown#comments), but couldn't make it work - not sure if our version of vale doesn't support that, or if the checks/styles we're using don't support it. If we could ignore `Google.WordList["k8s"]` for the whole doc that would feel a lot  better than ignoring `Google.WordList` entirely.

I'm inclined to accept the existing issues, since they should only get flagged when we update the CLI docs. But, I would definitely prefer to have vale return a clean status if we can figure out a way to do it.

adamwg:
I put some more time into this today and figured out how to ignore specific matches (does require a newer vale version). I've updated this PR with the output, and crossplane/cli#30 to include (a) the ability to disable a specific check for a specific command, and (b) a CI job that generates docs and runs vale with the config from this repo, so that we ensure no new issues creep in.

jbw976 (approved):
awesome @adamwg, thanks for finding a solution that will keep all the status checks green!! ✅

PR #1099, opened 2026-05-29
Update CLI install instructions

adamwg:
Update the CLI install instructions to reference the new artifact location (cli.crossplane.io) and use the install script hosted there rather than pulling it directly from GitHub. While we're here, clarify the language a bit.


jbw976 (approved):
LGTM @adamwg, i tested out a couple of commands and they worked OK for me.

Do you need to incorporate this into https://github.com/crossplane/docs/pull/1104 too?

adamwg:
> Do you need to incorporate this into #1104 too?

Yes, #1104 will need an update after merging this and #1103.

PR #1103, opened 2026-06-03
docs: document crossplane beta upgrade check

### LINKED ISSUE #1102: cli: document `upgrade check` command usage
After https://github.com/crossplane/crossplane/pull/7451 is merged and we have the new `upgrade check` command in the release-1.20 branch, we should update the docs for v1.20 to help understand how to use it. 

This could include:

* basic usage/help output on the CLI reference page: https://docs.crossplane.io/v1.20/cli/command-reference/
* mention in the v2 backwards compatibility page: https://docs.crossplane.io/latest/whats-new/#backward-compatibility 
* mentions on v2 upgrade page: https://docs.crossplane.io/latest/guides/upgrade-to-crossplane-v2/

jbw976:
Documents `crossplane beta upgrade check`, the read-only command added in crossplane/crossplane#7451. It scans a v1.x control plane for usage of features that v2 removes or changes, so operators can see what would break before upgrading to v2.

- **CLI reference** (`content/v1.20/cli/command-reference.md`): a new `beta upgrade check` section covering its five checks, flags, the `--skip-managed-resources` and `--concurrency` performance/load tradeoffs, and example output. It lives only in v1.20, since the command ships only in the v1.20 CLI.
- **Pointers from the v2.x docs** (`master` and `v2.1` through `v2.3`): tip boxes on the upgrade guide (under "Removed features") and the backward-compatibility section of the release notes. They tell readers to run the check with the v1.20 CLI while preparing to upgrade, and link to the v1.20 reference and the "Download other CLI versions" instructions.

The example output is a verbatim run against a seeded v1.20 cluster, so the findings, format, and UX are accurate.

Resolves #1102.

### DIFF
--- a/content/v1.20/cli/command-reference.md
+++ b/content/v1.20/cli/command-reference.md
@@ -37,7 +37,7 @@ Server Version: v1.17.0
 
 ## render
 
-The `crossplane render` command previews the output of a 
+The `crossplane render` command previews the output of a
 [composite resource]({{<ref "../concepts/composite-resources">}}) after applying
 any [composition functions]({{<ref "../concepts/compositions">}}).
 
@@ -46,20 +46,20 @@ The `crossplane render` command requires you to use composition functions.
 {{< /hint >}}
 
 The `crossplane render` command connects to the locally running Docker
-Engine to pull and run composition functions. 
+Engine to pull and run composition functions.
 
-{{<hint "important">}} 
+{{<hint "important">}}
 Running `crossplane render` requires [Docker](https://www.docker.com/).
 {{< /hint >}}
 
 Provide a composite resource, composition and composition function YAML
 definition with the command to render the output locally.
 
-For example, 
+For example,
 `crossplane render xr.yaml composition.yaml function.yaml`
 
-The output includes the original composite resource followed by the generated 
-managed resources. 
+The output includes the original composite resource followed by the generated
+managed resources.
 
 {{<expand "An example render output" >}}
 
@@ -118,7 +118,7 @@ to connect to the local Docker Engine and run composition functions.
 ### Provide function context
 
 The `--context-files` and `--context-values` flags can provide data
-to a function's `context`.  
+to a function's `context`.
 The context is JSON formatted data.
 
 ### Include function results
@@ -131,7 +131,7 @@ outputs.
 
 Composition functions can only change the `status` field of a composite
 resource. By default, the `crossplane render` command only prints the
-`status` field with `metadata.name`.  
+`status` field with `metadata.name`.
 
 Use `--include-full-xr` to print the full composite resource,
 including the `spec` and `metadata` fields.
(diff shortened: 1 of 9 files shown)


content/v1.20/cli/command-reference.md
```diff
@@ -867,29 +867,150 @@ Configuration/platform-ref-aws                             v0.9.0    True
...
+The `crossplane beta upgrade check` command inspects a live Crossplane `v1.x`
+control plane and reports any usage of features that v2 removes or changes the
+behavior of. Run it against a cluster _before_ upgrading to see what would break.
+
+{{<hint "important" >}}
+`crossplane beta upgrade check` only reads from a cluster. It never creates,
+modifies, or deletes resources.
+
+Run it against a Crossplane `v1.x` control plane. The command looks for the v2
+[breaking changes](https://docs.crossplane.io/latest/whats-new/#backward-compatibility),
```
jbw976, line 883: this uses a fully qualified URL instead of a relative `<ref "...` style link because we want to link to `/latest`, but I don't think you can do that with those relative style links because hugo tries to resolve it to a real content directory at build time (which doesn't exist, `latest` is a redirect at the netlify level) unfortunately...

adamwg: In https://github.com/crossplane/docs/commit/fff2c446294d72e4c4c354dde5ca7f553f0603d3 (part of #1104), I've updated CI to call `netlify_build.sh` rather than calling hugo directly, since that script actually creates `/latest` by copying the latest content.

jbw976: good to know! just to be clear, is this comment an FYI or are you also saying that you want me to incorporate this feedback and wait to merge this PR until after #1104 is merged?

adamwg: Totally fine with leaving this as-is - we can update it after #1104 is merged if we want, but I don't think the absolute link causes any harm.

content/v1.20/cli/command-reference.md
```diff
@@ -867,29 +867,150 @@ Configuration/platform-ref-aws                             v0.9.0    True
...
+* **Unqualified package sources**, which break because v2 removes the default
+  registry and requires fully qualified package references.
+
+By default the command uses the Kubernetes configuration in `~/.kube/config` and
+sweeps the entire control plane: cluster-scoped resources and all namespaces. Use
+`--kubeconfig` and `--context` to target a specific cluster. For example,
+`crossplane beta upgrade check --kubeconfig ~/.kube/prod.yaml --context prod`.
+
+Use `--namespace` to restrict namespaced checks, like Claims, to a single
+namespace. For example, `crossplane beta upgrade check --namespace team-a`.
```
adamwg: May be worth calling out that cluster-scoped checks still run when `--namespace` is specified.

adamwg (approved):
lgtm.

content/v1.20/cli/command-reference.md
jbw976: good call! added a quick follow-up sentence here to clarify

PR #1104, opened 2026-06-04
Version the CLI documentation separately

adamwg:
In the near future, the CLI version will diverge from the core Crossplane version (i.e., CLI v2.4 will be released before Crossplane v2.4). All the documentation is currently versioned together, which will be confusing or incorrect when the CLI version no longer matches the core version.

Move the CLI documentation into its own set of versioned directories, and update cross-links between the core and CLI documentation sets to be explicitly versioned. Give the CLI documentation its own sidebar, similar to the contributing guide, and its own version menu. Rename the "CLI Reference" page to "CLI Overview" so that it's a better section header. Update the sidebar to have a link to the CLI docs when browsing the core documentation, and a link back to the core docs when browsing the CLI docs; add an `<hr>` above these links to visually separate them from the regular TOC for the current section.

### DIFF
--- a/content/cli/v1.20/command-reference.md
+++ b/content/cli/v1.20/command-reference.md
@@ -1,5 +1,5 @@
 ---
-weight: 50
+weight: 100
 title: Command Reference
 description: "Command reference for the Crossplane CLI"
 ---
@@ -8,7 +8,7 @@ description: "Command reference for the Crossplane CLI"
 <!-- vale Google.Headings = NO -->
 The `crossplane` CLI provides utilities to make using Crossplane easier.
 
-Read the [Crossplane CLI overview]({{<ref "../cli">}}) page for information on
+Read the [Crossplane CLI overview]({{<ref "/cli/v1.20">}}) page for information on
 installing `crossplane`.
 
 ## Global flags
@@ -38,8 +38,8 @@ Server Version: v1.17.0
 ## render
 
 The `crossplane render` command previews the output of a
-[composite resource]({{<ref "../concepts/composite-resources">}}) after applying
-any [composition functions]({{<ref "../concepts/compositions">}}).
+[composite resource]({{<ref "/v1.20/concepts/composite-resources">}}) after applying
+any [composition functions]({{<ref "/v1.20/concepts/compositions">}}).
 
 {{< hint "important" >}}
 The `crossplane render` command requires you to use composition functions.
@@ -178,7 +178,7 @@ built-in support in [function-go-templating](https://github.com/crossplane-contr
 ## xpkg
 
 The `crossplane xpkg` commands create, install and update Crossplane
-[packages]({{<ref "../concepts/packages">}}) as well as enable authentication
+[packages]({{<ref "/v1.20/concepts/packages">}}) as well as enable authentication
 and publishing of Crossplane packages to a Crossplane package registry.
 
 ### xpkg build
@@ -193,9 +193,9 @@ The CLI applies the required annotations and values to meet the
 [Crossplane XPKG specification](https://github.com/crossplane/crossplane/blob/main/contributing/specifications/xpkg.md).
 
 The `crossplane` CLI supports building
-[configuration]({{< ref "../concepts/packages" >}}),
-[function]({{<ref "../concepts/compositions">}}) and
-[provider]({{<ref "../concepts/providers" >}}) package types.
+[configuration]({{< ref "/v1.20/concepts/packages" >}}),
+[function]({{<ref "/v1.20/concepts/compositions">}}) and
+[provider]({{<ref "/v1.20/concepts/providers" >}}) package types.
 
 #### Flags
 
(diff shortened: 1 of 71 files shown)


content/master/composition/compositions.md
```diff
@@ -339,7 +339,7 @@ don't need a Crossplane control plane to do this. The Crossplane CLI uses Docker
 Engine to run functions.
 
 {{<hint "tip">}}
-See the [Crossplane CLI docs]({{<ref "../cli">}}) to
+See the [Crossplane CLI docs]({{<ref "/cli/master">}}) to
```
jbw976: i haven't thought through deeply enough to understand the impact of this, but my first instinct is that this may be difficult to maintain over time, especially because of the way this repo does releases. Would this impact the release process where we essentially copy /master to /v2.new? would we need to search/replace all these instances each time we release? 🤔

adamwg: I agree, this will be difficult to maintain. I wonder if we should instead point links from the core docs to `cli/latest/...` (and vice versa), and then pin to a specific version later only if the content is removed/irrelevant in a future `latest` version (which should get flagged by the htmltest link checker).

adamwg: Pushed a commit to pin the references in `master/...` to `latest`, so they don't need an update when `master` becomes v2.4. I don't see any harm in leaving the existing versioned docs pointing to specific CLI versions, especially given that some of them reference commands that have moved in the latest CLI release.

content/cli/v1.20/command-reference.md
```diff
@@ -8,7 +8,7 @@ description: "Command reference for the Crossplane CLI"
 <!-- vale Google.Headings = NO -->
 The `crossplane` CLI provides utilities to make using Crossplane easier.
 
-Read the [Crossplane CLI overview]({{<ref "../cli">}}) page for information on
+Read the [Crossplane CLI overview]({{<ref "../">}}) page for information on
```
jbw976: this link seems to lose the version number, e.g. it goes to `/cli/latest/` when it should be honoring that you're on the v1.20 CLI docs.

i think this happens on all versions of this page.

netlify_build.sh
```diff
@@ -3,6 +3,10 @@
 # Which which version is the "Latest"?
 LATEST_VER="2.3"
 
+# Which CLI docs version is the "Latest"? Tracked separately from core so the
+# two version lines can diverge. Keep in sync with params.cliLatest in config.yaml.
+CLI_LATEST_VER="2.3"
```
jbw976, line 8: will there be a step in the cli release issue template to bump this and the value in config.yaml? 🤔

jbw976 (changes requested):
Thanks @adamwg - yeah I think it's okay to have the versions numbers directly in the links for the older versions of docs --> cli docs, then for master docs --> cli latest. When we cut our next crossplane docs release, the master folder will be copied to v2.next and that latest link will still make sense because at that point core and cli have diverged and just going to latest cli docs is really the only thing that makes sense 🤓 

A couple experience notes from trying out the preview site (i think these should be fixed):

* When I click on `Crossplane CLI Documentation` in the left hand navigation pane, it takes me to https://deploy-preview-1104--crossplane.netlify.app/cli/latest/ and there is no version picker on that page. That's a regression from the current functionality you get on say https://docs.crossplane.io/latest/cli/, the version picker dropdown is always shown on the page.
* Currently valid URLs are now broken, e.g. https://deploy-preview-1104--crossplane.netlify.app/latest/cli/command-reference/. This works fine on the live site https://docs.crossplane.io/latest/cli/command-reference/, but is broken on the preview site. We should set up a redirect so all the links out there in the wild pointing at current CLI documentation don't start becoming 404's.

netlify_build.sh
adamwg: Yes, good call-out. I have a release checklist in progress, and will make sure this is on it.

content/cli/v1.20/command-reference.md
adamwg: Yep, good catch - fixed.

adamwg:
@jbw976 

> A couple experience notes from trying out the preview site (i think these should be fixed):
> 
> * When I click on `Crossplane CLI Documentation` in the left hand navigation pane, it takes me to https://deploy-preview-1104--crossplane.netlify.app/cli/latest/ and there is no version picker on that page. That's a regression from the current functionality you get on say https://docs.crossplane.io/latest/cli/, the version picker dropdown is always shown on the page.

Indeed, in fact we were missing the version selector on the `latest` versions of the core docs as well. Fixed in the latest.

> * Currently valid URLs are now broken, e.g. https://deploy-preview-1104--crossplane.netlify.app/latest/cli/command-reference/. This works fine on the live site https://docs.crossplane.io/latest/cli/command-reference/, but is broken on the preview site. We should set up a redirect so all the links out there in the wild pointing at current CLI documentation don't start becoming 404's.

Good catch. Added a redirect (and tested that it works as expected) in the latest.

content/master/guides/upgrade-to-crossplane-v2.md
```diff
@@ -58,8 +58,8 @@ Crossplane v2 removes these deprecated features:
 
 {{<hint "tip" >}}
 The Crossplane CLI can find these features for you. Run
-[`crossplane beta upgrade check`]({{<ref "../../v1.20/cli/command-reference#beta-upgrade-check">}})
-with the [`v1.20` CLI]({{<ref "../../v1.20/cli#download-other-cli-versions">}}) against your control plane to scan for every removed feature
+[`crossplane beta upgrade check`]({{<ref "/cli/v1.20/command-reference#beta-upgrade-check">}})
```
jbw976, line 61: yay, thank you for keeping my new `crossplane beta upgrade check` references working in the v1.20 docs! 🎉

jbw976 (approved):
awesome, this all looks to be working as I would expect on the preview site. nice work, this is a bit complicated, but I think it came together with a nice experience that allows our core and CLI docs to be versioned independently now! 🙌 🙇‍♂️

PR #1105, opened 2026-06-05
Remove v2.0-preview docs and clean up old redirects

adamwg:
v2.0-preview was the docs branch for the v2 RC releases last year. v2 is now the main content of the docs, and v2.0 is EOL, so we should clean these up.

While we're here, clean up old redirects for EOL v1.x releases and replace them with a generic redirect that covers all EOL versions.


jbw976 (commented):
Cool, this seems like the same experience folks would have going to a v2.0 docs link which was EOL'd with the v2.3 release - so status quo would be achieved.

However, should we consider just reviving this redirect process that seems to have died with v1.12? e.g. should we keep these redirects, add ones for all other EOL'd versions since then, and update the release issue template to include adding a new redirect with each release? 🤔 

that may be a nice way to go here, to at least give a chance for an old link, that still has a  content path that exists on newer versions, to work. what do you think?

adamwg:
> However, should we consider just reviving this redirect process that seems to have died with v1.12? e.g. should we keep these redirects, add ones for all other EOL'd versions since then, and update the release issue template to include adding a new redirect with each release? 🤔
> 
> that may be a nice way to go here, to at least give a chance for an old link, that still has a content path that exists on newer versions, to work. what do you think?

@jbw976 I've pushed a commit that adds a wildcard redirect to cover all old versions. It will still return a 404 if a page doesn't exist in the latest docs, of course, but should work for many common cases since most of our docs are pretty stable.

This redirect relies on the fact that Netlify won't redirect requests for pages that exist. I tested it by manually building the docs with the `netlify_build.sh` script locally and then serving them as static content with `netlify dev -d public --framework '#static'`.

netlify.toml
```diff
@@ -27,19 +27,11 @@ from = "/docs/*"
 to = "/:splat"
 status = 302
 
-# Redirects for EOL versions
+# Redirect requests for EOL version docs to latest. These might return 404 if
+# docs have moved around, but for many docs the redirect will work. This relies
+# on the fact that Netlify will not perform a redirect for a page that exists.
```
jbw976: i think this behavior is described in https://docs.netlify.com/manage/routing/redirects/rewrites-proxies/#shadowing

So if it's a page that exists, nothing will happen, we'll just serve the requested page URL.

If it's a page from an old EOL version of the docs that still exists, this redirect will take it from `/v.EOL/<page>` --> `/latest/<page>` 

If it's a page from an old EOL version of the docs that no longer exist, the redirect will go `/v.EOL/<page>` --> `/latest/<page>` --> 404

If it's a current/supported version of a page that really doesn't exist, the redirect will go `/v2.3/<page>` --> `/latest/<page>` --> 404

Does that sound right to you? I think that's it

I wasn't able to test this on the docs preview site, e.g. https://deploy-preview-1105--crossplane.netlify.app/v2.3/foo/ just shows 404 right away as opposed to redirecting to `/latest/foo` as I would expect from these rules.

Perhaps the netlify.toml changes aren't respected on preview sites? 🤔

jbw976 (approved):
I think this looks reasonable - let me know if i'm not understanding the expected behavior for the redirects and if you expect to see this working on the preview site. Thanks @adamwg!

netlify.toml
adamwg: I just re-tested locally and my redirect rule was not working; must have missed restarting the dev server after making a change earlier.

I've updated the redirect rule so it does work now, and the behavior is as you describe. I also put the wildcard version matching rule at the end of the redirects, since Netlify processes them in order and uses the first match, so the version redirect would override the others if it was at the top. I'll check to see if this is also working on the preview site.

It's a little unfortunate that non-existent pages in supported versions redirect to `/latest/...` before hitting the 404, since users would have to replace the `/latest/` in the URL bar in addition to fixing their other error. But, not sure it matters that much in practice - I doubt we have a lot of people manually typing docs URLs, and redirecting for EOL versions helps with the much more common case of external stale links.

adamwg: Verified that it is working on the preview site now :+1:

PR #1110, opened 2026-06-15
Add a getting started guide for control plane projects

adamwg:
Walk users through building their first control plane project using a simple WebApp example. This guide includes two "choose your own adventure" sections:

- Generating an XRD from either an example XR or SimpleSchema.
- Writing a function in any of our supported languages: Go Templating, Python, Go, or KCL.

### DIFF
--- a/content/cli/master/get-started/_index.md
+++ b/content/cli/master/get-started/_index.md
@@ -1,0 +1,7 @@
+---
+title: "Get Started"
+weight: 200
+description: "Get started with the Crossplane CLI"
+---
+
+{{< auto-index >}}
--- a/content/cli/master/get-started/get-started-with-control-plane-projects.md
+++ b/content/cli/master/get-started/get-started-with-control-plane-projects.md
@@ -482,6 +482,8 @@ metadata:
     gotemplating.fn.crossplane.io/composition-resource-name: service
   name: {{ $xr.metadata.name }}
   namespace: {{ $xr.metadata.namespace }}
+  labels:
+    app.kubernetes.io/name: {{ $xr.metadata.name }}
 spec:
   selector:
     app.kubernetes.io/name: {{ $xr.metadata.name }}
@@ -766,6 +768,7 @@ func (f *Function) RunFunction(_ context.Context, req *fnv1.RunFunctionRequest) 
 		Metadata: &metav1.ObjectMeta{
 			Name:      xr.Metadata.Name,
 			Namespace: xr.Metadata.Namespace,
+			Labels:    &labels,
 		},
 		Spec: &corev1.ServiceSpec{
 			Selector: &labels,
--- a/content/cli/master/get-started/get-started-with-control-plane-projects.md
+++ b/content/cli/master/get-started/get-started-with-control-plane-projects.md
@@ -1,0 +1,1135 @@
+---
+title: Get Started with Control Plane Projects
+weight: 100
+description: "Build a control plane project with the Crossplane CLI"
+---
+
+This guide shows how to use the Crossplane CLI to build a platform API from
+scratch. You create a _project_, define a new `WebApp` custom resource, and
+configure how Crossplane composes it. When a user creates a `WebApp`, Crossplane
+creates a Kubernetes `Deployment` and a `Service`.
+
+The Crossplane CLI scaffolds the project, generates the API and composition, and
+runs the project on a local development control plane so you can test it without
+deploying to a shared cluster.
+
+{{<hint "tip">}}
+This guide shows how to write the composition function in Go, Python, KCL, and
+templated YAML. You can pick your preferred language.
+{{</hint>}}
+
+A `WebApp` custom resource looks like this:
+
+```yaml
+apiVersion: platform.example.com/v1alpha1
+kind: WebApp
+metadata:
+  name: podinfo
+  namespace: default
+spec:
+  image: docker.io/stefanprodan/podinfo:6.11.0
+  replicas: 3
+  ports: [9898]
+```
+
+**The `WebApp` is the custom API your users use to deploy a containerized
+workload.**
+
+When a user creates a `WebApp`, Crossplane creates a `Deployment` and a
+`Service`:
+
+```yaml
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: podinfo
+  namespace: default
+  labels:
+    app.kubernetes.io/name: podinfo  # Copied from the WebApp's name
+spec:
+  replicas: 3  # Copied from the WebApp's spec
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: podinfo
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: podinfo
+    spec:
+      containers:
+      - name: podinfo
+        image: docker.io/stefanprodan/podinfo:6.11.0  # Copied from the WebApp's spec
+        ports:
+        - containerPort: 9898  # Copied from the WebApp's spec
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: podinfo
+  namespace: default
+spec:
+  selector:
+    app.kubernetes.io/name: podinfo
+  ports:
+  - protocol: TCP
+    port: 9898
+    targetPort: 9898
+```
+
+## Prerequisites
+
+This guide requires:
+
+* The [Crossplane CLI]({{<ref "../_index.md">}}) installed
+* A Docker compatible container runtime
+
+The CLI builds the project's functions and runs a local development control
+plane in a [KIND](https://kind.sigs.k8s.io) cluster. Both require a working
+Docker installation. You don't need an existing Kubernetes cluster.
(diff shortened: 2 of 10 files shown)


content/cli/master/get-started/_index.md
```diff
@@ -0,0 +1,7 @@
+---
```
jbw976, line 1: would it make sense to link this new guide somehow from the core crossplane docs getting started section for more visibility? i.e. maybe a placeholder page so it shows up in that nav pane, but that page then just has a short intro and link to this content?

that may make things confusing, so feel free to disregard, just thinking of more ways to make this new guide discoverable

jbw976, line 1: will you add this to v2.3 also so it shows up in latest?

content/cli/master/get-started/get-started-with-control-plane-projects.md
```diff
@@ -0,0 +1,1135 @@
...
+tree example-project-webapp
+example-project-webapp
+├── apis
+├── crossplane-project.yaml
+├── examples
+├── functions
+├── operations
+└── tests
+
+6 directories, 1 file
```
jbw976, line 116: mine says 5 dirs and 1 file, but i don't appear to be missing anything 🔍 

maybe just a mac vs linux difference?

```
❯ tree example-project-webapp

example-project-webapp
├── apis
├── crossplane-project.yaml
├── examples
├── functions
├── operations
└── tests

5 directories, 1 file
```

content/cli/master/get-started/get-started-with-control-plane-projects.md
```diff
@@ -0,0 +1,1135 @@
...
+The `crossplane xrd generate` command reads the example XR, infers the field
+types from their contents, and generates an appropriate XRD:
+
+```shell
+crossplane xrd generate examples/webapp/podinfo.yaml --from xr
+```
+
+The command writes the generated XRD to `apis/webapps/definition.yaml`:
+
+```yaml
```
jbw976: we can consider using `{copy-lines="none"}` for all of these blocks that are just showing what the tooling was expected to output. we aren't expecting folks to copy/paste this anywhere right? using copy none may reduce any potential confusion.

This comment applies to a few different blocks in this doc.

content/cli/master/get-started/get-started-with-control-plane-projects.md
```diff
@@ -0,0 +1,1135 @@
...
+    step: crossplane-contrib-function-auto-ready
+```
+
+## Add an example
+
+If you generated your XRD from SimpleSchema, create an example `WebApp` so you
+can render and run the project against a realistic input. If you generated your
+XRD from an example XR, you already have the example and can skip this step.
+
+Create `examples/webapp/podinfo.yaml`:
```
jbw976: the directory name under `apis/` was `webapps`, but the directory here under `examples/` is singular `webapp`. is that intentional?

content/cli/master/get-started/get-started-with-control-plane-projects.md
```diff
@@ -0,0 +1,1135 @@
...
+
+Before running the project, use `crossplane composition render` to preview what
+your composition produces. The `render` command runs your composition pipeline
+locally and prints the resources the composition would create.
+
+In a project directory, `render` discovers and builds the composition's
+functions automatically, so you only pass the example XR and the composition:
+
+```shell
+crossplane composition render examples/webapp/podinfo.yaml apis/webapps/composition.yaml
```
jbw976: when i run this command for the go based function/composition, i get this error:
```
crossplane: error: cannot build embedded functions: failed to build function "compose-webapp": failed to build runtime images: failed to build function: build: go build: exit status 1: # example.com/my-org/example-project-webapp/compose-webapp
                   ./fn.go:124:29: undefined: corev1.ServiceAPIVersionV1 (but have ServiceApiVersionV1)
```

Indeed, the same error is shown for that line in my IDE by the language server, and it looks correct that `ServiceApiVersionV1` is the correct casing in `schemas/go/io/k8s/core/v1/core.go`.

Making that change seems to fix the error and the `render` is able to run.

content/cli/master/get-started/get-started-with-control-plane-projects.md
```diff
@@ -0,0 +1,1135 @@
...
+
+Check that Crossplane created the `Deployment` and `Service`:
+
+```shell {copy-lines="1"}
+kubectl get deployment,service -l app.kubernetes.io/name=podinfo
+NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
+deployment.apps/podinfo   3/3     3            3           45s
+
+NAME              TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
+service/podinfo   ClusterIP   10.96.142.110   <none>        9898/TCP   45s
```
jbw976, line 1080: the service is not shown in my output, but it does seem to exist. Looks like it's missing the label, I think the go code is not applying that label to the Service object.

```
❯ kubectl get deployment,service -l app.kubernetes.io/name=podinfo

NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/podinfo   3/3     3            3           20s

❯ k get service
NAME         TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
kubernetes   ClusterIP   10.96.0.1       <none>        443/TCP    106s
podinfo      ClusterIP   10.96.109.136   <none>        9898/TCP   28s

❯ k get service podinfo -o json | jq '.metadata.labels'
{
  "crossplane.io/composite": "podinfo"
}
```

jbw976 (commented):
nice guide! the new devex tooling really is powerful and smooth 💪 

I found a couple issues while walking through it using `go`, let me know if you have any questions about the feedback! 🙇‍♂️

content/cli/master/get-started/get-started-with-control-plane-projects.md
jbw976: i think this issue affects templated yaml (function-go-templating) also, but python and KCL do correctly apply the label to the service.

content/cli/master/get-started/_index.md
adamwg: Yep, I'll copy it to v2.3 and the new v2.4. Figured it would be easier to review and update in master first :-).

content/cli/master/get-started/get-started-with-control-plane-projects.md
adamwg: Huh, weird. Must be a Linux vs. BSD `tree` difference. I guess Linux counts the top-level directory?

content/cli/master/get-started/get-started-with-control-plane-projects.md
adamwg: This is due to https://github.com/crossplane/cli/issues/113, which is fixed now so uppercase `API` should always be generated. (I actually noticed the issue while testing this guide as well!)

content/cli/master/get-started/_index.md
adamwg: Done!

adamwg: Good idea. I thought about adding a note in the header of the existing "get started with compositions" page, but there's already a lot of frontmatter there so it kinda gets lost. I've taken your suggestion and added a stub page in that section that links to the CLI docs. Open to suggestions on where it sits in the nav and its content.

content/cli/master/get-started/get-started-with-control-plane-projects.md
adamwg: Good idea, done.

content/cli/master/get-started/get-started-with-control-plane-projects.md
adamwg: In the `up` tooling we had `up example generate`, which used the singular kind under `examples/`, so I had just followed that convention. But we don't have such a command in the upstream tooling (yet, at least), and I think the consistency of `apis/webapps` and `examples/webapps` is nice, so updated to use the plural.

content/cli/master/get-started/get-started-with-control-plane-projects.md
adamwg: Good catch - added the labels to the Go and go-templating examples so the command works consistently now.

content/master/get-started/get-started-with-control-plane-projects.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Get Started With Control Plane Projects
+weight: 150
+description: "Build a platform with the Crossplane CLI"
+---
+
+Starting in v2.3.0, the Crossplane CLI includes development tools that simplify
```
jbw976: consider omitting the `Starting in v2.3.0` part, perhaps it can be implicit from the fact this page only shows up in 2.3+.

jbw976 (approved):
awesome, thanks for the follow up here @adamwg, this is a great guide to get folks going to with devex/projects! 🎉

PR #1119, opened 2026-06-24
Add CLI v2.4 documentation

adamwg:
This is a copy of the CLI `master` documentation, with the command reference updated using `crossplane generate-docs`.

Ref crossplane/release#86


content/cli/v2.4/_index.md
```diff
@@ -0,0 +1,74 @@
+---
+weight: 50
+title: CLI Overview
+description: "Command-line tools for Crossplane development"
+cascade:
+    version: "2.4"
```
jbw976, line 6: does something also need to be set so this version is now "latest"? should that be in this PR or a follow-up?

jbw976 (approved):

content/cli/v2.4/_index.md
adamwg: Good catch - need to add that to the release checklist.

PR #1127, opened 2026-07-19
fix(docs): Configuration packages CAN include MRAPs

### LINKED ISSUE #1126: [Web Bug] - Managed Resource Activation Policies
There is the following claim in the docs "Configuration packages can include MRAPs to declare their resource dependencies"
But there is no available information anywhere else in the docs how you would do that.
This sentence somehow suggest to reader there exists an automatic way to include MRAPs into a configuration, but there seems to be no way?

URL: https://docs.crossplane.io/latest/managed-resources/managed-resource-activation-policies/#configuration-package-activation

boxcee-interview:
Closes #1126

## Problem
The MRAP docs claimed Configuration packages can include MRAPs, but the xpkg spec restricts Configuration packages to Configuration, CompositeResourceDefinition, and Composition objects only.

## Solution
- Clarify that MRAPs cannot be bundled in Configuration packages
- Link to the xpkg spec as authoritative reference
- Show the correct approach: use `dependsOn` in crossplane.yaml for Provider dependencies
- Explain that MRAPs are applied separately by cluster operators
- Recommend documenting MRD requirements in package READMEs
- Update best practices section to align

## How to Test
Review the updated MRAP documentation at the Configuration package activation section. Verify the changes align with the [xpkg spec](https://github.com/crossplane/crossplane/blob/main/contributing/specifications/xpkg.md).

### DIFF
--- a/content/v2.1/managed-resources/managed-resource-activation-policies.md
+++ b/content/v2.1/managed-resources/managed-resource-activation-policies.md
@@ -291,6 +291,10 @@ spec:
   - certificates.acm.aws.m.crossplane.io # For HTTPS
 ```
 
+The package linter accepts `ManagedResourceActivationPolicy` objects in
+Configuration packages, so an MRAP can ship in the package's `package.yaml`
+([package linter](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/xpkg/lint.go#L75)).
+
 <!-- vale Google.Headings = NO -->
 <!-- vale Microsoft.HeadingAcronyms = NO -->
 ## Working with MRAPs


content/v2.1/managed-resources/managed-resource-activation-policies.md
```diff
@@ -458,9 +485,10 @@ package dependencies.
    `*.s3.aws.m.crossplane.io` works for future S3 resources)
 3. **Group related resources logically** - Create MRAPs that activate
    resources teams actually use together
-4. **Include activation dependencies in Configuration packages** -
-   Configuration packages should declare what MRDs they need rather than
-   assuming resources are available
+4. **Document MRD requirements in Configuration packages** -
```
haarchri: thanks for enhancing the documentation, but it is possible to add an MRAP to an configuration Package: https://github.com/crossplane/crossplane-runtime/blob/main/pkg/xpkg/lint.go#L75 thats the linter the cli is using - can you update your PR ?

boxcee-interview: You're right, thanks for catching that — I missed the linter's `IsActivationPolicy` case. The previous wording was wrong; I've reverted to the original claim and added a pointer to the package linter as the authoritative check (commit 337eea73). Vale passes on the file.

haarchri: looks better - can we do the changes in all the active docs ? and please don't link the linter - lets link the cli docs for crossplane xpkg build which is enough in my point of view

Issue #1137, opened 2026-08-20
`hugo server` fails local build with deprecated warnings


jbw976:
### What's Wrong?

Running `hugo server` locally results in a few warning messages for using deprecated features:

```
WARN  deprecated: project config key languageCode was deprecated in Hugo v0.158.0 and will be removed in a future release. Use locale instead.
WARN  deprecated: module.mounts.includeFiles was deprecated in Hugo v0.153.0 and will be removed in a future release. Replaced by the simpler 'files' setting, see https://gohugo.io/configuration/module/#files
WARN  deprecated: .Site.Data was deprecated in Hugo v0.156.0 and will be removed in a future release. Use hugo.Data instead.
WARN  deprecated: .Language.LanguageCode was deprecated in Hugo v0.158.0 and will be removed in a future release. Use .Language.Locale instead.
```

We should look into moving off these to their modern equivalents.

`hugo version` output:
```
hugo v0.162.1-bba860e3eda3091efe8c6d1a96bc49a29ad2d5f6+extended darwin/arm64 BuildDate=2026-05-28T17:40:44Z VendorInfo=gohugoio
```

haarchri:
we using an older version atm: https://github.com/crossplane/docs/blob/master/netlify.toml#L19 

`  HUGO_VERSION = "0.147.3"`

Issue #1136, opened 2026-08-20
`hugo server` fails local build with many `REF_NOT_FOUND`


jbw976:
### What's Wrong?

Running `hugo server` locally reports a number of issues for `REF_NOT_FOUND`. This was discovered during the v2.4 release process https://github.com/crossplane/docs/issues/1135, but it appears to be reproducible on main before the v2.4 release also.

Errors:
```
ERROR [en] REF_NOT_FOUND: Ref "/latest/packages": "/Users/jared/dev/go/src/github.com/crossplane/docs/content/cli/master/get-started/get-started-with-control-plane-projects.md:131:23": page not found
ERROR [en] REF_NOT_FOUND: Ref "/latest/packages/configurations/": "/Users/jared/dev/go/src/github.com/crossplane/docs/content/cli/master/get-started/get-started-with-control-plane-projects.md:1:1": page not found
ERROR [en] REF_NOT_FOUND: Ref "/latest/packages": "/Users/jared/dev/go/src/github.com/crossplane/docs/content/cli/v2.3/get-started/get-started-with-control-plane-projects.md:131:23": page not found
ERROR [en] REF_NOT_FOUND: Ref "/latest/packages": "/Users/jared/dev/go/src/github.com/crossplane/docs/content/cli/v2.4/get-started/get-started-with-control-plane-projects.md:131:23": page not found
ERROR [en] REF_NOT_FOUND: Ref "/latest/packages/configurations/": "/Users/jared/dev/go/src/github.com/crossplane/docs/content/cli/v2.3/get-started/get-started-with-control-plane-projects.md:1:1": page not found
ERROR [en] REF_NOT_FOUND: Ref "/latest/packages/configurations/": "/Users/jared/dev/go/src/github.com/crossplane/docs/content/cli/v2.4/get-started/get-started-with-control-plane-projects.md:1:1": page not found
ERROR [en] REF_NOT_FOUND: Ref "/cli/latest/get-started/get-started-with-control-plane-projects": "/Users/jared/dev/go/src/github.com/crossplane/docs/content/master/get-started/get-started-with-control-plane-projects.md:10:25": page not found
ERROR [en] REF_NOT_FOUND: Ref "/cli/latest/get-started/get-started-with-control-plane-projects": "/Users/jared/dev/go/src/github.com/crossplane/docs/content/v2.3/get-started/get-started-with-control-plane-projects.md:10:25": page not found

ERROR error building site: logged 8 error(s)
```

[...]

PR #1138, opened 2026-08-20
chore: docs updates for v2.4 release

jbw976:
This PR addresses all items in the codebase for the v2.4 docs release issue:

* https://github.com/crossplane/docs/issues/1135

I also deleted the `content/cli/v2.1` directory since the CLI still shipped as part of core crossplane in that release, so it makes sense to remove those docs along with the main v2.1 docs too.  I thought about putting that into the release item checklist, but eventually we won't need to do this since the CLI split out from core and has its own lifecycle starting in v2.3.

Note that while trying to test locally with `hugo server`, I discovered these issues:

* https://github.com/crossplane/docs/issues/1136
* https://github.com/crossplane/docs/issues/1137

So I haven't been able to test locally and we'll see if we can rely on the docs preview site 🙏


netlify_build.sh
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
 
 # Which which version is the "Latest"?
-LATEST_VER="2.3"
+LATEST_VER="2.4"
 
 # Which CLI docs version is the "Latest"? Tracked separately from core so the
 # two version lines can diverge. Keep in sync with params.cliLatest in config.yaml.
-CLI_LATEST_VER="2.3"
+CLI_LATEST_VER="2.4"
```
jbw976, line 8: looks like the CLI v2.4 release may have missed this ^

params.cliLatest was already updated, so nothing to do there

jbw976:
looks like some Vale errors, which I don't expect given I didn't change any content...

jbw976:
Pushed https://github.com/crossplane/docs/pull/1138/changes/c8fb6850853e9d2c522295792c4c6bea9eeeb46b to fix the Vale errors.

they had been sitting there since we bumped the Vale version in https://github.com/crossplane/docs/pull/1097, then finally got tested today when this PR moved all the files around for v2.4.

negz (approved):

PR #1140, opened 2026-08-24
feature(cli): update v2.5.0

haarchri:
- add v2.5.0 for the cli pages


ezgidemirel (approved):

jbw976 (commented):
should v2.2 be removed actually? that CLI release was still shipped as part of core crossplane and Crossplane v2.2 is still supported right now 🤔

haarchri:
@jbw976  Happy to re-add v2.2, done. I don't think anything in those docs is broken or misleading, but "keep docs for versions we still support" is a reasonable default and it costs us nothing here.

That said, I want to flag why I removed it in the first place, because I think the rule needs an update rather than a one-off application.

The retention rule as stated is anchored to core's support window: core maintains the three most recent minors, CLI v2.2 shipped as part of core v2.2, therefore v2.2 CLI docs stay until core v2.5 ships. That works as long as the CLI ships with core. It doesn't from v2.3 onward, per @adamwg's one-pager (crossplane/crossplane#7350) the CLI versions and releases independently now. If we cut CLI v2.6 next month with no core release attached, the rule gives us no answer about what to keep or drop.

#7350 does have a partial answer in the latest revision: CLI minors are cut for substantial changes and alongside each core minor, and only the two most recent CLI minors are supported. But that doc is still open, it lives in c/c, and it covers CLI-to-core compatibility rather than docs retention.

So concretely for this PR: v2.2 is back in. For everything after, I've opened https://github.com/crossplane/cli/issues/300 in crossplane/cli to pin down the CLI's own maintenance and docs retention policy, so the next release doesn't reopen this. Two things need deciding there:

1. The CLI support window, is it "two most recent CLI minors" as #7350 says, "three minors" to mirror core, or a time-based window like core's nine months?
2. Docs retention — does it follow the CLI's window, or does it follow core's for the CLI versions that shipped inside core (<= v2.3)?

My suggestion is the CLI's own window governs going forward, with a grandfather clause for the versions that shipped in core: keep v2.2 docs until core v2.2 is EOL, and after that retention tracks the CLI support window only. That gets us out of this PR cleanly and gives us a rule that survives an off-cycle CLI release.

jbw976 (commented):
looks good to me @haarchri, thank you for opening the tracking issue to follow up on the confusion for CLI version maintenance policy! 🙇‍♂️

jbw976 (approved):

PR #1141, opened 2026-08-25
fix(404): fix 404 link

haarchri:
during some algolia Cralwer changes i found that https://docs.crossplane.io/latest/concepts/pods/#leader-election is not available anymore - as we restructured the docs page with the v2 release - so lets use the v1.20 link instead

image


jbw976 (approved):
ah nice, thanks for this clean up!! 🙌

PR #1051, opened 2026-01-28
feat(algolia): add facets based on version for search

haarchri:
When browsing older documentation versions (for example, v1.20), search results currently all content from latest....
This leads to confusion when users discover features or APIs in search results that don’t actually exist in the version they’re viewing.

This PR introduces version-aware search filtering using **Algolia** facet filters. https://www.algolia.com/doc/guides/managing-results/refine-results/faceting
Search results will be restricted to the documentation version the user is currently viewing.

- Added logic to detect the current page version via `.Page.Params.version`.
- Configured DocSearch to apply a `facetFilters` value dynamically:

```
['version:X.XX']
```

- Added special handling for the master version - uses `version:0.0.0-master`, which matches the existing meta tag convention where master is labeled `0.0.0-master` to keep it de-ranked in search

### Background
Each page already includes:
<meta name="docsearch:version" content="X.XX">

which the Algolia crawler indexes.

```
curl https://docs.crossplane.io/v2.0/composition/  | grep -A 10 "docsearch"
<meta name=docsearch:modified content='August 14, 2025'><meta name=docsearch:version content="2.0">
```

```
curl https://docs.crossplane.io/v1.20/concepts/  | grep -A 10 "docsearch"
<meta name=docsearch:modified content='May 21, 2025'><meta name=docsearch:version content="1.20">
```

### Required: Algolia configuration
After this PR is merged, someone with Algolia admin access must enable faceting:

- Open Algolia Dashboard
- Go to the **crossplane** index
- Navigate to Configuration → Facets
- Add:
searchable(version)
to Attributes for faceting

- Save

### Testing:

```
curl http://localhost:1313/v2.1/composition/ | grep -A 10 "docsearch"
...
<script>
    docsearch({
    container: '#docSearch',
    appId: 'xxx',
    indexName: 'crossplane',
    apiKey: 'xxxxx',
    placeholder: 'Search the docs',
    searchParameters: {
      facetFilters: ['version:2.1']
    }
    });
</script>
```

To verify this PR:
- Open any versioned docs page (for example, /v1.20/concepts/providers/)
- Open DevTools → Network
- Perform a search
- Inspect the request to algolia.net
```
[...]

PR #1142, opened 2026-08-25
feat(search): restrict search results to the current track (CLI vs core docs)

haarchri:
Since 2.3 the CLI is versioned and released separately from Crossplane core, so the two tracks have diverged (CLI is at v2.5, core at v2.4). The `version` facet from #1051 can't tell them apart: core v2.4 and CLI v2.4 pages both index as `version:2.4`, so search results leaked across tracks.

- Add a `track` facet filter: CLI pages (`/cli/**`) filter on `track:cli`, everything else (core, knowledge base, contributing) on `track:core`, ANDed with the existing version filter.
- Fix the "older version" search banner on CLI pages: it compared against the core `latest` (2.4), so CLI v2.5 pages wrongly showed it. It now uses `cliLatest` and links to `/cli/v2.5/`.

Algolia is already configured: the crawler tags records with `track`, a full recrawl has run, and `filterOnly(track)` is set on the index.

Check after the Crawl the track field is available:
Screenshot 2026-08-25 at 16 52 54

Check on CLI Page Search:
image

Check that the Track FacetFilter is set during Search:
Screenshot 2026-08-25 at 16 53 59

image

### DIFF
--- a/themes/geekboot/layouts/partials/scripts.html
+++ b/themes/geekboot/layouts/partials/scripts.html
@@ -19,6 +19,13 @@
         {{ end }}
     {{ end }}
 
+    {{/* Which track is this page on? Records are tagged "cli" or "core" by the
+         crawler; everything that isn't the CLI track (core, knowledge base,
+         contributing guide) searches the "core" records. */}}
+    {{ $track := cond (eq (.Page.Params.track | default "core") "cli") "cli" "core" }}
+    {{ $trackLatest := cond (eq $track "cli") .Site.Params.cliLatest .Site.Params.latest }}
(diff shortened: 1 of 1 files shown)


jbw976 (approved):
this looks reasonable, thanks for digging into this issue! go ahead and merge if you're feeling good about your testing.

i do have one question about the behavior i'm seeing on the preview site, and i'm not sure if it'll just start working once we merge, or we still don't have something configured correctly yet:

when searching for `trace` in the CLI docs area, e.g. on https://deploy-preview-1142--crossplane.netlify.app/cli/latest/command-reference/, no results newer than v2.2 are shown, even though `crossplane resource trace` is definitely on the newer pages.

Image

is something preventing v2.3+ CLI pages from being indexed?

haarchri:
the command-reference page was too big after crawl > 100KB - so i updated the cralwer - looks now as expected 

image

PR #1144, opened 2026-08-25
fix(netlify): set CLI_LATEST_VER to 2.5 ...

haarchri:
we missed to bump this


jbw976 (approved):
good catch, i think we found the same thing in https://github.com/crossplane/docs/pull/1138 for v2.4.

do you want to update the crossplane/release CLI issue templates so this doesn't keep getting missed? 😄

PR #1145, opened 2026-08-25
ci: sync Algolia crawler config with published doc versions

haarchri:
The Algolia crawler config for docs.crossplane.io maintains an explicit pathsToMatch allow-list of the doc versions it should index. Today that list lives only in the Algolia dashboard and is edited by hand, so a version that ships without someone remembering to update it is crawled but never indexed under its own version facet, search for it silently returns nothing. The same applies in reverse: EOL'd versions linger in the config after their folders are removed.

This adds a workflow that makes the version folders under /content the single source of truth and syncs the crawler config to them:

Derives the version list from the content/v* and content/cli/v* folders, no version inputs needed, and core and CLI stay independent since each has its own folder set.
Confirms every derived version is actually live (HTTP 200 on /<version>/ and /cli/<version>/) before touching anything, and refuses to run if the folder list resolves empty.
GETs the current crawler config, replaces all versioned (vX.Y) pathsToMatch entries with the computed set — adding new versions and pruning EOL ones in the same pass and PATCHes it back. Non-versioned entries (latest, master, contribute, knowledge-base) pass through untouched, so the workflow doesn't need to know about them.
Optionally triggers a full reindex (needed after a release ;)).
Because the sync is idempotent (re-running against an in-sync config produces zero diff), it's safe to run on every release, the step summary shows the resulting list plus explicit added/removed sections, and a dry-run input prints the patch without pushing.

The Crawler API only accepts whole top-level properties, so this is a read-modify-write of the entire actions array rather than a targeted edit. Two consequences shaped the implementation: a concurrency group prevents two runs from clobbering each other, and the jq keeps negation patterns (e.g. !https://docs.crossplane.io/latest/concepts/composition/) at the end of pathsToMatch, since micromatch applies them positionally. Versioned negations are pruned along with their version.

The release issue template is updated accordingly: the two manual steps (edit config in the dashboard, trigger reindex) collapse into "run the Sync Docs Search workflow after publishing".

Access is gated by workflow_dispatch itself, which requires write access on the repo.

Tested on my fork Repo:

Update config
with dry-run:
Screenshot 2026-08-26 at 09 51 49

Trigger the Cralwer:
Screenshot 2026-08-25 at 18 07 50

We need to add the following Secrets - if PR accepted ping me that we can set them up:
      ALGOLIA_CRAWLER_USER_ID
      ALGOLIA_CRAWLER_API_KEY
      ALGOLIA_CRAWLER_ID

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -1,0 +1,129 @@
+name: Sync Docs Search
+
+on:
+  workflow_dispatch:
+    inputs:
+      reindex:
+        description: 'Trigger a full reindex afterwards'
+        type: boolean
+        default: true
+      dry-run:
+        description: 'Resolve and print the patch without pushing'
+        type: boolean
+        default: false
+  workflow_call:
+    inputs:
+      reindex: { type: boolean, required: false, default: true }
+      dry-run: { type: boolean, required: false, default: false }
+    secrets:
+      ALGOLIA_CRAWLER_USER_ID: { required: true }
+      ALGOLIA_CRAWLER_API_KEY: { required: true }
+      ALGOLIA_CRAWLER_ID:      { required: true }
+
+permissions:
+  contents: read
+
+# the config update is read-modify-write — never run two at once
+concurrency:
+  group: docs-search-config
+  cancel-in-progress: false
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    env:
+      CRAWLER: https://crawler.algolia.com/api/1/crawlers
+      ID: ${{ secrets.ALGOLIA_CRAWLER_ID }}
+      AUTH: ${{ secrets.ALGOLIA_CRAWLER_USER_ID }}:${{ secrets.ALGOLIA_CRAWLER_API_KEY }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Resolve versions from content folders
+        run: |
+          set -euo pipefail
+          core=$(find content -mindepth 1 -maxdepth 1 -type d -name 'v[0-9]*' -printf '%f\n' | sort -V)
+          cli=$(find content/cli -mindepth 1 -maxdepth 1 -type d -name 'v[0-9]*' -printf '%f\n' | sort -V)
+          [[ -n "$core" && -n "$cli" ]] || { echo "::error::no version folders under content/ — refusing to sync"; exit 1; }
+
+          # every version we are about to index must already be deployed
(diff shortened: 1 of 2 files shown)


.github/workflows/promote.yml
```diff
@@ -0,0 +1,111 @@
...
+        default: ''
+      reindex:
+        description: 'Trigger a full reindex afterwards'
+        type: boolean
+        default: true
+      dry-run:
+        description: 'Resolve and print the patch without pushing'
+        type: boolean
+        default: false
+  workflow_call:
```
haarchri, line 14: added this as we want to have this part of a different ci-run (crossplane/cli) for example

.github/workflows/promote.yml
```diff
@@ -0,0 +1,111 @@
+name: Promote Docs Search
```
jbw976: does this new workflow mean we can remove the algolia steps from the docs release issue template? https://github.com/crossplane/docs/blob/master/.github/ISSUE_TEMPLATE/new_release.md 

if so, want to include that together in this PR?

.github/workflows/promote.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Build patch
+        run: |
+          set -euo pipefail
+          curl -sf -u "$AUTH" "$CRAWLER/$ID?withConfig=true" > current.json
+
+          jq -e '.config.actions | map(select(.name=="CrossplaneCrawler")) | length == 1' current.json >/dev/null \
+            || { echo "::error::action 'CrossplaneCrawler' not found in crawler config"; exit 1; }
+
+          # negation patterns must stay at the end of pathsToMatch
+          jq --argjson add "$ADD" '
```
jbw976: part of the release workflow is to also remove the EOL docs version - i'm not seeing any removal logic in this step, so that would still have to be done manually. is it possible to add that into this workflow too?

jbw976 (commented):
cool, thanks @haarchri for trying to automate this work! 🔥

.github/workflows/promote.yml
haarchri: updated

.github/workflows/promote.yml
haarchri: wanted to run only add for the first iteration - but with delete we can use the folders for core and cli and check for new / old entries and update the crawler - updated!

jbw976 (approved):
awesome, thank you for incorporating the EOL version removal too, as well as listing this in the release issue template steps! 🎉

PR #1146, opened 2026-08-28
fix(search): scope search on /latest pages to the latest version

haarchri:
### Problem

Searching from `/latest/` or `/cli/latest/` pages returns hits from every indexed version (v2.5, v2.4, v2.3, master, latest), showing the same page multiple times. This became visible on the CLI track, where nearly all content lives in a single command-reference page per version, but the core track has the same problem, it's just masked because Algolia's version-descending ranking fills the first screen of results with latest-version hits before the older duplicates appear.

The Netlify build strips the `version` front matter from the `latest` copies (`netlify_build.sh`), so `.Page.Params.version` is empty on all `/latest/`pages and the search template skipped the `version:` facet filter entirely, sending only `facetFilters: [["track:cli"]]`. That was correct when the crawler only indexed `/latest`, but is wrong now that every version folder is indexed with a `version` attribute.

Also guards the "Searching in an older version" banner so it doesn't fire on latest pages, where the effective search version is now set but the page has no version param.

Additionally, search results now stay on the host being viewed: Algolia records store absolute docs.crossplane.io URLs, which made search untestable on deploy previews since every hit navigated back to production. transformItems rewrites each hit's origin to window.location.origin (a no-op on production), so this PR's changes can be verified directly on its own preview.

### DIFF
--- a/themes/geekboot/layouts/partials/scripts.html
+++ b/themes/geekboot/layouts/partials/scripts.html
@@ -9,22 +9,28 @@
 
 <script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3"></script>
 <script>
-    {{ $cur_ver := "" }}
-    {{/* Only filter if we have a version parameter and it's not "0" (static pages) */}}
-    {{ if and .Page.Params.version (ne .Page.Params.version "0") }}
-        {{ if eq .Page.Params.version "master" }}
-            {{ $cur_ver = "0.0.0-master" }}
-        {{ else }}
-            {{ $cur_ver = .Page.Params.version }}
-        {{ end }}
-    {{ end }}
-
     {{/* Which track is this page on? Records are tagged "cli" or "core" by the
          crawler; everything that isn't the CLI track (core, knowledge base,
          contributing guide) searches the "core" records. */}}
     {{ $track := cond (eq (.Page.Params.track | default "core") "cli") "cli" "core" }}
     {{ $trackLatest := cond (eq $track "cli") .Site.Params.cliLatest .Site.Params.latest }}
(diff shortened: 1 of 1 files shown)


themes/geekboot/layouts/partials/scripts.html
```diff
@@ -34,12 +40,25 @@
...
         transformItems(items) {
+            // Records store absolute production URLs; point hits at the host
+            // being viewed so deploy previews and local servers stay testable.
+            items.forEach(item => {
+                try {
+                    const itemUrl = new URL(item.url);
+                    if (itemUrl.origin !== window.location.origin) {
+                        item.url = window.location.origin + itemUrl.pathname + itemUrl.search + itemUrl.hash;
+                    }
+                } catch (e) {
```
jbw976, line 54: nit: the catch all exceptions and swallow them looks a bit suspicious - do we have a clear idea of the errors we want to avoid here? is there a reasonable way to handle them more purposefully?

jbw976 (approved):
awesome @haarchri! this seems to work as expected on the preview site right now while doing a couple of search queries. 

probably the best improvement is searching for "install" on the live docs site on a /latest page gives you the v1.20 installation page as the top result, but now this change gives you the installation page for the latest concrete version (e.g. v2.4). much better! 🙌 

just one question about swallowing exceptions

themes/geekboot/layouts/partials/scripts.html
haarchri: The only thing that can throw in here is `new URL(item.url)`, a TypeError if an Algolia record ever has a missing or malformed URL. In that case the right behavior is exactly what the catch does: leave the hit untouched and let it point at wherever the record says. A broken search result is much better than a broken results list, so I'd rather not have this take down `transformItems` for the whole page of hits.

